PR #16, opened 2025-05-05
feat: iOS autogrow

szydlovsky:
known issues:
- deleting all the text from the input makes it too small - this will get nicely fixed with the next PR that introduces fonts and font sizes
- the first few frames blink at some point between the initial height and the first calculated height. This happens because component gets mounted before the `ComponentView` - which is the only thing capable of calculating estimated height. I have an idea for an upgrade here and will think of it soon

https://github.com/user-attachments/assets/ab00653a-3b6f-46a1-b3b1-8471e068ce4d

### DIFF
--- a/ios/cpp/ReactNativeRichTextEditorViewComponentDescriptor.h
+++ b/ios/cpp/ReactNativeRichTextEditorViewComponentDescriptor.h
@@ -4,9 +4,6 @@
 #include <react/renderer/core/ConcreteComponentDescriptor.h>
 #include <ReactNativeRichTextEditor/ReactNativeRichTextEditorViewShadowNode.h>
 
-// to be deleted
-#include <iostream>
-
 namespace facebook::react {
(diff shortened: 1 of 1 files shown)


ios/cpp/ReactNativeRichTextEditorViewComponentDescriptor.h
```diff
@@ -0,0 +1,20 @@
+#pragma once
+#include <react/debug/react_native_assert.h>
+#include <ReactNativeRichTextEditor/Props.h>
+#include <react/renderer/core/ConcreteComponentDescriptor.h>
+#include <ReactNativeRichTextEditor/ReactNativeRichTextEditorViewShadowNode.h>
+
+// to be deleted
+#include <iostream>
```
exploIF: Q: should it be removed in this PR or in the future?

szydlovsky: oh yeah, this PR, good catch 😄

PR #17, opened 2025-05-06
feat: Android - links

exploIF:


android/src/main/java/com/swmansion/reactnativerichtexteditor/events/LinkHandler.kt
```diff
@@ -0,0 +1,16 @@
+package com.swmansion.reactnativerichtexteditor.events
+
+import com.facebook.react.bridge.ReactContext
+import com.facebook.react.uimanager.UIManagerHelper
+import com.swmansion.reactnativerichtexteditor.ReactNativeRichTextEditorView
+
+class LinkHandler(private val editorView: ReactNativeRichTextEditorView) {
+  fun onClick(url: String) {
+    val formattedUrl = if (!url.startsWith("https://") && !url.startsWith("http://")) "https://$url" else url
```
szydlovsky: I'd think of adding `www` as well

android/src/main/java/com/swmansion/reactnativerichtexteditor/styles/SpecialStyles.kt
```diff
@@ -0,0 +1,77 @@
...
+    return Triple(result, start, end)
+  }
+
+  private fun afterTextChangedLinks(result: Triple<String, Int, Int>) {
+    val linkHandler = editorView.linkHandler ?: return
+    val spannable = editorView.text as Spannable
+    val (word, start, end) = result
+
+    // TODO: Consider using more reliable regex, this one matches almost anything
+    val urlPattern = android.util.Patterns.WEB_URL.matcher(word)
```
szydlovsky, line 65: I can suggest some regular expressions I used in my native playground:
https://github.com/software-mansion-labs/ios-text-editor-playground/blob/f48fe2374a3138f47a97df89a57f63fb1a3fb21d/RichTextEditorPlayground/RichTextEditorPlayground/Playground/TextEditor/Utils/LinksUtils.swift#L88-L97
I am checking firstly for full url, then for at least `www` starting one and lastly any url-like structure.
Mind that the escape characters are doubled in there because swift says so (you normally need just one) 😄

szydlovsky: It still isn't the most elegant solution but it's better than nothing.

exploIF: This sounds good, however I would like to discuss this topic a little bit more once our app is stable enough - I will add it to my list

android/src/main/java/com/swmansion/reactnativerichtexteditor/events/LinkHandler.kt
exploIF: I've changed logic a little bit here. Instead of formatting url inside library I moved this logic to example app so It's library user responsibility to properly format url. Why? I can imagine that in some apps for example `mailto:` is valid url, in such scenarios we should not prepend url with http protocol. As for now we don't have to detect such urls in text automatically, but ideally we should add such logic in the future

PR #21, opened 2025-05-08
feat: android - mentions style

exploIF:


example/src/components/MentionPopup.tsx
```diff
@@ -0,0 +1,96 @@
...
+    </View>
+  );
+};
+
+const styles = StyleSheet.create({
+  container: {
+    position: 'absolute',
+    width: '100%',
+    zIndex: 1,
+    // padding: 16,
```
szydlovsky: still needed?

exploIF: Nope, thanks

PR #31, opened 2025-05-14
feat: add basic inline styles for iOS

szydlovsky:
Adds `italic`, `underline` and `strikethrough` styles

### DIFF
--- a/ios/ReactNativeRichTextEditorView.h
+++ b/ios/ReactNativeRichTextEditorView.h
@@ -1,3 +1,4 @@
+#pragma once
 #import <React/RCTViewComponentView.h>
 #import <UIKit/UIKit.h>
 #import "EditorConfig.h"
--- a/ios/config/EditorConfig.h
+++ b/ios/config/EditorConfig.h
@@ -1,3 +1,4 @@
+#pragma once
 #import <UIKit/UIKit.h>
 
 @interface EditorConfig: NSObject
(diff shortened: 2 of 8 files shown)


ios/styles/UnderlineStyle.mm
```diff
@@ -0,0 +1,90 @@
...
+
++ (StyleType)getStyleType { return Underline; }
+
+- (instancetype)initWithEditor:(id)editor {
+  self = [super init];
+  _editor = (ReactNativeRichTextEditorView *) editor;
+  return self;
+}
+
+- (void)applyStyle:(NSRange)range {
```
exploIF, line 17: I see that some of the methods looks the same or almost the same for different styles. Is it possible to share at least some code between them? That way it should be easier to maintain code in the future. This is not something that has to be done in this PR, but if it's possible it would be great to simplify this logic at some point

exploIF:
Btw, please remember about squashing commits ;)

ios/styles/UnderlineStyle.mm
szydlovsky: It is like that by design. `applyStyle` is the general toggle/apply style method that gets called by the editor on the style class itself. While `addAttributes`, `removeAttributes`, `removeTypingAttributes` and `addTypingAttributes` are there to do some partial tasks and are used both in `applyStyle` and are gonna be used in the conflicts and blocking of the styles out there. So far the styles I added are super simple so the code indeed looks very similar but soon it will deviate a lot and that's why I decided to have all these methods working for each style to then use them in more generalised contexts - like the said blocking and conflicts. Either way, I can come up with some at least slight code sharing here

exploIF: Understandable, we can verify whether it makes sense to have some code sharing once all styles are functional, wdyt?

PR #48, opened 2025-05-23
iOS links

szydlovsky:
adds:
- manual links via `addLink`
- `onLinkDetected` event
- setup for words' modifications reactions
- automatic links detection together with any conflicting styles handling
- rearranges files a bit


ios/styles/LinkStyle.mm
```diff
@@ -0,0 +1,345 @@
...
+      error:nullptr
+  ];
+  
+  NSString *regexPassedUrl = nullptr;
+  
+  if ([fullRegex numberOfMatchesInString:word options:0 range:NSMakeRange(0, word.length)]) {
+    [self addLink:word url:word range:wordRange manual:NO];
+    regexPassedUrl = word;
+  } else if ([wwwRegex numberOfMatchesInString:word options:0 range:NSMakeRange(0, word.length)]) {
+    NSString *httpWord = [NSString stringWithFormat:@"http://%@", word];
```
exploIF: I think it's safer if we default to `https`

PR #49, opened 2025-05-23
fix: change onFocus and onBlur event names

szydlovsky:
Similar to what RNScreens went through here: https://github.com/software-mansion/react-native-screens/pull/2154, iOS crashes when naming events `onFocus` and `onBlur`, thus renaming them to something right and similar. Checked that Android builds correctly but some double-check much appreciated @exploIF.

### DIFF
--- a/src/RichTextInput.tsx
+++ b/src/RichTextInput.tsx
@@ -69,8 +69,8 @@ export interface RichTextInputProps extends Omit<ViewProps, 'children'> {
   cursorColor?: ColorValue;
   selectionColor?: ColorValue;
   style?: ViewStyle | TextStyle;
-  onInputFocus?: () => void;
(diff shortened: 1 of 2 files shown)


src/RichTextInput.tsx
```diff
@@ -69,8 +69,8 @@ export interface RichTextInputProps extends Omit<ViewProps, 'children'> {
   cursorColor?: ColorValue;
   selectionColor?: ColorValue;
   style?: ViewStyle | TextStyle;
-  onFocus?: () => void;
-  onBlur?: () => void;
+  onInputFocus?: () => void;
```
exploIF: I'm not sure about this part. Won't be more clear for devs using this lib to use `onFocus` and `onBlur`?

PR #54, opened 2025-05-28
feat: android mentions - allow setting dynamic attributes

exploIF:


src/ReactNativeRichTextEditorViewNativeComponent.ts
```diff
@@ -34,7 +34,7 @@ export interface OnChangeStateEvent {
   isMention: boolean;
 }
 
-export interface OnPressLinkEvent {
+export interface OnPressLink {
```
szydlovsky, line 37: why the name change? all other events end with `event` out here

src/RichTextInput.tsx
```diff
@@ -75,12 +80,12 @@ export interface RichTextInputProps extends Omit<ViewProps, 'children'> {
...
   onChangeHtml?: (e: NativeSyntheticEvent<OnChangeHtmlEvent>) => void;
   onChangeState?: (e: NativeSyntheticEvent<OnChangeStateEvent>) => void;
-  onPressLink?: (e: NativeSyntheticEvent<OnPressLinkEvent>) => void;
+  onPressLink?: (e: OnPressLink) => void;
   onLinkDetected?: (e: NativeSyntheticEvent<OnLinkDetectedEvent>) => void;
   onStartMention?: (indicator: string) => void;
   onChangeMention?: (e: OnChangeMentionEvent) => void;
   onEndMention?: (indicator: string) => void;
-  onPressMention?: (e: NativeSyntheticEvent<OnPressMentionEvent>) => void;
+  onPressMention?: (e: OnPressMention) => void;
```
szydlovsky, line 88: and why is it no longer `NativeSyntheticEvent`?

src/RichTextInput.tsx
```diff
@@ -237,6 +249,19 @@ export const RichTextInput = ({
...
+    e: NativeSyntheticEvent<OnPressMentionEventInternal>
+  ) => {
+    const { text, attributes } = e.nativeEvent;
+    const parsedAttributes = JSON.parse(attributes) as Record<string, string>;
+
+    onPressMention?.({ text, attributes: parsedAttributes });
+  };
+
+  const handleLinkPress = (e: NativeSyntheticEvent<OnPressLink>) => {
+    onPressLink?.({ url: e.nativeEvent.url });
```
szydlovsky, line 262: I think there could be use cases for passing the link's text as well (especially when the link has different text and url)

exploIF: I think it's not related to this PR, we haven't emitted `text` before too

src/ReactNativeRichTextEditorViewNativeComponent.ts
exploIF: It's no longer `NativeSyntheticEvent` so I removed `event` from the name

src/RichTextInput.tsx
exploIF: We have to modify `onPressMention` on the JS side (parse json string to object), which makes it impossible to emit back `NativeSyntheticEvent`. I've also changed behavior of `OnLinkPress` - so it's compatible with `OnMentionPress`

PR #66, opened 2025-06-04
iOS mentions

szydlovsky:


example/src/App.tsx
```diff
@@ -108,6 +108,7 @@ export default function App() {
 
   const openMentionPopup = () => {
     setIsMentionPopupOpen(true);
+    onMentionChange('');
```
exploIF: Super ultra minor, I would rather call it once we end a mention (so when mention is ended and not opened we have a clear state)

PR #76, opened 2025-06-11
iOS html parser

szydlovsky:


src/ReactNativeRichTextEditorViewNativeComponent.ts
```diff
@@ -147,6 +147,9 @@ export interface NativeProps extends ViewProps {
   fontFamily?: string;
   fontWeight?: string;
   fontStyle?: string;
+
+  // Used for onChangeHtml event performance optimization
+  onChangeHtmlSet: boolean;
```
exploIF: Can we rename it to `isOnChangeHtmlSet`? Current name looks almost like a callback rather than boolean

PR #92, opened 2025-06-16
iOS autoCapitalization

szydlovsky:


ios/ReactNativeRichTextEditorView.mm
```diff
@@ -288,6 +290,31 @@ - (void)updateProps:(Props::Shared const &)props oldProps:(Props::Shared const &
...
+      textView.autocapitalizationType = UITextAutocapitalizationTypeNone;
+    } else if([str isEqualToString: @"sentences"]) {
+      textView.autocapitalizationType = UITextAutocapitalizationTypeSentences;
+    } else if([str isEqualToString: @"words"]) {
+      textView.autocapitalizationType = UITextAutocapitalizationTypeWords;
+    } else if([str isEqualToString: @"characters"]) {
+      textView.autocapitalizationType = UITextAutocapitalizationTypeAllCharacters;
+    } else {
+      // default value
+      textView.autocapitalizationType = UITextAutocapitalizationTypeSentences;
```
exploIF: In Android native part the default is set to "none". Maybe we should be consistent and assign default value only once (on the JS side)?

PR #105, opened 2025-07-14
iOS rich style configs

szydlovsky:


example/ios/Podfile.lock
```diff
@@ -1679,31 +1679,6 @@ PODS:
     - React-perflogger (= 0.79.1)
     - React-utils (= 0.79.1)
   - ReactNativeRichTextEditor (0.1.0):
-    - DoubleConversion
```
exploIF, line 1682: Out of curiosity, where this change comes from?

szydlovsky: there was an unneeded subspec that seemingly added these dependencies just because of its existence

PR #125, opened 2025-08-13
Improve iOS emojis handling

szydlovsky:


ios/ReactNativeRichTextEditorView.mm
```diff
@@ -368,7 +368,11 @@ - (void)updateProps:(Props::Shared const &)props oldProps:(Props::Shared const &
     defaultTypingAttributes[NSFontAttributeName] = [config primaryFont];
     defaultTypingAttributes[NSUnderlineColorAttributeName] = [config primaryColor];
     defaultTypingAttributes[NSStrikethroughColorAttributeName] = [config primaryColor];
-    defaultTypingAttributes[NSParagraphStyleAttributeName] = [[NSParagraphStyle alloc] init];
+    // emoji cutoff fix: apple returns wrong estimates when calculating height of a text with emojis
+    // statically setting lineHeightMultiple to 110% appears to fix the issue without causing other problems
+    NSMutableParagraphStyle *pStyle = [[NSMutableParagraphStyle alloc] init];
+    pStyle.lineHeightMultiple = 1.1;
```
exploIF: How does it work? Does it change actual input size? If so, is it creating inconsistent input height between platforms?

szydlovsky: @exploIF input height between platforms will anyway differ a bit because of different ways they estimate the heights for e.g. emojis. It works here because some of apple's emojis' height is being estimated wrongly. Making the general lineHeight 110% of what it would normally be fixes that issue and keeps the input almost the same size.

PR #139, opened 2025-09-02
autoCapitalize prop typo fix

szydlovsky:


exploIF:
@szydlovsky can we merge it before release?

PR #140, opened 2025-09-03
example app further improvements

szydlovsky:
- improves links handling and makes them both editable and insertable
- adds a modal for manually setting input value via `setValue`
- adds optional live html display with copyable text

https://github.com/user-attachments/assets/290277cc-2525-4049-9dd7-22447200e65a


example/src/components/Button.tsx
```diff
@@ -1,18 +1,34 @@
...
 }
 
-export const Button: FC<ButtonProps> = ({ title, onPress }) => {
+export const Button: FC<ButtonProps> = ({
+  title,
+  onPress,
+  disabled = false,
+  style = {},
+}) => {
+  const [pressed, setPressed] = useState(false);
```
exploIF: Out of curiosity, why do we need this change? Cannot we rely on the previous approach?

example/src/components/HtmlSection.tsx
```diff
@@ -0,0 +1,62 @@
...
+    contentHeight: number
+  ) => {
+    setHtmlTextHeight(Math.min(contentHeight + 20, 190));
+  };
+
+  return (
+    <>
+      <Button
+        title={showHtml ? 'Hide HTML' : 'Show HTML'}
+        onPress={() => setShowHtml(!showHtml)}
```
exploIF: There is a chance that this logic won't work properly (especially when clicking very fast). Due to the fact that state is not updated immediately.
Instead we should do: `setShowHtml((current) => !current)`. That way it guarantees value will be toggled properly

example/src/App.tsx
```diff
@@ -79,12 +81,21 @@ export default function App() {
   const userMention = useUserMention();
   const channelMention = useChannelMention();
 
+  const insideCurrentLink =
+    stylesState.isLink &&
+    currentLink.url.length > 0 &&
+    (currentLink.start !== 0 || currentLink.end !== 0) &&
```
exploIF: just `currentLink.start || currentLink.end`

example/src/App.tsx
```diff
@@ -147,12 +162,35 @@ export default function App() {
...
+    if (!selection || url.length === 0) {
+      closeLinkModal();
+      return;
+    }
+
+    if (insideCurrentLink) {
+      ref.current?.setLink(
+        currentLink.start,
+        currentLink.end,
+        text.length > 0 ? text : url,
```
exploIF: I would move it to separate variable, defined like `newUrl` that can be reused for both conditions

example/src/components/Button.tsx
szydlovsky: Must have been a brainfog 😄

Issue #184, opened 2025-10-03
[Bug] onBlur not called as expected


andrewcoelho:
**Describe the bug**
onBlur doesn't work when tapping on ScrollView which prevents keyboard from being dismissed

**To Reproduce**
Default TextInput component works:

```tsx
export default function Page() {
  const insets = useSafeAreaInsets();

  return (
    <View style={{ flex: 1 }}>
      <ScrollView>
        <View style={{ flex: 1 }} />
      </ScrollView>
      <View style={{ paddingBottom: insets.bottom }}>
        <TextInput
          placeholder="Enter your message"
          onBlur={() => {
            console.log('onBlur'); // fires as expected
          }}
        />
      </View>
    </View>
  );
}
```

EnrichedTextInput does not work:

```tsx
export default function Page() {
  const insets = useSafeAreaInsets();

  return (
    <View style={{ flex: 1 }}>
      <ScrollView>
        <View style={{ flex: 1 }} />
      </ScrollView>
      <View style={{ paddingBottom: insets.bottom }}>
        <EnrichedTextInput
          placeholder="Enter your message"
          onBlur={() => {
            console.log('onBlur'); // never fires
          }}
        />
      </View>
    </View>
  );
}
```

**Expected behavior**
It should behave the same as TextInput so that integrating with ScrollView works as expected

**Screenshots**

**Device (please complete the following information):**

**Additional context**

valerii-starlight:
@andrewcoelho this works:
```
import { ClickOutsideProvider, useClickOutside } from 'react-native-click-outside';

export default function Page() {
  return (
    <View style={{ flex: 1 }}>
      <ClickOutsideProvider>
        <ScrollView>
          <View style={{ flex: 1 }} />
          <RichTextInout />
        </ScrollView>
      </ClickOutsideProvider>
      <RichTextInputToolbar />
    </View>
  );
}

export default function RichTextInput() {
  const containerRef = useClickOutside<View>(() => {
    richTextRef.current?.blur?.();
  });

  return (
    <View ref={containerRef}>
      <EnrichedTextInput
        placeholder="Enter your message"
        onBlur={() => {
          console.log('onBlur'); // fired when clicked in area of ClickOutsideProvider, but ignored when RichTextInputToolbar clicked (which we want)
        }}
      />
    </View>
  )
}
```

[...]

Issue #172, opened 2025-09-29
The Keyboard.dismiss() method does not work properly


kurdev2020:
The Keyboard.dismiss() method does not work properly when called in the onBlur event handler of the EnrichedTextInput component from react-native-enriched. Despite explicitly calling Keyboard.dismiss() in the handleOnBlur callback, the keyboard remains visible and does not dismiss as expected.

## To Reproduce

1. 1.
   Import and use EnrichedTextInput from react-native-enriched in a React Native component
2. 2.
   Set up an onBlur handler that calls Keyboard.dismiss()
3. 3.
   Focus on the EnrichedTextInput to bring up the keyboard
4. 4.
   Tap outside the input or trigger the blur event to lose focus
5. 5.
   Observe that the keyboard does not dismiss despite the Keyboard.dismiss() call

```tsx
import { EnrichedTextInput } from 'react-native-enriched';
import { Keyboard } from 'react-native';

const handleOnBlur = useCallback(() => {
  setIsFocused(false);
  Keyboard.dismiss(); // This doesn't work with EnrichedTextInput
  onBlur?.();
}, [onBlur]);

<EnrichedTextInput
  ref={ref}
  onBlur={handleOnBlur}
  onFocus={handleOnFocus}
  // ... other props
/>
```

## Expected behavior
When the onBlur event is triggered and Keyboard.dismiss() is called, the keyboard should immediately dismiss and hide from the screen, similar to how it works with React Native's standard TextInput component.

Device:
IOS 18.0
- React Native Version : 0.81.1
- react-native-enriched Version : ^0.1.2

## Additional context
- The issue occurs consistently across different devices and platforms
- Standard React Native TextInput components work correctly with Keyboard.dismiss()
- The onBlur callback is being called correctly (verified through logging)
- Other keyboard dismiss calls in the same project work fine when not using EnrichedTextInput
- The component is being used within a complex editor interface with formatting toolbar and AI assistance features
- Project uses Expo SDK 54.0.6 and React Native Reanimated 4.1.0
Workaround attempts:

- Tried calling Keyboard.dismiss() with different timing (setTimeout)
- Attempted to blur the input programmatically before calling dismiss
- None of the workarounds have been successful

exploIF:
Hi, thanks for reporting this. While we still may consider making `Keyboard.dismiss` work, you should be able to achieve desired behavior by calling blur using editor input reference, like below:
```ts
ref.current?.blur()
```

[...]

PR #192, opened 2025-10-06
wip: focus handling

exploIF:
This is still WIP, as I would like to avoid using deprecated API, please [see](https://github.com/react-native-community/discussions-and-proposals/discussions/893#discussioncomment-14606813)

Once it's done it should:
- fix https://github.com/software-mansion/react-native-enriched/issues/184
- address https://github.com/software-mansion/react-native-enriched/issues/172


joeltjames:
I toyed around with this in our app, and it worked well to fix the focus issue (I applied it with `patch-package`). The issue it has now is that when I clicked one of the toolbar buttons to apply Bold, Italic, etc it blurred the input.

PR #202, opened 2025-10-14
fix: unmounting component

exploIF:


exploIF:
Closing, as I found out a better solution in https://github.com/software-mansion/react-native-enriched/pull/205

PR #209, opened 2025-10-21
fix: ios layout issues

exploIF:
Fixes:
- scrolling on iOS - which was unresponsive in some circumstances
- invisible text in 5th line - super weird edge case


szydlovsky (approved):

szydlovsky:
image

PR #223, opened 2025-11-03
fix(iOS): add a check that the href attribute exists in link params

IvanIhnatsiuk:
### Description:

If we pass a default value on iOS, like:

```.html
<html>
 <a>Test link</a>
</html>
```
Then we get a crash because we're trying to get a range of href that doesn't exist

### What's changed:

- added href location check on iOS

### Screenshots:

Before:

https://github.com/user-attachments/assets/a29b75be-b8e5-4626-8789-cd20b763fc7f

After:

https://github.com/user-attachments/assets/e0c62962-2533-4c4f-8b6e-a3fc5499c844

Crash stack trace:

[crash.txt](https://github.com/user-attachments/files/23306147/crash.txt)

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,31 @@
+<!-- Thanks for submitting a pull request! We appreciate you spending the time to work on these changes. Please follow the template so that the reviewers can easily understand what the code changes affect -->
+
+# Summary
+
+Explain the **motivation** for making this change: here are some points to help you:
+
+- What issues does the pull request solve? Please tag them so that they will get automatically closed once PR is merged
+- What is the feature? (if applicable)
+- How did you implement the solution?
+- What areas of the library does it impact?
+
+## Test Plan
+
+Provide **clear steps so another contributor can reproduce the behavior or verify the feature works**.  
+For example:
+
+- Steps to reproduce the bug (if this is a bug fix)
+- Steps to verify the new feature
+- Expected vs actual results
+- Any special conditions or edge cases to test
+
+## Screenshots / Videos
+
+Include any visual proof that helps reviewers understand the change — UI updates, bug reproduction or the result of the fix.
+
+## Compatibility
+
+| OS      | Implemented |
+| ------- | :---------: |
+| iOS     |    ✅❌     |
+| Android |    ✅❌     |
--- a/android/src/main/java/com/swmansion/enriched/EnrichedTextInputView.kt
+++ b/android/src/main/java/com/swmansion/enriched/EnrichedTextInputView.kt
@@ -438,7 +438,7 @@ class EnrichedTextInputView : AppCompatEditText {
   }
 
   private fun verifyStyle(name: String): Boolean {
-    val mergingConfig = EnrichedSpans.mergingConfig[name] ?: return true
+    val mergingConfig = EnrichedSpans.getMergingConfigForStyle(name, htmlStyle) ?: return true
     val conflictingStyles = mergingConfig.conflictingStyles
     val blockingStyles = mergingConfig.blockingStyles
     val isEnabling = spanState?.getStart(name) == null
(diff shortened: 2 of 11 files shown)


szydlovsky:
Thanks for the PR, looks promising! I'll need to check for places where url might be taken for a granted in link and will come back to you

ios/inputParser/InputParser.mm
```diff
@@ -597,10 +597,12 @@ - (NSArray *)getTextAndStylesFromHtml:(NSString *)fixedHtml {
...
-      [styleArr addObject:@([LinkStyle getStyleType])];
-      // cut only the url from the href="..." string
-      NSString *url = [params substringWithRange:NSMakeRange(6, params.length - 7)];
-      stylePair.styleValue = url;
+       [styleArr addObject:@([LinkStyle getStyleType])];
+       NSRange hrefRange = [params rangeOfString:@"href="];
+       if(hrefRange.location != NSNotFound) {
+         NSString *url = [params substringWithRange:NSMakeRange(6, params.length - 7)];
+         stylePair.styleValue = url;
+       }
```
szydlovsky: Okay so indeed not setting `styleValue` (essentially, a `nullptr` in there) breaks some logic. The link attribute isn't properly detected, the html output doesn't recognize the styling as a link and the `onLinkDetected` event won't fire. It's because `getLinkDataAt` method from `LinkStyle` returns no data (`nullptr`) if the url doesn't exist. And that logic needs to stay there because we wouldn't be able to differentiate between Manual and Automatic links this way - query result for one is defined, the other is not. With a `nullptr` url there, both are undefined and thus no link data returned. I'd honestly just set an empty url and everything should work flawlessly;

```suggestion
       stylePair.styleValue = @"";
       if(hrefRange.location != NSNotFound) {
         NSString *url = [params substringWithRange:NSMakeRange(6, params.length - 7)];
         stylePair.styleValue = url;
       }
```

szydlovsky (changes requested):

szydlovsky:
And one more thing @IvanIhnatsiuk; if you want the emitted html to not have the `href` attribute, you have to modify `tagContentForStyle` method in the InputParser as well and add a check to not add it when `url` is empty

IvanIhnatsiuk:
@szydlovsky, from my point of view, if we generate a link, then the href attribute should always be there (ideally). This bug is addressed to only to prevent a crash when we get the link from the initial value or manually call setValue with HTML content

szydlovsky:
Okay @IvanIhnatsiuk this sounds reasonable. Let me test things out for the last time on Monday and we can proceed with the PR

ios/inputParser/InputParser.mm
```diff
@@ -597,10 +597,13 @@ - (NSArray *)getTextAndStylesFromHtml:(NSString *)fixedHtml {
...
-      // cut only the url from the href="..." string
-      NSString *url = [params substringWithRange:NSMakeRange(6, params.length - 7)];
-      stylePair.styleValue = url;
+       [styleArr addObject:@([LinkStyle getStyleType])];
+       NSRange hrefRange = [params rangeOfString:@"href="];
+       stylePair.styleValue = @"";
+       if(hrefRange.location != NSNotFound) {
+         NSString *url = [params substringWithRange:NSMakeRange(6, params.length - 7)];
+         stylePair.styleValue = url;
+       }
```
szydlovsky: Super tiny but the new code is one space too far to the right. Could you please remove them?

szydlovsky (commented):
Looks and works good. Just this tiny one cosmetic change and we can merge it.

szydlovsky:
Hey @IvanIhnatsiuk, we spoke with @exploIF a little about the changes in here and created a counter-offer PR, [see here](https://github.com/software-mansion/react-native-enriched/pull/260). 
The issue with this approach is the library never produces `<a>` tags with no `href` attribute, so there is no reason to consider such tag a proper link. Instead, the PR I made just gets rid of the crash and doesn't recognize links if there is no `href` or an empty one. Is that approach still okay with your use-case? Was it to just get rid of a crash or are the empty links something you need?

IvanIhnatsiuk:
@IvanIhnatsiuk yes, it sounds good to me 🙂

szydlovsky:
Then closing the PR and merging the other option, nonetheless thanks for the idea!

PR #226, opened 2025-11-05
feat: add scrollEnabled prop

IvanIhnatsiuk:
### Description

The scrollEnabled option has been added to prohibit scrolling of content inside a text field. In some cases, this can be useful (as in our application) when we want to show multiple EnrichedTextInput and a single ScrollView.

### Screenshots

Android: 

https://github.com/user-attachments/assets/f8c09315-ffe1-4a53-b9c8-e07e64a70884

iOS:

https://github.com/user-attachments/assets/0e61ef2a-a15a-4d5a-ac9b-9b33651bceab

### DIFF
--- a/android/src/main/java/com/swmansion/enriched/EnrichedTextInputView.kt
+++ b/android/src/main/java/com/swmansion/enriched/EnrichedTextInputView.kt
@@ -71,6 +71,8 @@ class EnrichedTextInputView : AppCompatEditText {
   private var fontFamily: String? = null
   private var fontStyle: Int = ReactConstants.UNSET
   private var fontWeight: Int = ReactConstants.UNSET
+  private var defaultValue: CharSequence? = null
+  private var defaultValueDirty: Boolean = false
 
   private var inputMethodManager: InputMethodManager? = null
 
@@ -369,7 +371,24 @@ class EnrichedTextInputView : AppCompatEditText {
     return false
   }
 
-  fun updateTypeface() {
+  fun afterUpdateTransaction() {
+    updateTypeface()
+    updateDefaultValue()
+  }
+
+  fun setDefaultValue(value: CharSequence?) {
+    defaultValue = value
+    defaultValueDirty = true
+  }
+
+  private fun updateDefaultValue() {
+    if (!defaultValueDirty) return
+
+    defaultValueDirty = false
+    setValue(defaultValue ?: "")
+  }
+
+  private fun updateTypeface() {
     if (!typefaceDirty) return
     typefaceDirty = false
 
--- a/ios/EnrichedTextInputView.mm
+++ b/ios/EnrichedTextInputView.mm
@@ -605,6 +605,9 @@ Class<RCTComponentViewProtocol> EnrichedTextInputViewCls(void) {
   // style updates are emitted only if something differs from the previously active styles
   BOOL updateNeeded = NO;
   
+  // active styles are kept in a separate set until we're sure they can be emitted
+  NSMutableSet *newActiveStyles = [_activeStyles mutableCopy];
+  
   // data for onLinkDetected event
   LinkData *detectedLinkData;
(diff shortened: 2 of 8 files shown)


ios/EnrichedTextInputView.mm
```diff
@@ -862,6 +866,10 @@ - (void)tryEmittingOnChangeHtmlEvent {
   }
 }
 
+-(void)setScrollEnabled:(bool)scrollEnabled {
+  [textView setScrollEnabled: scrollEnabled];
+}
+
```
szydlovsky: Honestly, I don't see why we need that as a separate method if it essentially runs only a single line of code. The file is already pretty massive and such small methods just make the code clarity even worse. So I suggest just calling `[textView setScrollEnabled]` where the prop changes.

android/src/main/java/com/swmansion/enriched/EnrichedTextInputView.kt
```diff
@@ -54,6 +54,7 @@ class EnrichedTextInputView : AppCompatEditText {
   val parametrizedStyles: ParametrizedStyles? = ParametrizedStyles(this)
   var isDuringTransaction: Boolean = false
   var isRemovingMany: Boolean = false
+  var scrollEnabled = true
```
exploIF: Super minor, just for consistency with above declarations, would be good to explicitly type it

exploIF (commented):
@IvanIhnatsiuk, thanks for the contribution. It looks good to me. If you can fix those two tiny comments, I will merge it and we can include it in the upcoming release

ios/inputParser/InputParser.mm
```diff
@@ -597,10 +597,12 @@ - (NSArray *)getTextAndStylesFromHtml:(NSString *)fixedHtml {
     } else if([tagName isEqualToString:@"code"]) {
       [styleArr addObject:@([InlineCodeStyle getStyleType])];
     } else if([tagName isEqualToString:@"a"]) {
-      [styleArr addObject:@([LinkStyle getStyleType])];
-      // cut only the url from the href="..." string
-      NSString *url = [params substringWithRange:NSMakeRange(6, params.length - 7)];
-      stylePair.styleValue = url;
+       [styleArr addObject:@([LinkStyle getStyleType])];
```
exploIF: I think this change shouldn't land in this PR, I guess it's some leftover coming from another one

IvanIhnatsiuk: Yes, my bad

ios/EnrichedTextInputView.mm
```diff
@@ -357,6 +357,10 @@ - (void)updateProps:(Props::Shared const &)props oldProps:(Props::Shared const &
     stylePropChanged = YES;
   }
   
+  if(newViewProps.scrollEnabled != oldViewProps.scrollEnabled || textView.scrollEnabled != newViewProps.scrollEnabled) {
+    [textView setScrollEnabled: newViewProps.scrollEnabled];
```
tomekzaw:
```suggestion
    [textView setScrollEnabled:newViewProps.scrollEnabled];
```

ios/inputParser/InputParser.mm
```diff
@@ -597,10 +597,12 @@ - (NSArray *)getTextAndStylesFromHtml:(NSString *)fixedHtml {
...
+       NSRange hrefRange = [params rangeOfString:@"href="];
+       if(hrefRange.location != NSNotFound) {
```
tomekzaw:
```suggestion
       if (hrefRange.location != NSNotFound) {
```
We'll need some formatter/linter and setup GitHub Actions.

exploIF:
Hi @IvanIhnatsiuk, if you could remove unnecessary parser changes, we can ship this feature this week

exploIF (approved):
Thank you 🔥

PR #228, opened 2025-11-07
feat: add ability to get html value imperatively

IvanIhnatsiuk:
### Description

Added the ability to get the HTML value from the reference of the EnrichedTextInput itself. This way, we don't need to parse the text with every change, which improves speed.

### How to test?

1. Open app
2. Write any text in the input
3. Press on the "Get input's HTML value"
4. Observe value in the logs

### Videos

iOS:

https://github.com/user-attachments/assets/1aae8185-00c9-49b2-8ea8-d51873ee75d8

Android:

https://github.com/user-attachments/assets/fb96f012-af75-4754-a26c-6b1960744006


android/src/main/java/com/swmansion/enriched/EnrichedTextInputModule.kt
```diff
@@ -0,0 +1,33 @@
+package com.swmansion.enriched
+
+import com.facebook.react.bridge.ReactApplicationContext
+import com.facebook.react.bridge.UiThreadUtil
+import com.facebook.react.module.annotations.ReactModule
+import com.facebook.react.uimanager.UIManagerHelper
+import com.swmansion.enriched.utils.EnrichedParser
+
+@ReactModule(name = EnrichedTextInputModule.NAME)
+class EnrichedTextInputModule(val reactContext: ReactApplicationContext) :
```
tomekzaw, line 10: I don't like the fact that this PR introduces a native module. Why exactly is that needed?

ios/EnrichedTextInputModule.mm
```diff
@@ -0,0 +1,46 @@
...
+@implementation EnrichedTextInputModule
+
+- (std::shared_ptr<facebook::react::TurboModule>)getTurboModule:(const facebook::react::ObjCTurboModule::InitParams &)params {
+  return std::make_shared<facebook::react::NativeEnrichedTextInputModuleSpecJSI>(params);
+}
+
+- (nonnull NSString *)getHTMLValue:(NSInteger)inputTag {
+  __block NSString *value = @"";
+
+  dispatch_sync(dispatch_get_main_queue(), ^{
```
tomekzaw, line 31: I don't like the fact that we need `dispatch_sync` here. It blocks the current thread.

android/src/main/new_arch/RNEnrichedTextInputViewSpec.h
```diff
@@ -23,4 +23,4 @@ namespace facebook::react {
 JSI_EXPORT
 std::shared_ptr<TurboModule> RNEnrichedTextInputViewSpec_ModuleProvider(const std::string &moduleName, const JavaTurboModule::InitParams &params);
 
-} // namespace facebook::react
+} // namespace facebook::react
```
tomekzaw: Please restore the trailing newline
```suggestion
} // namespace facebook::react

```

android/src/main/java/com/swmansion/enriched/EnrichedTextInputModule.kt
IvanIhnatsiuk: This is necessary because native commands do not allow you to return any values. If you have any other ideas on how the second option in [this](https://github.com/software-mansion/react-native-enriched?tab=readme-ov-file#future-plans) list can be implemented without a module, I would be happy to change the current implementation 🙂

ios/EnrichedTextInputModule.mm
IvanIhnatsiuk: Okay, I have a compromise approach. We can keep this method and add another one

1. The new method will work like this: On the main or background thread, we can copy the editable/attributed text value and then parse it on the background thread. Since I think the performant work is done in the parser
2. We can add a **BOOOOLD** note that the sync method might block the UI thread, and it's not recommended for usage or use it with caution.

It would be nice to hear your point of view 🙂

IvanIhnatsiuk:
Closed due to https://github.com/facebook/react-native/issues/50800#issuecomment-2841429695

PR #233, opened 2025-11-09
fix(Android): recalculating layout after changing html styles

IvanIhnatsiuk:
### Description

If we change the HTML styles on Android, they will not be applied in any way until we first enter any styles. We also got a bug with the initial defaultValue, because everything started to depend on who was the first to be counted (it depends on who was the first to be passed from JS to the component). 
Thus, I added the calculation of the lead when changing the HTML style on Android.

### Videos

Before:

https://github.com/user-attachments/assets/0ccaae0c-d5e0-4270-b000-534260803ea3

https://github.com/user-attachments/assets/e89536be-c787-477a-8aba-fba2e0a6919d

After:

https://github.com/user-attachments/assets/6d27d9f7-cbbf-40f3-9010-557298a3eaf7

### DIFF
--- a/README.md
+++ b/README.md
@@ -238,14 +238,10 @@ See the [API Reference](docs/API_REFERENCE.md) for a detailed overview of all th
 ## Known limitations
 
 - Only one level of lists is supported. We currently do not support nested lists.
-- Inline images are supported only on Android.
-- Codeblocks are supported only on Android.
 - iOS headings can't have the same `fontSize` in their config as input's `fontSize`. Doing so results in incorrect headings behavior.
 
 ## Future Plans
 
-- Adding Codeblocks and Inline Images to iOS input.
-- Making some optimizations around `onChangeHtml` event, maybe some imperative API to get the HTML output.
 - Creating `EnrichedText` text component that supports our HTML output format with all additional interactions like pressing links or mentions.
 - Adding API for custom link detection regex.
 - Web library implementation via `react-native-web`.
--- a/docs/API_REFERENCE.md
+++ b/docs/API_REFERENCE.md
@@ -47,6 +47,10 @@ If `false`, text is not editable.
 |--------|---------------|----------|
 | `bool` | `true`        | Both     |
 
+> [!NOTE]
+> Setting `editable` to `false` will disable all user interactions with the input.
+> However, some programmatic changes (like toggling styles or changing value imperatively) via ref methods will still work.
+
 ### `htmlStyle`
 
 A prop for customizing styles appearances.
@@ -82,6 +86,11 @@ interface OnChangeHtmlEvent {
   value: string;
 }
 ```
+
+> [!TIP]
+> Specifying `onChangeHtml` may have performance implications, especially with large documents, as it requires continuous HTML parsing.
+> If you only need the HTML content at specific moments (e.g., when saving), consider using the [`getHTML`](#gethtml) ref method instead.
+> When `onChangeHtml` is not provided, the component optimizes performance by avoiding unnecessary HTML parsing.
 
 - `value` is the new HTML.
 


exploIF:
@IvanIhnatsiuk thanks for this, as this PR can have some impact on apps which are already using library, I would like to discuss this a little bit more. It would be super cool if you can catch up with me on discord (username `exploif`) or on x.com (username is linked in my profile )

android/src/main/java/com/swmansion/enriched/EnrichedTextInputView.kt
```diff
@@ -577,6 +583,27 @@ class EnrichedTextInputView : AppCompatEditText {
     }
   }
 
+  private fun reapplyTextWithNewStyles() {
```
exploIF: I don't exactly get main idea here. Why do we have to parse text twice just to update span styles? Isn't it enough to iterate over spans and update their style?

android/src/main/java/com/swmansion/enriched/EnrichedTextInputView.kt
```diff
@@ -577,6 +583,27 @@ class EnrichedTextInputView : AppCompatEditText {
...
+    val currentText = text
+    if (currentText != null && currentText.isNotEmpty()) {
+      runAsATransaction {
+        val currentHtml = EnrichedParser.toHtml(currentText as Spannable)
+        val newText = parseText(currentHtml)
+
+        val currentSelectionStart = selectionStart
+        val currentSelectionEnd = selectionEnd
+
+        setText(newText)
```
exploIF: This is incorrect, for manually setting text we `setValue`, which also handles applying necessary listeners

example/src/App.tsx
```diff
@@ -304,8 +312,9 @@ export default function App() {
             ref={ref}
             mentionIndicators={['@', '#']}
             style={styles.editorInput}
-            htmlStyle={htmlStyle}
+            htmlStyle={currentHtmlStyle}
             placeholder="Type something here..."
+            defaultValue="<html><h1>test</h1><ul><li>test <a href='test' /></li></ul></html>"
```
exploIF: Let's remove this

example/src/App.tsx
```diff
@@ -84,6 +84,8 @@ export default function App() {
   const [isImageModalOpen, setIsImageModalOpen] = useState(false);
   const [isValueModalOpen, setIsValueModalOpen] = useState(false);
   const [currentHtml, setCurrentHtml] = useState('');
+  const [currentHtmlStyle, setCurrentHtmlStyle] =
+    useState<HtmlStyle>(htmlStyle);
```
exploIF: There is really no need to store whole styles object inside state. Instead, we can store some kind of flag (boolean or one of union of strings) which will be responsible for determining which style should be used

example/src/App.tsx
```diff
@@ -379,6 +393,64 @@ export default function App() {
   );
 }
 
+const anotherHtmlStyle: HtmlStyle = {
```
exploIF: Nit, if it's called `anotherHtmlStyle` it should be rendered below base `htmlStyle`

android/src/main/java/com/swmansion/enriched/EnrichedTextInputView.kt
```diff
@@ -578,6 +591,37 @@ class EnrichedTextInputView : AppCompatEditText {
...
+          val end = getSpanEnd(span)
+          val flags = getSpanFlags(span)
+
+          if ((span is EnrichedH1Span && shouldRemoveBoldSpanFromH1Span) || (span is EnrichedH2Span && shouldRemoveBoldSpanFromH2Span) || (span is EnrichedH3Span && shouldRemoveBoldSpanFromH3Span)) {
+            removeStyle(EnrichedSpans.BOLD, start, end)
+          }
+
+          removeSpan(span)
+          val newSpan = span.rebuildWithStyle(htmlStyle)
+          setSpan(newSpan, start, end, flags)
```
exploIF: Maybe we should check start and end one more time after calling `removeStyle`? This method can potentially modify text content (zero-width space manipulations), so it's better to validate that again

android/src/main/java/com/swmansion/enriched/EnrichedTextInputView.kt
```diff
@@ -578,6 +591,37 @@ class EnrichedTextInputView : AppCompatEditText {
...
+      val spans = spannable.getSpans(0, spannable.length, EnrichedSpan::class.java)
+      for (span in spans) {
+        if(!span.dependsOnHtmlStyle) {
+          continue
+        }
+
+        spannable.apply {
+          val start = getSpanStart(span)
+          val end = getSpanEnd(span)
+          val flags = getSpanFlags(span)
```
exploIF: Every time we create a new span based on `getSpanStart` or `getSpanEnd` we should check if they are different than `-1`, as different operation may already modify such span

android/src/main/java/com/swmansion/enriched/EnrichedTextInputView.kt
```diff
@@ -59,6 +65,13 @@ class EnrichedTextInputView : AppCompatEditText {
 
   val mentionHandler: MentionHandler? = MentionHandler(this)
   var htmlStyle: HtmlStyle = HtmlStyle(this, null)
+    set(value) {
+        if (field != value) {
```
exploIF, line 68: Currently it's not safe operation. Instead we should implement `equals` and `hashCode` methods for `HtmlStyle` class. This will ensure comparison is safe and reliable

example/src/App.tsx
```diff
@@ -379,6 +393,64 @@ export default function App() {
...
+  h1: {
+    fontSize: 100,
```
exploIF:
```suggestion
    fontSize: 52,
```

example/src/App.tsx
```diff
@@ -379,6 +393,64 @@ export default function App() {
...
+    bold: false,
+  },
+  h2: {
+    fontSize: 50,
```
exploIF:
```suggestion
    fontSize: 44,
```

android/src/main/java/com/swmansion/enriched/EnrichedTextInputView.kt
```diff
@@ -578,6 +591,37 @@ class EnrichedTextInputView : AppCompatEditText {
...
+        }
+
+        spannable.apply {
+          val start = getSpanStart(span)
+          val end = getSpanEnd(span)
+          val flags = getSpanFlags(span)
+
+          if ((span is EnrichedH1Span && shouldRemoveBoldSpanFromH1Span) || (span is EnrichedH2Span && shouldRemoveBoldSpanFromH2Span) || (span is EnrichedH3Span && shouldRemoveBoldSpanFromH3Span)) {
+            removeStyle(EnrichedSpans.BOLD, start, end)
+          }
```
exploIF: We remove style, but we don't emit new `onChangeState` event with updated styles state. 

Suggestion:
- `removeStyle` should return bool whether style has been removed or not
- we track if at least one style has been removed
- if it has been removed we have to emit span state update (maybe via selection.validateStyles)

exploIF (commented):
I really like this approach, we just need couple minor adjustments

example/src/App.tsx
IvanIhnatsiuk: Sorry, was added to test that typing works well

android/src/main/java/com/swmansion/enriched/EnrichedTextInputView.kt
IvanIhnatsiuk: Hmm, in this case I have a question

Should we determine this as a transaction since it might modify existing text, and if any other changes happen? Let's say dev is setting a new value + updating htmlStyles, or during updating html styles, dev called setValue -> crash

Am I right?

android/src/main/java/com/swmansion/enriched/EnrichedTextInputView.kt
IvanIhnatsiuk: Good catch

android/src/main/java/com/swmansion/enriched/EnrichedTextInputView.kt
exploIF: In general we use `runAsTransaction` every time style modification happens because of your code rather than by user action. Some of the styles processing is handled based on selection or text change listeners, and we don't want to fire them when processing manually. For your case, I believe we should use `runAsTransaction`, but we don't have to nest one transaction inside different one, so the one that is already implemented should be enough

kacperzolkiewski:
Hi @IvanIhnatsiuk, 
I found some bug during tests, that content disappears until I focus the input:

https://github.com/user-attachments/assets/c83bf044-4cef-41ec-bb34-3a3b49123e01

IvanIhnatsiuk:
@kacperzolkiewski I don't think it disappears, it looks like the layout is not updated 🤔

IvanIhnatsiuk:
@kacperzolkiewski could you check on the latest branch version?

kacperzolkiewski:
@IvanIhnatsiuk unfortunately still the same, it looks like it is not scrolling to the top:

https://github.com/user-attachments/assets/c0eb8329-e6e3-44ea-a417-8a1576593906

IvanIhnatsiuk:
> @IvanIhnatsiuk unfortunately still the same, it looks like it is not scrolling to the top:
> 
>  Screen.Recording.2025-12-12.at.13.42.12.mov

could you retest on the latest branch version?

kacperzolkiewski:
Works fine now 👏

PR #242, opened 2025-11-13
feat: add onKeyPress callback

IvanIhnatsiuk:
### Description

I added the `onKeyPress` implementation so that we could track which key on the keyboard was pressed on the JS side.

The implementation is almost the same as in ReactEditText itself (the TextInput component in react-native).

This way, we get the same events as in react-native itself.

### Videos:

Android:

https://github.com/user-attachments/assets/b241e19d-0dd8-4eef-81eb-0b7ef09471eb

iOS:

https://github.com/user-attachments/assets/27075e3f-50f9-4234-9310-4dee4732f63a

### DIFF
--- a/docs/API_REFERENCE.md
+++ b/docs/API_REFERENCE.md
@@ -403,9 +403,9 @@ export interface OnKeyPressEvent {
 }
 ```
 
-| Type                                                     | Default Value | Platform |
-|----------------------------------------------------------|---------------|----------|
-| `(event: NativeSyntheticEvent<OnKeyPressEvent>) => void` | -             | Both     |
+| Type                                                     | Platform |
+|----------------------------------------------------------|----------|
+| `(event: NativeSyntheticEvent<OnKeyPressEvent>) => void` | Both     |
 
 ### `placeholder`
 


IvanIhnatsiuk:
@exploIF , @szydlovsky kind reminder 🙂

android/src/main/java/com/swmansion/enriched/EnrichedTextInputConnectionWrapper.kt
```diff
@@ -0,0 +1,137 @@
+package com.swmansion.enriched
+
+import android.view.KeyEvent
+import android.view.inputmethod.InputConnection
+import android.view.inputmethod.InputConnectionWrapper
+import com.facebook.react.bridge.ReactContext
+import com.facebook.react.uimanager.UIManagerHelper
+import com.swmansion.enriched.events.OnInputKeyPressEvent
+
+class EnrichedTextInputConnectionWrapper(
```
exploIF, line 13: Minor one, could you just add a comment with link to the react-native implementation? That way it's easier for contributors to see what's going on and track potential changes/issues

src/EnrichedTextInputNativeComponent.ts
```diff
@@ -256,6 +260,7 @@ export interface NativeProps extends ViewProps {
   onMention?: DirectEventHandler<OnMentionEvent>;
   onChangeSelection?: DirectEventHandler<OnChangeSelectionEvent>;
   onRequestHtmlResult?: DirectEventHandler<OnRequestHtmlResultEvent>;
+  onInputKeyPress?: DirectEventHandler<OnKeyPressEvent>;
```
exploIF, line 263: Minor, maybe just `onKeyPress`?

ios/EnrichedTextInputView.h
```diff
@@ -40,6 +40,7 @@ NS_ASSUME_NONNULL_BEGIN
 - (BOOL)handleStyleBlocksAndConflicts:(StyleType)type range:(NSRange)range;
 - (NSArray<NSNumber *> *)getPresentStyleTypesFrom:(NSArray<NSNumber *> *)types
                                             range:(NSRange)range;
+- (void)emitOnKeyPressEvent:(NSString *)key;
```
szydlovsky: Minor; can you move it near other `emit...` public methods?

ios/EnrichedTextInputView.mm
```diff
@@ -1735,10 +1735,42 @@ - (void)textViewDidEndEditing:(UITextView *)textView {
   }
 }
 
+- (void)emitOnKeyPressEvent:(NSString *)key {
+  auto emitter = [self getEventEmitter];
+  if (emitter != nullptr) {
+    emitter->onInputKeyPress({.key = [key toCppString]});
+  }
+}
```
szydlovsky: Let's move that method to `Native commands and events` section

ios/EnrichedTextInputView.mm
```diff
@@ -1735,10 +1735,42 @@ - (void)textViewDidEndEditing:(UITextView *)textView {
...
+
+- (void)handleKeyPressInRange:(NSString *)text range:(NSRange)range {
+  NSString *key = nil;
+
+  // if text was pasted we should not emit key press event
+  if (self->textView.textWasPasted) {
+    self->textView.textWasPasted = NO;
```
szydlovsky: No need to use `self` here

ios/inputTextView/InputTextView.h
```diff
@@ -3,4 +3,5 @@
 
 @interface InputTextView : UITextView
 @property(nonatomic, weak) id input;
+@property(nonatomic) BOOL textWasPasted;
```
szydlovsky: The logic wtih `textWasPasted` doesn't work properly. In the scenario:
1. I paste something, `textWasPasted` is set to YES.
2. `shouldChangeTextInRange` doesn't run, so the `textWasPasted` isn't reset to `NO`
3. Now I press any key, now the `textWasPasted` is reset but the key press isn't emitted.
4. Only the next key press works.

Of course it needs to be checked, but it seems to me that pasting doesn't cause `shouldChangeTextInRange` to run, because pasting programatically inserts the text. So I think the whole `textWasPasted` logic is not needed and may be just removed.

ios/inputTextView/InputTextView.mm
```diff
@@ -153,4 +155,17 @@ - (void)cut:(id)sender {
...
+- (void)deleteBackward {
+  EnrichedTextInputView *typedInput = (EnrichedTextInputView *)_input;
+  if (typedInput != nullptr) {
+    [typedInput emitOnKeyPressEvent:@"Backspace"];
+  }
+
+  [super deleteBackward];
+
+  if (typedInput != nullptr) {
+    [typedInput anyTextMayHaveBeenModified];
```
szydlovsky: What is that for in here?

src/EnrichedTextInputNativeComponent.ts
IvanIhnatsiuk: It will conflict with RN

ios/inputTextView/InputTextView.mm
IvanIhnatsiuk: 👀

szydlovsky (approved):
iOS works properly - and really little code got added which is two birds with one stone 😄

exploIF (approved):

PR #253, opened 2025-11-17
feat: add support for codeblock on iOS

kacperzolkiewski:
## Summary
Add support for `codeblock`.

## Test Plan
Run Example app and play with `codeblock`

## Screenshots
image

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅ (already handled that way)     |

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,11 +101,11 @@ export default function App() {
     <View style={styles.container}>
       <EnrichedTextInput
         ref={ref}
-        onChangeState={(e) => setStylesState(e.nativeEvent)}
+        onChangeState={e => setStylesState(e.nativeEvent)}
         style={styles.input}
       />
       <Button
-        title="Toggle bold"
+        title={stylesState?.isBold ? 'Unbold' : 'Bold'}
         color={stylesState?.isBold ? 'green' : 'gray'}
         onPress={() => ref.current?.toggleBold()}
       />
--- a/ios/styles/LinkStyle.mm
+++ b/ios/styles/LinkStyle.mm
@@ -169,7 +169,6 @@ static NSString *const AutomaticLinkAttributeName = @"AutomaticLinkAttributeName
   }
   
   [self manageLinkTypingAttributes];
-  [_input anyTextMayHaveBeenModified];
 }
 
 // get exact link data at the given location if it exists
@@ -226,23 +225,22 @@ static NSString *const AutomaticLinkAttributeName = @"AutomaticLinkAttributeName
     }
   }
   
-  NSString *manualUrl = [_input->textView.textStorage
+  NSString *manualLink = [_input->textView.textStorage
     attribute:ManualLinkAttributeName
     atIndex:searchLocation
     longestEffectiveRange: &manualLinkRange
     inRange:inputRange
   ];
-  [_input->textView.textStorage
+  NSString *automaticLink = [_input->textView.textStorage
     attribute:AutomaticLinkAttributeName
     atIndex:searchLocation
     longestEffectiveRange: &automaticLinkRange
     inRange:inputRange
   ];
   
-  return
-    manualLink == nullptr
-      ? automaticLink == nullptr ? NSMakeRange(0,0) : automaticLinkRange
-      : manualLinkRange;
+  return manualLink == nullptr
+    ? automaticLink == nullptr ? NSMakeRange(0, 0) : automaticLinkRange
+    : manualLinkRange;
 }
 
 - (void)manageLinkTypingAttributes {
@@ -360,10 +358,9 @@ static NSString *const AutomaticLinkAttributeName = @"AutomaticLinkAttributeName
     }
     if(addStyle) {
       [self addLink:word url:regexPassedUrl range:wordRange manual:NO];
+      // emit onLinkDetected if style was added
+      [_input emitOnLinkDetectedEvent:word url:regexPassedUrl range:wordRange];
     }
-  
-    // emit onLinkDetected
-    [_input emitOnLinkDetectedEvent:word url:regexPassedUrl range:wordRange];
   }
 }
 
--- a/android/src/main/java/com/swmansion/enriched/EnrichedTextInputViewManager.kt
+++ b/android/src/main/java/com/swmansion/enriched/EnrichedTextInputViewManager.kt
@@ -279,7 +279,8 @@ class EnrichedTextInputViewManager : SimpleViewManager<EnrichedTextInputView>(),
     heightMode: YogaMeasureMode?,
     attachmentsPositions: FloatArray?
   ): Long {
-    return MeasurementStore.getMeasureById(localData?.getInt("viewTag"), width)
+    val id = localData?.getInt("viewTag")
+    return MeasurementStore.getMeasureById(context, id, width, props)
   }
 
   companion object {
--- a/android/src/main/java/com/swmansion/enriched/MeasurementStore.kt
+++ b/android/src/main/java/com/swmansion/enriched/MeasurementStore.kt
@@ -1,14 +1,23 @@
 package com.swmansion.enriched
 
+import android.content.Context
 import android.graphics.Typeface
 import android.graphics.text.LineBreaker
 import android.os.Build
 import android.text.Spannable
 import android.text.StaticLayout
 import android.text.TextPaint
+import android.util.Log
+import com.facebook.react.bridge.ReadableMap
 import com.facebook.react.uimanager.PixelUtil
+import com.facebook.react.views.text.ReactTypefaceUtils.applyStyles
+import com.facebook.react.views.text.ReactTypefaceUtils.parseFontStyle
+import com.facebook.react.views.text.ReactTypefaceUtils.parseFontWeight
 import com.facebook.yoga.YogaMeasureOutput
+import com.swmansion.enriched.styles.HtmlStyle
+import com.swmansion.enriched.utils.EnrichedParser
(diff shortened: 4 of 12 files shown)


ios/utils/ParagraphAttributesUtils.mm
```diff
@@ -53,6 +54,11 @@ + (BOOL)handleBackspaceInRange:(NSRange)range replacementText:(NSString *)text i
       [bqStyle addAttributes:NSMakeRange(range.location, 0)];
       return YES;
     }
+    if([cbStyle detectStyle:nonNewlineRange]) {
+      [TextInsertionUtils replaceText:text at:range additionalAttributes:nullptr input:typedInput withSelection:YES];
+      [cbStyle addAttributes:NSMakeRange(range.location, 0)];
+      return YES;
+    }
```
szydlovsky: We could simplify the code as all of these checks essentially do the exact same thing;
(everything since the `if` statement is to be changed)
```suggestion
    // for lists, quotes and codeblocks we want to remove the characters but keep the attributes so that a zero width space appears here
    // so we do the removing manually and reapply attributes
    NSArray *handledStyles = @[ulStyle, olStyle, bqStyle, cbStyle];
    for(id<BaseStyleProtocol> style in handledStyles) {
      if([style detectStyle:nonNewlineRange]) {
        [TextInsertionUtils replaceText:text at:range additionalAttributes:nullptr input:typedInput withSelection:YES];
        return YES;
      }
    }
```

ios/styles/H1Style.mm
```diff
@@ -3,6 +3,7 @@
 
 @implementation H1Style
 + (StyleType)getStyleType { return H1; }
++ (BOOL)isParagraphStyle { return NO; }
```
szydlovsky: It is a paragraph style.

ios/styles/H2Style.mm
```diff
@@ -3,6 +3,7 @@
 
 @implementation H2Style
 + (StyleType)getStyleType { return H2; }
++ (BOOL)isParagraphStyle { return NO; }
```
szydlovsky: It is a paragraph style.

ios/styles/H3Style.mm
```diff
@@ -3,6 +3,7 @@
 
 @implementation H3Style
 + (StyleType)getStyleType { return H3; }
++ (BOOL)isParagraphStyle { return NO; }
```
szydlovsky: It is a paragraph style.

ios/styles/LinkStyle.mm
```diff
@@ -294,6 +297,11 @@ - (void)handleAutomaticLinks:(NSString *)word inRange:(NSRange)wordRange {
     return;
   }
   
+  // we don't recognize links among codeblock
```
szydlovsky:
```suggestion
  // we don't recognize links in codeblocks
```

ios/utils/ParagraphAttributesUtils.mm
```diff
@@ -64,4 +70,57 @@ + (BOOL)handleBackspaceInRange:(NSRange)range replacementText:(NSString *)text i
   return NO;
 }
 
++ (BOOL)handleNewlineBackspaceInRange:(NSRange)range replacementText:(NSString *)text input:(id)input {
```
szydlovsky, line 77: Let's add some verbose comment what that function does and why it needs to run as a last of the callbacks

ios/utils/ZeroWidthSpaceUtils.mm
```diff
@@ -42,9 +42,10 @@ + (void)removeSpacesIfNeededinInput:(EnrichedTextInputView *)input {
       UnorderedListStyle *ulStyle = input->stylesDict[@([UnorderedListStyle getStyleType])];
       OrderedListStyle *olStyle = input->stylesDict[@([OrderedListStyle getStyleType])];
       BlockQuoteStyle *bqStyle = input->stylesDict[@([BlockQuoteStyle getStyleType])];
+      CodeBlockStyle *cbStyle = input->stylesDict[@([CodeBlockStyle getStyleType])];
       
       // zero width spaces with no lists/blockquote styles on them get removed
```
szydlovsky:
```suggestion
      // zero width spaces with no lists/blockquotes/codeblocks on them get removed
```

ios/styles/CodeBlockStyle.mm
```diff
@@ -0,0 +1,222 @@
...
+      [self removeAttributes:paragraphRange];
+      return YES;
+    }
+  }
+  return NO;
+}
+
+- (BOOL)styleCondition:(id _Nullable)value :(NSRange)range {
+  NSParagraphStyle *paragraph = (NSParagraphStyle *)value;
+  return paragraph != nullptr && paragraph.textLists.count == 1 && [paragraph.textLists.firstObject.markerFormat isEqual:@"codeblock"];
```
szydlovsky: I know it works but it's generally safer to compare just the contents (`isEqualToString`) since `NSTextListMarkerFormat` is a typedef for `NSString*`. And we don't use any defined markerFormat but rather compare to a locally created string.
```suggestion
  return paragraph != nullptr && paragraph.textLists.count == 1 && [paragraph.textLists.firstObject.markerFormat isEqualToString:@"codeblock"];
```

ios/utils/ParagraphAttributesUtils.mm
kacperzolkiewski: fixed

ios/utils/ParagraphAttributesUtils.mm
kacperzolkiewski: I add comment with explanation. What do you think?

ios/styles/LinkStyle.mm
```diff
@@ -239,7 +239,7 @@ - (NSRange)getFullLinkRangeAt:(NSUInteger)location {
     inRange:inputRange
   ];
   
-  return manualLinkRange.length == 0 ? automaticLinkRange : manualLinkRange;
+  return manualUrl == nil ? automaticLinkRange : manualLinkRange;
```
szydlovsky: For the sake of safety I'd also assign `automaticUrl` to the automatic link range search result and do something like;
```suggestion
  return
    manualLink == nullptr
      ? automaticLink == nullptr ? NSMakeRange(0,0) : automaticLinkRange
      : manualLinkRange;
```

szydlovsky: It shouldn't happen but if the internals of some iOS function change again it will be less chaotic (:

ios/utils/OccurenceUtils.h
```diff
@@ -40,4 +40,6 @@
   withInput:(EnrichedTextInputView* _Nonnull)input
   inRange:(NSRange)range
   withCondition:(BOOL (NS_NOESCAPE ^_Nonnull)(id _Nullable value, NSRange range))condition;
++ (NSArray *_Nonnull)getRangesWithout:(NSArray<NSNumber *> *_Nonnull)types withInput:(EnrichedTextInputView* _Nonnull)input
```
szydlovsky: Could you please spread the arguments into different lines as in all the other functions?

ios/utils/OccurenceUtils.mm
```diff
@@ -152,4 +152,49 @@ + (BOOL)anyMultiple
   return occurences;
 }
 
++ (NSArray *_Nonnull)getRangesWithout:(NSArray<NSNumber *> *_Nonnull)types withInput:(EnrichedTextInputView* _Nonnull)input
```
szydlovsky: Arguments spreading would also be cool here if we keep it that way in the file. Some objc linter will eventually fix it though so not a big deal 😄

ios/utils/ParagraphAttributesUtils.mm
szydlovsky: Looks great 💣

ios/utils/OccurenceUtils.h
kacperzolkiewski: fixed

ios/utils/OccurenceUtils.mm
kacperzolkiewski: Fixed. But yes, we have to implement some code formatter for that.

README.md
```diff
@@ -152,9 +152,6 @@ Supported styles:
 - ordered list
 - unordered list
 
-> [!NOTE]
-> The iOS doesn't support codeblocks just yet, but it's planned in the near future!
-
```
szydlovsky, line 157: Let's revert the changes to the README, we'll update it just before the next release to not confuse anyone we have iOS codeblocks before it ;)

kacperzolkiewski: Done

exploIF:
Functionally it works well, however, I've noticed one issue. When you press backspace at beginning of the line it should remove style rather than go to previous line. I believe it should work the same as it works for block quote currently

https://github.com/user-attachments/assets/a509f72b-f00a-4815-8067-bc7eef0e491f

kacperzolkiewski:
@exploIF we decided with @szydlovsky that when you press backspace it will go back to the previous line, it is expected behavior. When you look how for example Slack works with `codeblock` it works like that.

szydlovsky:
Also, this way backspacing inside of a codeblock lets you just move stuff inside it instead of separating it into two codeblocks because style got removed for the line.

exploIF:
@kacperzolkiewski @szydlovsky So, this means that we also have to adjust that on Android, right now it's inconsistent

Issue #256, opened 2025-11-17
[Android] link styles applied in codeblock


kacperzolkiewski:
**Describe the bug**
When inserting a link manually inside a code block using the toolbar, the editor correctly restricts link styling - the link is not applied.
However, when a URL is automatically detected from plain text (e.g., typed like https://example.com), the link detection logic still applies link styles even inside a code block, where styling should be restricted.

**To Reproduce**
1. Go to Example app
2. Create codeblock
3. Type example.com inside codeblock

**Expected behavior**
Links inside a code block should never receive link styling, regardless of whether they are: inserted through the toolbar or automatically detected from typed text.

**Screenshots**

https://github.com/user-attachments/assets/b1617574-846a-4b9f-85f6-ad831c931ae0

PR #257, opened 2025-11-17
fix: handle style conflicts during automatic links detection

kacperzolkiewski:
# Summary
Fixes: #256 

Previously, links that were automatically detected from typed text could still receive link styles even in areas where link styling should be restricted (e.g. inside a code block).

This PR fixes that, making the behavior consistent with manual link application:
- Adds a new helper function `detectLinkConflicts` that checks whether the selected range contains blocking or conflicting styles.
- Updates `afterTextChangedLinks` to skip applying auto-detected links if any conflicts are detected.

## Test Plan

1. Go to Example app
2. Create codeblock
3. Type example.com inside codeblock

## Screenshots / Videos

https://github.com/user-attachments/assets/d1e66dec-527e-4bd5-adce-e1db2f2867ba

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

### DIFF
--- a/android/src/main/java/com/swmansion/enriched/styles/ParametrizedStyles.kt
+++ b/android/src/main/java/com/swmansion/enriched/styles/ParametrizedStyles.kt
@@ -104,29 +104,26 @@ class ParametrizedStyles(private val view: EnrichedTextInputView) {
     return Triple(result, start, end)
   }
 
-  private fun detectLinkConflicts(): Boolean {
-    val mergingConfig = EnrichedSpans.getMergingConfigForStyle(EnrichedSpans.LINK, view.htmlStyle)?: return false
+  private fun canLinkBeApplied(): Boolean {
+    val mergingConfig = EnrichedSpans.getMergingConfigForStyle(EnrichedSpans.LINK, view.htmlStyle)?: return true
     val conflictingStyles = mergingConfig.conflictingStyles
     val blockingStyles = mergingConfig.blockingStyles
 
     for (style in blockingStyles) {
-      if (view.spanState?.getStart(style) != null) return true
+      if (view.spanState?.getStart(style) != null) return false
     }
 
     for (style in conflictingStyles) {
-      if (view.spanState?.getStart(style) != null) return true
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/swmansion/enriched/styles/ParametrizedStyles.kt
```diff
@@ -104,15 +104,34 @@ class ParametrizedStyles(private val view: EnrichedTextInputView) {
     return Triple(result, start, end)
   }
 
+  private fun detectLinkConflicts(): Boolean {
```
exploIF: It would be cleaner if function name would suggest what's returned, maybe we can name it like `canLinkBeApplied`? Then obviously we have to flip conditions

android/src/main/java/com/swmansion/enriched/styles/ParametrizedStyles.kt
```diff
@@ -104,15 +104,34 @@ class ParametrizedStyles(private val view: EnrichedTextInputView) {
...
+
+    for (style in conflictingStyles) {
+      if (view.spanState?.getStart(style) != null) return true
+    }
+
+    return false
+  }
+
   private fun afterTextChangedLinks(result: Triple<String, Int, Int>) {
+    if (detectLinkConflicts()) {
```
exploIF: It would be more performant if this condition is checked after `if (isSettingLinkSpan) return`, no need to do additional computation in such scenario

android/src/main/java/com/swmansion/enriched/styles/ParametrizedStyles.kt
kacperzolkiewski: I rename the function into `canLinkBeApplied`

android/src/main/java/com/swmansion/enriched/styles/ParametrizedStyles.kt
kacperzolkiewski: Fixed

exploIF (approved):

Issue #235, opened 2025-11-11
HTML Entity Support


danielma:
We would appreciate finding a way to support HTML entities (e.g. `&lt;`, & `&rt;`). We have a backend that supports them, and for the most part in our case, we can replace these with their unicode equivalent and everything works, but it doesn't work with `<` and `>`. If we attempt that conversion, this library interprets those characters as HTML source, and can render improperly.

Ultimately, we would like users to be able to use this field to be allowed to use `<` and `>` to describe less than or greater than.

**Describe alternatives you've considered**

As mentioned above, we can handle all other HTML entities by replacing our source entities with their corresponding characters (listed [here, in the w3 reference](https://developer.mozilla.org/en-US/docs/Glossary/Character_reference))

szydlovsky:
Hey @danielma thanks for the issue! Indeed that's quite a thing we must have missed, expect a fix soon 😄

PR #258, opened 2025-11-17
feat: iOS Html entities handling

szydlovsky:
# Summary

Fixes https://github.com/software-mansion/react-native-enriched/issues/235; an issue that the iOS parser didn't handle special Html entities, `<`, `>` and `&` both ways.

## Test Plan

- Run example app on iOS
- Input a text consisting of these characters and see that they are properly switched to their codenames in emitted html
- For the other way around, paste the following HTML into either default value pr `Set input's value` field and see the codenames are properly translated to the characters:
```
<html>
<p>&lt;html&gt;<b>BOLD</b>&lt;/html&gt;</p>
</html>
```

Also worth double checking the same on Android, its system parser handles that well.

## Screenshots / Videos

image

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅ (system)     |

### DIFF
--- a/ios/utils/StringExtension.mm
+++ b/ios/utils/StringExtension.mm
@@ -39,13 +39,12 @@
 
     while(searchRange.location < text.length) {
       foundRange = [text rangeOfString:key options:0 range:searchRange];
-      if(foundRange.location != NSNotFound) {
-        results[@(foundRange.location)] = @[key, unescapeMap[key]];
-        searchRange.location = foundRange.location + foundRange.length;
-        searchRange.length = text.length - searchRange.location;
-      } else {
+      if(foundRange.location == NSNotFound) {
         break;
       }
+      results[@(foundRange.location)] = @[key, unescapeMap[key]];
+      searchRange.location = foundRange.location + foundRange.length;
+      searchRange.length = text.length - searchRange.location;
     }
   }
   


exploIF (approved):

ios/utils/StringExtension.mm
```diff
@@ -26,20 +24,32 @@ + (NSString *)stringByEscapingHtml:(NSString *)html {
...
+
+    while(searchRange.location < text.length) {
+      foundRange = [text rangeOfString:key options:0 range:searchRange];
+      if(foundRange.location != NSNotFound) {
+        results[@(foundRange.location)] = @[key, unescapeMap[key]];
+        searchRange.location = foundRange.location + foundRange.length;
+        searchRange.length = text.length - searchRange.location;
+      } else {
+        break;
+      }
```
kacperzolkiewski: Tiny cosmetic change but for me it looks cleaner: 
```suggestion
      if(foundRange.location == NSNotFound) {
      	break;
      } 
      results[@(foundRange.location)] = @[key, unescapeMap[key]];
      searchRange.location = foundRange.location + foundRange.length;
      searchRange.length = text.length - searchRange.location;
      }
```

kacperzolkiewski (approved):

PR #266, opened 2025-11-19
feat: add foreground color style

IvanIhnatsiuk:
# Summary

In this PR, I added foreground color style support for iOS. Currently, it allows changing color in any kind of style, and it uses the font tag to identify colored parts of attributed text from HTML. To track color in already colored styles, we need to check that the current location is in this style and the color is not the same as the configured one. 

Also, I'm not sure about the onSelectionColor event, do we need to run it only when we're in the color style, or would it be better to do it always?

## Test Plan

1. Open an app
2. Try to add any colored text
3. Check the HTML output

## Screenshots / Videos

https://github.com/user-attachments/assets/eac3ae6d-0fbf-4a0a-9ddf-421527f4f568

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

### DIFF
--- a/src/EnrichedTextInputNativeComponent.ts
+++ b/src/EnrichedTextInputNativeComponent.ts
@@ -32,6 +32,11 @@ export interface OnChangeStateEvent {
     isConflicting: boolean;
     isBlocking: boolean;
   };
+  colored: {
+    isActive: boolean;
+    isConflicting: boolean;
+    isBlocking: boolean;
+  };
   italic: {
     isActive: boolean;
     isConflicting: boolean;
@@ -138,6 +143,7 @@ export interface OnChangeStateDeprecatedEvent {
   isLink: boolean;
   isImage: boolean;
   isMention: boolean;
+  isColored: boolean;
 }
 
 export interface OnLinkDetected {
@@ -185,6 +191,10 @@ type Heading = {
   fontSize?: Float;
   bold?: boolean;
 };
+
+export interface OnChangeColorEvent {
+  color: string | null;
+}
 
 export interface HtmlStyleInternal {
   h1?: Heading;
@@ -256,6 +266,7 @@ export interface NativeProps extends ViewProps {
   onMention?: DirectEventHandler<OnMentionEvent>;
   onChangeSelection?: DirectEventHandler<OnChangeSelectionEvent>;
   onRequestHtmlResult?: DirectEventHandler<OnRequestHtmlResultEvent>;
+  onColorChangeInSelection?: DirectEventHandler<OnChangeColorEvent>;
 
   // Style related props - used for generating proper setters in component's manager
   // These should not be passed as regular props
@@ -330,6 +341,10 @@ interface NativeCommands {
     viewRef: React.ElementRef<ComponentType>,
     requestId: Int32
   ) => void;
+  toggleColor: (
+    viewRef: React.ElementRef<ComponentType>,
+    color: string
+  ) => void;
 }
 
 export const Commands: NativeCommands = codegenNativeCommands<NativeCommands>({
@@ -361,6 +376,7 @@ export const Commands: NativeCommands = codegenNativeCommands<NativeCommands>({
     'startMention',
     'addMention',
     'requestHTML',
+    'toggleColor',
   ],
 });
 
--- a/src/EnrichedTextInputNativeComponent.ts
+++ b/src/EnrichedTextInputNativeComponent.ts
@@ -193,7 +193,7 @@ type Heading = {
 };
 
 export interface OnChangeColorEvent {
-  color: string | null;
+  color: string;
 }
 
 export interface HtmlStyleInternal {
--- a/src/EnrichedTextInputNativeComponent.ts
+++ b/src/EnrichedTextInputNativeComponent.ts
@@ -341,10 +341,8 @@ interface NativeCommands {
     viewRef: React.ElementRef<ComponentType>,
     requestId: Int32
   ) => void;
-  toggleColor: (
-    viewRef: React.ElementRef<ComponentType>,
-    color: string
-  ) => void;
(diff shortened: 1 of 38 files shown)


szydlovsky:
Hey @IvanIhnatsiuk, it's super nice to get such verbose contribution from the community! While the feature is potentially interesting for us we cannot really proceed with it for now; we want the API and new features to be working on both mobile platforms from the get-go. Either way, great work!

android/src/main/new_arch/CMakeLists.txt
exploIF, line 1: Those changes seems completely unrelated to PR scope

.gitignore
exploIF, line 1: Redundant change

example/react-native.config.js
exploIF, line 1: Same here, let's keep this comment

example/src/components/ColorPreview.tsx
```diff
@@ -0,0 +1,32 @@
+import { StyleSheet, Text, View } from 'react-native';
+
+type Props = {
+  color: string;
+};
+
+const ColorPreview: React.FC<Props> = ({ color }) => {
```
exploIF: As a convention, we do not rely on global variables. so `FC` should be imported from `React`

example/src/components/ColorPreview.tsx
```diff
@@ -0,0 +1,32 @@
...
+const styles = StyleSheet.create({
+  preview: {
+    marginVertical: 8,
+    width: 40,
+    height: 40,
+    borderRadius: 20,
+  },
+});
+
+export default ColorPreview;
```
exploIF: As a convention, we use named exports

example/src/components/ToolbarColorButton.tsx
```diff
@@ -0,0 +1,47 @@
...
+    onPress(color);
+  };
+
+  const containerStyle = useMemo(
+    () => [
+      styles.container,
+      { backgroundColor: isActive ? color : 'rgba(0, 26, 114, 0.8)' },
+    ],
+    [isActive, color]
+  );
```
exploIF: `useMemo` seems redundant here, it's simple computation

package.json
exploIF, line 1: Out of scope change

src/EnrichedTextInputNativeComponent.ts
```diff
@@ -203,6 +209,10 @@ interface NativeCommands {
     text: string,
     payload: string
   ) => void;
+  toggleColor: (
```
exploIF: I think it should be rather called `setColor`, as we are explicitly passing color that should be used. We use `toggle` for two state styles (which is either enabled or disabled)

exploIF (changes requested):
I've added couple general comments. iOS part is still not reviewed.

Our main goal currently is to have consistent behavior and functionality between platforms. So I'm afraid we won't be able to ship that until Android part is implemented.
However, if this feature is something that community would like to see, we may consider implementing Android part ourselves

example/src/components/ToolbarColorButton.tsx
IvanIhnatsiuk: 1. I added this because of the warning about inline styles.
2. useMemo is useful here because it is an array that will create a new reference each time it is re-rendered and cause re-rendering in child components. Ideally, all values that are callbacks, arrays and objects should be wrapped with useCallback/useMemo if you are not using react-compiler (sometimes even if you are using it).

exploIF: Component is relatively simple, there is nothing wrong with re-renders if they are cheap (I believe this one is not complex at all). Any memoization method shouldn't be overused, it comes with a cost, and in this scenario can be avoided.

Also, you are not memoizing styles in `ColorPreview`, and there is no need to do that here

IvanIhnatsiuk: As I mentioned earlier, I added it to suppress the warning about the built-in style 🙂 

Screenshot 2025-12-08 at 22 15 12

In `ColorPreview`, I didn't get this warning, so I just ignored this case since it's a project example 🙂 . However, in real projects, it's better to wrap all reference props in a memo/useMemo/useCallback. 

Let's say you've updated the state in some component that passes styles as an array of styles in a View(generated, of course, during rendering). This array of styles must be pre-processed each time before sending styles to the native side (flattening, colour processing, marginHorizontal, etc.). It then calls C++(new arch) or serializes(old arch) this data and calls native methods that update the styles, but the styles remain the same.

IvanIhnatsiuk:
@exploIF @szydlovsky kind reminder 🙂

exploIF:
As we discussed, let's hold on with this PR for a while. It may potentially introduce regression on iOS, we are also thinking about using different - more flexible API. Once we solve current iOS implementation limitations we can come back to this one

nathan-hadley:
Just wanted to add a voice from the community here — foreground color styling is a must-have for our use case, and the lack of it is the main reason we haven't been able to adopt `react-native-enriched` yet. Great to see that @IvanIhnatsiuk has added Android support too — we would need both platforms. Let me know if there is anything we can do to help.

szydlovsky:
@nathan-hadley it'll probably need an entirely new implementation as @exploIF previously stated. Our current iOS implementation limits the freedom we have with font colors by quite a lot. Once the remake is done, we can easily make it happen 🫡. But we can't tell you for sure when it will be.

szydlovsky:
Coming back here with an announcement - the whole styling logic on iOS has been reworked which lets us build this feature (along with `background color` style too) cleanly and much more future-proofed.

Moreover, we've already begun the work on the feature.

Having said that, both iOS and Android base implementations have changed so much since this PR's arrival, that we're going to do it in a new PR with an up-to-date branch directly in the repo (forks are difficult to work with 😄).

Issue #170, opened 2025-09-24
Would really love to use this library but deeply need inline image support for iOS


salieflewis:
Is this next on the roadmap? So excited to see this project drop!

szt217:
+1! Awesome work here!!

salieflewis:
Any insight on this? @exploIF @szydlovsky

ashishmangukiya:
Any update on the feature? any tentative release month ?

kacperzolkiewski:
Hi @salieflewis, @szt217, @ashishmangukiya,

I’ve opened a PR that adds inline image support on iOS: #271.

PR #271, opened 2025-11-20
feat: add support for inline image on iOS

kacperzolkiewski:
# Summary
Fixes: #170 
Add support for inline image

## Test Plan
Run Example app and play with inline image

## Screenshots / Videos
https://github.com/user-attachments/assets/47fbc7e3-4441-4f17-9827-723d7c9d74d2

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

### DIFF
--- a/example/src/App.tsx
+++ b/example/src/App.tsx
@@ -203,12 +203,10 @@ export default function App() {
       selectionLimit: 1,
     });
 
-    let imageUri;
-    if (Platform.OS === 'android') {
-      imageUri = response.assets?.[0]?.originalPath;
-    } else {
-      imageUri = response.assets?.[0]?.uri;
-    }
+    const imageUri =
+      Platform.OS === 'android'
+        ? response.assets?.[0]?.originalPath
+        : response.assets?.[0]?.uri;
 
     if (!imageUri) return;
 


ios/EnrichedTextInputView.mm
```diff
@@ -917,6 +933,17 @@ - (void)addMention:(NSString *)indicator text:(NSString *)text attributes:(NSStr
   }
 }
 
+- (void)addImage:(NSString *)uri
```
exploIF, line 958: One general note, I think we should add support for setting image width and height directly when applying image style. That way we can specify dimensions for each image separately, which will be very useful as images may have different aspect ratios.

This should be done for both Android and iOS, but we can definitely handle that in separate PR

README.md
```diff
@@ -212,9 +212,6 @@ You can insert an image into the input using [setImage](docs/API_REFERENCE.md#se
 
 The image will be put into a single line in the input and will affect the line's height as well as input's height. Keep in mind, that image will replace currently selected text or insert into the cursor position if there is no text selection.
 
-> [!NOTE]
-> The iOS doesn't support inline images just yet, but it's planned in the near future!
-
```
szydlovsky, line 217: Once more, let's do the changes in docs just before the release ;)

ios/EnrichedTextInputView.mm
```diff
@@ -95,39 +95,42 @@ - (void)setDefaults {
...
     @([H2Style getStyleType]): @[@([H1Style getStyleType]), @([H3Style getStyleType]), @([UnorderedListStyle getStyleType]), @([OrderedListStyle getStyleType]), @([BlockQuoteStyle getStyleType])],
     @([H3Style getStyleType]): @[@([H1Style getStyleType]), @([H2Style getStyleType]), @([UnorderedListStyle getStyleType]), @([OrderedListStyle getStyleType]), @([BlockQuoteStyle getStyleType])],
     @([UnorderedListStyle getStyleType]): @[@([H1Style getStyleType]), @([H2Style getStyleType]), @([H3Style getStyleType]), @([OrderedListStyle getStyleType]), @([BlockQuoteStyle getStyleType])],
     @([OrderedListStyle getStyleType]): @[@([H1Style getStyleType]), @([H2Style getStyleType]), @([H3Style getStyleType]), @([UnorderedListStyle getStyleType]), @([BlockQuoteStyle getStyleType])],
-    @([BlockQuoteStyle getStyleType]): @[@([H1Style getStyleType]), @([H2Style getStyleType]), @([H3Style getStyleType]), @([UnorderedListStyle getStyleType]), @([OrderedListStyle getStyleType])]
+    @([BlockQuoteStyle getStyleType]): @[@([H1Style getStyleType]), @([H2Style getStyleType]), @([H3Style getStyleType]), @([UnorderedListStyle getStyleType]), @([OrderedListStyle getStyleType])],
```
szydlovsky: More like a TODO; but once we merge iOS codeblocks, we also need to setup the blocking/conflicts in here and test it out

ios/inputParser/InputParser.mm
```diff
@@ -603,6 +629,14 @@ - (NSArray *)getTextAndStylesFromHtml:(NSString *)fixedHtml {
       [styleArr addObject:@([BoldStyle getStyleType])];
     } else if([tagName isEqualToString:@"i"]) {
       [styleArr addObject:@([ItalicStyle getStyleType])];
+    } else if([tagName isEqualToString:@"img"]) {
+      [styleArr addObject:@([ImageStyle getStyleType])];
+      // cut only the uri from the src="..." string
+      NSString *uri = [params substringWithRange:NSMakeRange(5, params.length - 6)];
```
szydlovsky: I think we could do the same trick as with the links; use a regex for finding `src` and if there is none (or empty `src`), then we don't recognize the tag. This way we can dodge any crashes from not getting the params for the tag.

ios/styles/ImageStyle.mm
```diff
@@ -0,0 +1,149 @@
+#import "StyleHeaders.h"
+#import "EnrichedTextInputView.h"
+#import "OccurenceUtils.h"
+#import "TextInsertionUtils.h"
+
+// custom NSAttributedStringKey to differentiate the image
+static NSString *const ImageAttributeName = @"ImageAttributeName";
+
```
szydlovsky, line 8:
```suggestion
```

ios/utils/LayoutManagerExtension.mm
```diff
@@ -132,6 +135,48 @@ - (void)my_drawBackgroundForGlyphRange:(NSRange)glyphRange atPoint:(CGPoint)orig
   }
 }
 
+-(void)drawImages:(EnrichedTextInputView *)typedInput origin:(CGPoint)origin inputRange:(NSRange)inputRange
+{
+  ImageStyle *imgStyle = typedInput->stylesDict[@([ImageStyle getStyleType])];
+  
+  if (imgStyle == nullptr) { return; }
+    NSArray *allImages = [imgStyle findAllOccurences:inputRange];
```
szydlovsky: NIT: everything below that `if` is tabbed once when it doesn't need to be

ios/EnrichedTextInputView.mm
```diff
@@ -95,39 +95,42 @@ - (void)setDefaults {
...
     @([BoldStyle getStyleType]) : @[],
     @([ItalicStyle getStyleType]) : @[],
     @([UnderlineStyle getStyleType]) : @[],
     @([StrikethroughStyle getStyleType]) : @[],
-    @([InlineCodeStyle getStyleType]) : @[],
-    @([LinkStyle getStyleType]): @[],
-    @([MentionStyle getStyleType]): @[],
+    @([InlineCodeStyle getStyleType]) : @[@([ImageStyle getStyleType])],
+    @([LinkStyle getStyleType]): @[@([ImageStyle getStyleType])],
+    @([MentionStyle getStyleType]): @[@([ImageStyle getStyleType])],
```
szydlovsky: I'd also make all the inline styles blocked by the image, because currently you can just apply e.g. bold onto the image attachment and it represents kinda nothing. By inline styles I mean; bold, italic, underline, strikethrough.

ios/EnrichedTextInputView.mm
```diff
@@ -95,39 +95,42 @@ - (void)setDefaults {
...
+    @([ImageStyle getStyleType]): [[ImageStyle alloc] initWithInput:self]
   };
   
   _conflictingStyles = @{
     @([BoldStyle getStyleType]) : @[],
     @([ItalicStyle getStyleType]) : @[],
     @([UnderlineStyle getStyleType]) : @[],
     @([StrikethroughStyle getStyleType]) : @[],
-    @([InlineCodeStyle getStyleType]) : @[@([LinkStyle getStyleType]), @([MentionStyle getStyleType])],
+    @([InlineCodeStyle getStyleType]) : @[@([LinkStyle getStyleType]), @([MentionStyle getStyleType]), @([ImageStyle getStyleType])],
```
szydlovsky: We have to decide if Image blocks the InlineCode or is conflicting with it (and gets removed when applying inline code). Currently the blocking takes precedence and this line does nothing.

szydlovsky (commented):
Generally works fine. Left a few technical comments. The blocking/conflicting needs to be discussed because I feel like it's not too clear right now

ios/EnrichedTextInputView.mm
kacperzolkiewski: I deleted Image Style from here

ios/EnrichedTextInputView.mm
kacperzolkiewski: Fixed

ios/utils/LayoutManagerExtension.mm
kacperzolkiewski: This code was deleted because drawing is handled automatically when we use `NSTextAttachment`

ios/styles/ImageStyle.mm
kacperzolkiewski: fixed

README.md
kacperzolkiewski: Fixed

ios/inputParser/InputParser.mm
kacperzolkiewski: I am using the regex right now

example/src/App.tsx
```diff
@@ -202,7 +203,13 @@ export default function App() {
       selectionLimit: 1,
     });
 
-    const imageUri = response.assets?.[0]?.originalPath;
+    let imageUri;
+    if (Platform.OS === 'android') {
+      imageUri = response.assets?.[0]?.originalPath;
+    } else {
+      imageUri = response.assets?.[0]?.uri;
+    }
```
exploIF: Minor one, I'm hater of `let`, can we use ternary instead?

kacperzolkiewski: fixed

ios/EnrichedTextInputView.mm

szydlovsky (approved):

Issue #238, opened 2025-11-11
BUG: initial layout delay


IvanIhnatsiuk:
**Describe the bug**

The initial layout calculation is delayed, and as a result, it's possible to see a delay in rendering on both iOS and Android. 

**To Reproduce**

Apply this code first:

```.tsx
import {
  View,
  StyleSheet,
  Text,
  type NativeSyntheticEvent,
  ScrollView,
} from 'react-native';
import {
  EnrichedTextInput,
  type OnChangeTextEvent,
  type EnrichedTextInputInstance,
  type OnLinkDetected,
  type OnChangeMentionEvent,
  type OnChangeHtmlEvent,
  type OnChangeStateEvent,
  type OnChangeSelectionEvent,
  type HtmlStyle,
} from 'react-native-enriched';
import { useRef, useState } from 'react';
import { Button } from './components/Button';
import { Toolbar } from './components/Toolbar';
import { LinkModal } from './components/LinkModal';
import { ValueModal } from './components/ValueModal';
import { launchImageLibrary } from 'react-native-image-picker';
import { type MentionItem, MentionPopup } from './components/MentionPopup';
import { useUserMention } from './useUserMention';
import { useChannelMention } from './useChannelMention';
import { HtmlSection } from './components/HtmlSection';

type StylesState = OnChangeStateEvent;

type CurrentLinkState = OnLinkDetected;

interface Selection {
  start: number;
  end: number;
  text: string;
}

const DEFAULT_STYLE: StylesState = {
  isBold: false,
  isItalic: false,
  isUnderline: false,
  isStrikeThrough: false,
  isInlineCode: false,
  isH1: false,
  isH2: false,
  isH3: false,
  isH4: false,
  isH5: false,
  isH6: false,
  isBlockQuote: false,
  isCodeBlock: false,
  isOrderedList: false,
  isUnorderedList: false,
  isLink: false,
  isImage: false,
  isMention: false,
};

const DEFAULT_LINK_STATE = {
  text: '',
  url: '',
  start: 0,
  end: 0,
};

const DEBUG_SCROLLABLE = false;

// Enabling this prop fixes input flickering while auto growing.
// However, it's still experimental and not tested well.
const ANDROID_EXPERIMENTAL_SYNCHRONOUS_EVENTS = true;

export default function App() {
  const [isChannelPopupOpen, setIsChannelPopupOpen] = useState(false);
  const [isUserPopupOpen, setIsUserPopupOpen] = useState(false);
[...]

PR #277, opened 2025-11-21
feat: proper iOS first renders measurements

szydlovsky:
# Summary

Fixes https://github.com/software-mansion/react-native-enriched/issues/238 on iOS.

First render on iOS was looking pretty choppy because there was no way to measure the component on the very first render, because it's still not defined at that time.
This PR adds a mock text input that gets created as soon as the Shadow Node is created, gets the initial component's props and is used for measurements before the real input is defined. After that the mock gets unmounted (when the real input is finally defined).

## Before vs After

Please play the videos at x0.25 speed to notice the choppiness:

### Before:

https://github.com/user-attachments/assets/69aa524a-5ec1-4fbc-8220-610d2280eaf4

### After:

https://github.com/user-attachments/assets/8f07e5a1-67e5-4c6d-90df-d0b4f44e0f62

## Test Plan

Add a state that toggles input's visibility, with the default visibility at `false`. Then toggle it and see properly measured component appear with no intermediate states.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅    |
| Android |    ✅(already done)    |


szydlovsky:
@exploIF oh definitely. Will add it

PR #285, opened 2025-11-25
feat: add dimensions to images

kacperzolkiewski:
# Summary

Add possibility to set `width` and `height` to particular image.

## Test Plan
Run example app and add images with different dimensions.

## Screenshots / Videos
iOS:

https://github.com/user-attachments/assets/a18b77a6-e171-4f5e-b55d-ed6cfa7c4213

Android:

https://github.com/user-attachments/assets/fe0dc041-d544-46c6-9d5e-c883e65223ef

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

### DIFF
--- a/ios/styles/ImageStyle.mm
+++ b/ios/styles/ImageStyle.mm
@@ -103,7 +103,7 @@ static NSString *const ImageAttributeName = @"ImageAttributeName";
   return imageData;
 }
 
-- (void)addImageAtRange:(NSRange)range imageData:(ImageData *)imageData
+- (void)addImageAtRange:(NSRange)range imageData:(ImageData *)imageData withSelection:(BOOL)withSelection
 {
   UIImage *img = [self prepareImageFromUri:imageData.uri];
   
@@ -112,7 +112,7 @@ static NSString *const ImageAttributeName = @"ImageAttributeName";
     return;
   }
   
-  NSMutableDictionary *attributes = [@{
+  NSDictionary *attributes = [@{
     NSAttachmentAttributeName: [self prepareImageAttachement:img width:imageData.width height:imageData.height],
     ImageAttributeName: imageData,
   } mutableCopy];
@@ -122,13 +122,13 @@ static NSString *const ImageAttributeName = @"ImageAttributeName";
   NSString *imagePlaceholder = @"\uFFFC";
   
   if (range.length == 0) {
-    [TextInsertionUtils insertText:imagePlaceholder at:range.location additionalAttributes:attributes input:_input withSelection:YES];
+    [TextInsertionUtils insertText:imagePlaceholder at:range.location additionalAttributes:attributes input:_input withSelection:withSelection];
   } else {
     [TextInsertionUtils replaceText:imagePlaceholder
                                  at:range
                additionalAttributes:attributes
                               input:_input
-                      withSelection:YES];
+                      withSelection:withSelection];
   }
 }
 
@@ -138,7 +138,7 @@ static NSString *const ImageAttributeName = @"ImageAttributeName";
   data.width = width;
   data.height = height;
   
-  [self addImageAtRange:_input->textView.selectedRange imageData:data];
+  [self addImageAtRange:_input->textView.selectedRange imageData:data withSelection:YES];
 }
 
 -(NSTextAttachment *)prepareImageAttachement:(UIImage *)image width:(CGFloat)width height:(CGFloat)height
--- a/ios/inputParser/InputParser.mm
+++ b/ios/inputParser/InputParser.mm
@@ -459,7 +459,7 @@
         [((MentionStyle *)baseStyle) addMentionAtRange:styleRange params:params];
       } else if([styleType isEqualToNumber: @([ImageStyle getStyleType])]) {
         ImageData *imgData = (ImageData *)stylePair.styleValue;
-        [((ImageStyle *)baseStyle) addImageAtRange:styleRange imageData:imgData];
+        [((ImageStyle *)baseStyle) addImageAtRange:styleRange imageData:imgData withSelection:NO];
       } else {
         [baseStyle addAttributes:styleRange];
       }
--- a/ios/utils/StyleHeaders.h
+++ b/ios/utils/StyleHeaders.h
@@ -83,6 +83,6 @@
 
 @interface ImageStyle : NSObject<BaseStyleProtocol>
 - (void)addImage:(NSString *)uri width:(CGFloat)width height:(CGFloat)height;
-- (void)addImageAtRange:(NSRange)range imageData:(ImageData *)imageData;
+- (void)addImageAtRange:(NSRange)range imageData:(ImageData *)imageData withSelection:(BOOL)withSelection;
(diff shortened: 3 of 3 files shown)


android/src/main/java/com/swmansion/enriched/spans/EnrichedImageSpan.kt
```diff
@@ -35,7 +38,17 @@ class EnrichedImageSpan : ImageSpan, EnrichedInlineSpan {
 
   override fun getDrawable(): Drawable {
     val drawable = super.getDrawable()
-    drawable.setBounds(0, 0, htmlStyle!!.imgWidth, htmlStyle!!.imgHeight)
+    val scale = Resources.getSystem().displayMetrics.density
+    
+    drawable.setBounds(0, 0, (width * scale).toInt() , (height * scale).toInt())
```
kacperzolkiewski, line 43: Without scale it was not the same size like on iOS

android/src/main/java/com/swmansion/enriched/styles/ParametrizedStyles.kt
```diff
@@ -170,26 +170,28 @@ class ParametrizedStyles(private val view: EnrichedTextInputView) {
...
+        spannable.removeSpan(s)
+      }
+
+      spannable.replace(start, originalEnd, "\uFFFC")
     }
 
+    val imageEnd = start + 1;
     val uri = Uri.fromFile(File(src))
-    val span = EnrichedImageSpan(view.context, uri, view.htmlStyle)
-    val (safeStart, safeEnd) = spannable.getSafeSpanBoundaries(start, end)
```
exploIF, line 191: I would still use `scannable.getSafeSpanBoundries`, it just ensures we are not running out of spannable text range. I think it's good practice to use it everywhere

android/src/main/java/com/swmansion/enriched/utils/EnrichedParser.java
```diff
@@ -679,20 +694,51 @@ private static void end(Editable text, Class kind, Object repl) {
...
+      Bitmap bitmap = BitmapFactory.decodeFile(cleanPath);
+      if (bitmap != null) {
+        drawable = new BitmapDrawable(Resources.getSystem(), bitmap);
+        // set bounds so it knows how big it is naturally,
+        // though EnrichedImageSpan will override this with the HTML width/height later.
+        drawable.setBounds(0, 0, bitmap.getWidth(), bitmap.getHeight());
+      }
+    } catch (Exception e) {
+      // Failed to load file
+      android.util.Log.e("EnrichedParser", "Failed to load image from path: " + cleanPath, e);
```
exploIF: Can we just import `Log` so we use this class directly?

example/src/components/ImageModal.tsx
```diff
@@ -0,0 +1,107 @@
...
+  onClose,
+  onSubmit,
+}) => {
+  const [width, setWidth] = useState('');
+  const [height, setHeight] = useState('');
+
+  const handleSave = () => {
+    const parsedWidth = parseFloat(width);
+    const parsedHeight = parseFloat(height);
+    const finalWidth = isNaN(parsedWidth) ? DEFAULT_IMAGE_WIDTH : parsedWidth;
```
exploIF, line 26: Maybe we can slightly change example app logic:
- let's say in example width and height are optional fields
- if they are provided, we use them
- if they are not, we use default image height and calculate width based on filesystem image size to nicely handle aspect ratio

I think this will be good example, so lib users can see how it can be handled if you want to automatically handle image sizing.

Wdyt? Btw, I'm fine with handling than in separate PR

kacperzolkiewski: Cool idea. But I'd prefer to do it in a separate PR.

android/src/main/java/com/swmansion/enriched/styles/ParametrizedStyles.kt
kacperzolkiewski: Fixed

android/src/main/java/com/swmansion/enriched/utils/EnrichedParser.java
kacperzolkiewski: Fixed

ios/EnrichedTextInputView.mm
```diff
@@ -955,13 +948,13 @@ - (void)addMention:(NSString *)indicator text:(NSString *)text attributes:(NSStr
   }
 }
 
-- (void)addImage:(NSString *)uri
+- (void)addImage:(NSString *)uri width:(float)width height:(float)height
```
szydlovsky, line 951:
```suggestion
- (void)addImage:(NSString *)uri width:(CGFloat)width height:(CGFloat)height
```

ios/styles/ImageStyle.mm
```diff
@@ -103,48 +103,61 @@ - (ImageData *)getImageDataAt:(NSUInteger)location
...
+- (void)addImageAtRange:(NSRange)range imageData:(ImageData *)imageData
+{
+  UIImage *img = [self prepareImageFromUri:imageData.uri];
+  
+  if (img == nil) {
+    // Could not create image from URI, stop here to avoid the "OBJ" icon
+    return;
+  }
   
   NSMutableDictionary *attributes = [@{
```
szydlovsky: Doesn't need to be Mutable, right?

ios/styles/ImageStyle.mm
```diff
@@ -103,48 +103,61 @@ - (ImageData *)getImageDataAt:(NSUInteger)location
...
+  if (range.length == 0) {
+    [TextInsertionUtils insertText:imagePlaceholder at:range.location additionalAttributes:attributes input:_input withSelection:YES];
   } else {
     [TextInsertionUtils replaceText:imagePlaceholder
-                                 at:_input->textView.selectedRange
-               additionalAttributes:nullptr
+                                 at:range
+               additionalAttributes:attributes
                               input:_input
                       withSelection:YES];
```
szydlovsky: Not sure if we want to always set the selection (`withSelection` flag). I'd rather have that method parametrised for that selection. In the `InputParser` we don't want selection to be changed because it can emit excess `onSelectionChanged` events. On the other hand, the `addImage uri: width: height:` method needs the selection when using it.

ios/EnrichedTextInputView.mm
kacperzolkiewski: image

When I use `CGFloat` here it causes warning.

ios/styles/ImageStyle.mm
kacperzolkiewski: fixed

ios/styles/ImageStyle.mm
kacperzolkiewski: Yeap

szydlovsky (approved):
:shipit:

ios/EnrichedTextInputView.mm
szydlovsky: Okay, scratch that, keep the float then.

PR #286, opened 2025-11-25
feat: mentions with space - android

exploIF:
# Summary

This PR brings possibility to use mentions with one space inside it on Android platform.

## Test Plan

For example:

- run example app
- enter `@John Doe`
- notice that entering whitespace does not close mention selector

## Screenshots / Videos

https://github.com/user-attachments/assets/49e0baa0-acc4-498f-a0bf-e2c3c0a67625

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

### DIFF
--- a/android/src/main/java/com/swmansion/enriched/styles/ParametrizedStyles.kt
+++ b/android/src/main/java/com/swmansion/enriched/styles/ParametrizedStyles.kt
@@ -85,7 +85,7 @@ class ParametrizedStyles(private val view: EnrichedTextInputView) {
     }
   }
 
-  private fun getWordAtIndex(s: Editable, index: Int): TextRange? {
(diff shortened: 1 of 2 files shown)


szydlovsky:
Works mostly fine but I found one issue: sometimes we don't get on mention event where it definitely should be. See here on the recording; we start typing a mention, then write a second word, then come back with selection to the first one. It should definitely start editing again but it doesn't:

https://github.com/user-attachments/assets/39d9fc00-5c97-4976-b4e6-bbc5eb3391e8

exploIF:
Thanks for catching it, this is something that we should definitely consider improving in the future. It mostly comes from the fact that we are relying on text change, rather than on selection change. As it's not regression, and it's working exactly the same on current main branch, I will add it to our roadmap and we can think on it later

PR #289, opened 2025-11-26
feat: make iOS mentions work with a single space

szydlovsky:
# Summary

This PR makes iOS mentions tolerate a single space. 
Code responsible for detecting a currently edited mention would just take the current word into the account.
Now it works by trying to search for one more word before the current one if needed. This way all of the following scenarios will be detected as editable mention:
- current word with a mention indicator
- a space character, but previous word has a mention indicator
- current word with no mention indicator but previous word has one

Effectively these scenarios are enough to achieve space tolerating mentions. 

## Test Plan

Open iOS example app and play with mentions and spaces.

## Screenshots / Videos

https://github.com/user-attachments/assets/80b53b4d-509a-4d43-8a33-53e911e27132

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

### DIFF
--- a/ios/EnrichedTextInputView.mm
+++ b/ios/EnrichedTextInputView.mm
@@ -1035,7 +1035,12 @@ Class<RCTComponentViewProtocol> EnrichedTextInputViewCls(void) {
   MentionStyle *mentionStyleClass = (MentionStyle *)stylesDict[@([MentionStyle getStyleType])];
   if(mentionStyleClass != nullptr) {
     [mentionStyleClass manageMentionTypingAttributes];
-    [mentionStyleClass manageMentionEditing];
+    
+    // mention editing runs if only a selection was done (no text change)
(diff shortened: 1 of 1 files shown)


ios/styles/MentionStyle.mm
```diff
@@ -299,45 +299,127 @@ - (void)manageMentionEditing {
...
+        // no current word, previous has no mention indicators
+        [self removeActiveMentionRange];
+        return;
+      }
+    } else {
+      // no current word, no previous word
+      [self removeActiveMentionRange];
+      return;
+    }
   }
```
kacperzolkiewski: Maybe we can separate this logic into different function, because currently our `manageMentionEditing` function is doing multiple things and it is getting bigger? e.g. 
`- (NSDictionary * _Nullable)detectMentionInContext` where we will return `{ finalText, finalRange } `or `nil` based on results?

PR #297, opened 2025-11-27
fix: use default codgen output dirs

IvanIhnatsiuk:
# Summary

Currently, it creates codegen files in platform-specific folders(Android and iOS). When you fork this library, codegen can not generate/re-generate code in these folders. So, it just skips this phase.

### Why does it work for now if codegen doesn't generate code? 

If you take a look at the published files, you may notice that codegen is already included https://www.npmjs.com/package/react-native-enriched?activeTab=code . Go to the generated folders in Android or iOS.

So, in the library, you use imports from these files; however, you don't modify them at all and it's just a cosmetic

### What problems does it solve?

1. If you work on a fork version of this library, you need to store it with already "generated" code. Since codegen can not run properly. It causes additional steps for each change in the fork, and it's really problematic to handle.
2. If you would like to apply a patch package with new props/commands or any codegen-specific logic it won't work, since codegen doesn't regenerate the new code

## Test Plan

build iOS and Android

## Screenshots / Videos

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

### DIFF
--- a/android/src/main/new_arch/CMakeLists.txt
+++ b/android/src/main/new_arch/CMakeLists.txt
@@ -5,7 +5,7 @@ set(LIB_LITERAL RNEnrichedTextInputViewSpec)
 set(LIB_TARGET_NAME react_codegen_${LIB_LITERAL})
 
 set(LIB_ANDROID_DIR ${CMAKE_CURRENT_SOURCE_DIR}/../../..)
-set(LIB_ANDROID_GENERATED_JNI_DIR ${LIB_ANDROID_DIR}/generated/jni)
+set(LIB_ANDROID_GENERATED_JNI_DIR ${LIB_ANDROID_DIR}/build/generated/source/codegen/jni)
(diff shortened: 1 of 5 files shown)


exploIF:
Hi, I would definitely not call this PR a "fix", current setup works well and comes with many benefits, that you can check [here](https://reactnative.dev/docs/the-new-architecture/codegen-cli#including-generated-code-into-libraries).

I see your point, but not especially agree with arguments. If someone forks or patches the library, they might as well change this setting.
I think bigger limitation is when it comes to installing library from branch or commit, then, Codegen has to be run manually in the client app, which can be problematic.

It would be the best to continue this topic as a discussion, and listen to the community's opinion. If there are many people interested in such change we may consider changing this setting (and reopen this PR), but as for now, we are happy with current setup.

PR #301, opened 2025-12-01
feat: add support for remote images on Android

kacperzolkiewski:
# Summary

This PR:

-  adds support for remote images on Android 
-  use local asset [broken_image](https://fonts.google.com/icons?selected=Material+Symbols+Outlined:broken_image:FILL@0;wght@400;GRAD@0;opsz@24&icon.query=broken+ima&icon.size=24&icon.color=%23e3e3e3) as a fallback on Android
-  extends example to handle aspect ratio when adding images

## Test Plan
Run Example app and try to add remote images:

1. through toolbar menu
2. through  `Set input's value` button (provide html with remote img tag)

## Screenshots / Videos

https://github.com/user-attachments/assets/9a623128-bac1-4763-b2e0-9179cbc1c473

https://github.com/user-attachments/assets/6c4f1679-a67f-4beb-b8c8-824fe05a5797

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |


android/src/main/java/com/swmansion/enriched/EnrichedTextInputView.kt
```diff
@@ -262,6 +264,23 @@ class EnrichedTextInputView : AppCompatEditText {
...
 
+  /**
+   * Finds all async images in the current text and sets up listeners
+   * to redraw the text layout when they finish downloading.
+   */
+  private fun observeAsyncImages() {
+    val liveText = text
+    if (liveText == null) {
+      return
+    }
```
exploIF:
```suggestion
    val liveText = text ?: return
```

android/src/main/java/com/swmansion/enriched/spans/EnrichedImageSpan.kt
```diff
@@ -39,16 +46,121 @@ class EnrichedImageSpan : ImageSpan, EnrichedInlineSpan {
...
+    if (d !is AsyncDrawable) {
+      return
+    }
+
+    registerDrawableLoadCallback(d, text)
+
+    // If it's already loaded (race condition), run logic immediately
+    if (d.isLoaded) {
+      d.onLoaded?.invoke()
+    }
```
exploIF, line 111: Question, shouldn't we check if it's loaded before calling `registerDrawableLoadCallback` and not call it at all if it's loaded already?

android/src/main/java/com/swmansion/enriched/spans/EnrichedImageSpan.kt
```diff
@@ -39,16 +46,121 @@ class EnrichedImageSpan : ImageSpan, EnrichedInlineSpan {
...
+        val bitmap = BitmapFactory.decodeFile(cleanPath)
+        if (bitmap != null) {
+          drawable = bitmap.toDrawable(Resources.getSystem())
+          // set bounds so it knows how big it is naturally,
+          // though EnrichedImageSpan will override this with the HTML width/height later.
+          drawable.setBounds(0, 0, bitmap.getWidth(), bitmap.getHeight())
+        }
+      } catch (e: Exception) {
+        // Failed to load file
+        Log.e("EnrichedParser", "Failed to load image from path: $cleanPath", e)
```
exploIF: Let's use a file name as a tag, easier to find in the code later

```suggestion
        Log.e("EnrichedImageSpan", "Failed to load image from path: $cleanPath", e)
```

android/src/main/java/com/swmansion/enriched/spans/EnrichedImageSpan.kt
```diff
@@ -39,16 +46,121 @@ class EnrichedImageSpan : ImageSpan, EnrichedInlineSpan {
...
     return height
   }
+
+  companion object {
+    @SuppressLint("UseCompatLoadingForDrawables")
+    fun createEnrichedImageSpan(src: String, width: Int, height: Int): EnrichedImageSpan {
+      var imgDrawable = prepareDrawableForImage(src)
+
+      if (imgDrawable == null) {
+        val systemIcon = Resources.getSystem().getDrawable(android.R.drawable.ic_menu_report_image)
```
exploIF: While it may work fine for most of the apps and systems version, I'm afraid there is no guarantee this resource will be available for every case.

I think, instead, we should use local assets. We can get if from https://m3.material.io/styles/icons/overview

android/src/main/java/com/swmansion/enriched/utils/AsyncDrawable.kt
```diff
@@ -0,0 +1,93 @@
...
+          isLoaded = true
+          onLoaded?.invoke()
+        }
+      } catch (e: Exception) {
+        Log.e("UrlDrawable", "Failed to load: $url", e)
+
+        isLoaded = true
+        loadPlaceholderImage()
+        onLoaded?.invoke()
+      }
```
exploIF, line 60: Looks like below can be handled in `finally` clause
```kt
isLoaded = true
onLoaded?.invoke()
```

android/src/main/java/com/swmansion/enriched/utils/AsyncDrawable.kt
```diff
@@ -0,0 +1,93 @@
...
+            d.bounds = bounds
+            internalDrawable = d
+          } else {
+            loadPlaceholderImage()
+          }
+          isLoaded = true
+          onLoaded?.invoke()
+        }
+      } catch (e: Exception) {
+        Log.e("UrlDrawable", "Failed to load: $url", e)
```
exploIF:
```suggestion
        Log.e("AsyncDrawable", "Failed to load: $url", e)
```

android/src/main/java/com/swmansion/enriched/utils/AsyncDrawable.kt
```diff
@@ -0,0 +1,93 @@
...
+        isLoaded = true
+        loadPlaceholderImage()
+        onLoaded?.invoke()
+      }
+    }
+  }
+
+  @SuppressLint("UseCompatLoadingForDrawables")
+  private fun loadPlaceholderImage() {
+    val systemIcon = Resources.getSystem().getDrawable(android.R.drawable.ic_menu_report_image)
```
exploIF: Same here, let's use local resource

android/src/main/java/com/swmansion/enriched/spans/utils/ForceRedrawSpan.kt
exploIF, line 1: Let's put this file in `swmansion.enriched.spans.utils`

example/src/App.tsx
```diff
@@ -207,19 +210,40 @@ export default function App() {
...
     const response = await launchImageLibrary({
       mediaType: 'photo',
       selectionLimit: 1,
     });
 
-    const imageUri =
-      Platform.OS === 'android'
-        ? response.assets?.[0]?.originalPath
-        : response.assets?.[0]?.uri;
+    if (response?.assets?.[0] !== undefined) {
```
exploIF: Let's flip this condition, and do early return if first asset from response is undefined

example/src/App.tsx
```diff
@@ -207,19 +210,40 @@ export default function App() {
...
-      ref.current?.setImage(imageUri, width, height);
+      const { finalWidth, finalHeight } = prepareImageDimensions(
+        asset,
+        width,
+        height
+      );
+
+      if (imageUri) {
+        ref.current?.setImage(imageUri, finalWidth, finalHeight);
+      }
```
exploIF: Minor, if there is no `imageUri` there is no point in calling `prepareImageDimensions `

example/src/App.tsx
```diff
@@ -356,6 +380,50 @@ export default function App() {
   );
 }
 
+const prepareImageDimensions = (
```
exploIF: Let's move this to separate file, `App.tsx` is already pretty big

example/src/components/ImageModal.tsx
```diff
@@ -6,29 +6,33 @@ import { Icon } from './Icon';
...
-      : parsedHeight;
+    const finalWidth = isNaN(parsedWidth) ? '' : parsedWidth;
+    const finalHeight = isNaN(parsedHeight) ? '' : parsedHeight;
 
-    onSubmit(finalWidth, finalHeight);
+    onSubmit(
+      finalWidth === '' ? undefined : finalWidth,
+      finalHeight === '' ? undefined : finalHeight,
+      url.trim() === '' ? undefined : url
+    );
```
exploIF: I don't exactly get point of first assigning empty string if it's NaN, and then using undefined if it's empty string. I think, it can be done in one step

example/src/components/ImageModal.tsx
```diff
@@ -6,29 +6,33 @@ import { Icon } from './Icon';
...
+    width: number | undefined,
+    height: number | undefined,
+    url?: string
+  ) => void;
 }
 
-const DEFAULT_IMAGE_WIDTH = 80;
-const DEFAULT_IMAGE_HEIGHT = 80;
-
 export const ImageModal: FC<ImageModalProps> = ({
```
exploIF, line 16: Let's clear state on close, so width, height and src uses default value every time we reopen the modal

android/src/main/java/com/swmansion/enriched/spans/EnrichedImageSpan.kt
kacperzolkiewski: If we will not call it if the drawable is loaded we would have loaded image but we would not see it because the span would not be redrawn.

android/src/main/java/com/swmansion/enriched/spans/EnrichedImageSpan.kt
kacperzolkiewski: Good idea 🚀

PR #302, opened 2025-12-04
fix: merging adjacent continuous paragraph styles on Android

kacperzolkiewski:
# Summary

Merge adjacent continuous spans to maintain single set of html tags and proper rounded corners for code blocks.

## Test Plan
Run example app and:
1. Create two codeblocks/blockquotes separated with line without codeblock style -> delete this line (blocks should merge)
2. Create one codeblock/blockquotes go to next line and create the other one (blocks should merge)

## Screenshots / Videos

https://github.com/user-attachments/assets/172cec6c-88ff-49d7-b897-4e6433983ad3

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

### DIFF
--- a/android/src/main/java/com/swmansion/enriched/styles/ParagraphStyles.kt
+++ b/android/src/main/java/com/swmansion/enriched/styles/ParagraphStyles.kt
@@ -3,6 +3,7 @@ package com.swmansion.enriched.styles
 import android.text.Editable
 import android.text.Spannable
 import android.text.SpannableStringBuilder
+import android.util.Log
 import com.swmansion.enriched.EnrichedTextInputView
 import com.swmansion.enriched.spans.EnrichedBlockQuoteSpan
 import com.swmansion.enriched.spans.EnrichedCodeBlockSpan
@@ -17,6 +18,12 @@ class ParagraphStyles(private val view: EnrichedTextInputView) {
     val (previousParagraphStart, previousParagraphEnd) = spannable.getParagraphBounds(paragraphStart - 1)
     val spans = spannable.getSpans(previousParagraphStart, previousParagraphEnd, type)
 
+    // A paragraph implies a single cohesive style. having multiple spans of the
+    // same type (e.g., two codeblock spans) in one paragraph is an invalid state in current library logic
+    if (spans.size > 1) {
+      Log.w("ParagraphStyles", "getPreviousParagraphSpan(): Found more than one span in the paragraph!")
+    }
+
     if (spans.isNotEmpty()) {
       return spans.first()
     }
@@ -30,6 +37,12 @@ class ParagraphStyles(private val view: EnrichedTextInputView) {
     val (nextParagraphStart, nextParagraphEnd) = spannable.getParagraphBounds(paragraphEnd + 1)
 
     val spans = spannable.getSpans(nextParagraphStart, nextParagraphEnd, type)
+
+    // A paragraph implies a single cohesive style. having multiple spans of the
+    // same type (e.g., two codeblock spans) in one paragraph is an invalid state in current library logic
+    if (spans.size > 1) {
+      Log.w("ParagraphStyles", "getNextParagraphSpan(): Found more than one span in the paragraph!")
+    }
 
     if (spans.isNotEmpty()) {
       return spans.first()
@@ -65,7 +78,7 @@ class ParagraphStyles(private val view: EnrichedTextInputView) {
 
 
   private fun <T>setSpan(spannable: Spannable, type: Class<T>, start: Int, end: Int) {
-    if (type == EnrichedBlockQuoteSpan::class.java || type == EnrichedCodeBlockSpan::class.java) {
+    if (EnrichedSpans.isTypeContinuous(type)) {
       setContinuousSpan(spannable, start, end, type)
       return
     }
@@ -248,8 +261,8 @@ class ParagraphStyles(private val view: EnrichedTextInputView) {
 
     if (start == end) {
       spannable.insert(start, "\u200B")
-      view.spanState?.setStart(name, start + 1)
       setAndMergeSpans(spannable, type, start, end + 1)
+      view.selection.validateStyles()
 
       return
     }
@@ -264,8 +277,8 @@ class ParagraphStyles(private val view: EnrichedTextInputView) {
       currentStart = currentEnd + 1
     }
 
-    view.spanState?.setStart(name, start)
     setAndMergeSpans(spannable, type, start, currentEnd)
+    view.selection.validateStyles()
   }
 
   fun getStyleRange(): Pair<Int, Int> {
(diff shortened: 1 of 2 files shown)


android/src/main/java/com/swmansion/enriched/styles/ParagraphStyles.kt
```diff
@@ -94,14 +154,48 @@ class ParagraphStyles(private val view: EnrichedTextInputView) {
...
     val isBackspace = s.length < previousTextLength
     val isNewLine = endCursorPosition == 0 || endCursorPosition > 0 && s[endCursorPosition - 1] == '\n'
 
     for ((style, config) in EnrichedSpans.paragraphSpans) {
       val spanState = view.spanState ?: continue
-      val styleStart = spanState.getStart(style) ?: continue
+      val styleStart = spanState.getStart(style)
+
+      if (styleStart == null) {
+        if (config.isContinuous) {
```
exploIF, line 207: Can we merge those two?

android/src/main/java/com/swmansion/enriched/styles/ParagraphStyles.kt
```diff
@@ -4,12 +4,72 @@ import android.text.Editable
...
 class ParagraphStyles(private val view: EnrichedTextInputView) {
+  private fun <T>getPreviousParagraphSpan(spannable: Spannable, paragraphStart: Int, type: Class<T>): T? {
+    if (paragraphStart <= 0) return null
+
+    val (previousParagraphStart, previousParagraphEnd) = spannable.getParagraphBounds(paragraphStart - 1)
+    val spans = spannable.getSpans(previousParagraphStart, previousParagraphEnd, type)
+
+    if (spans.isNotEmpty()) {
+      return spans.first()
+    }
```
exploIF, line 29: What if there are more spans there? There is no guarantee that `spans.first()` is valid one. Shouldn't we iterate over spans and find this one which end is the biggest one?

android/src/main/java/com/swmansion/enriched/styles/ParagraphStyles.kt
```diff
@@ -4,12 +4,72 @@ import android.text.Editable
...
+  private fun <T>getNextParagraphSpan(spannable: Spannable, paragraphEnd: Int, type: Class<T>): T? {
+    if (paragraphEnd >= spannable.length - 1) return null
+
+    val (nextParagraphStart, nextParagraphEnd) = spannable.getParagraphBounds(paragraphEnd + 1)
+
+    val spans = spannable.getSpans(nextParagraphStart, nextParagraphEnd, type)
+
+    if (spans.isNotEmpty()) {
+      return spans.first()
+    }
```
exploIF, line 49: And here, shouldn't we find span with smallest span start?

exploIF:
I've noticed one issue with current logic, see video below


https://github.com/user-attachments/assets/7992cfba-2438-4cdd-863d-a5660f5d6e25

android/src/main/java/com/swmansion/enriched/styles/ParagraphStyles.kt
```diff
@@ -4,12 +4,72 @@ import android.text.Editable
...
+
+    return null
+  }
+
+  /**
+   * Applies a continuous span to the specified range.
+   * If the new range touches existing continuous spans, they are coalesced into a single span
+   */
+  private fun <T>setContinuousSpan(spannable: Spannable, start: Int, end: Int, type: Class<T>) {
+    val span = type.getDeclaredConstructor(HtmlStyle::class.java).newInstance(view.htmlStyle)
```
exploIF, line 59: This doesn't look correct. We are not using `HtmlStyle` as a span anywhere, shouldn't this be a class of type `Class<T>`?

android/src/main/java/com/swmansion/enriched/styles/ParagraphStyles.kt
```diff
@@ -4,12 +4,72 @@ import android.text.Editable
...
+      spannable.removeSpan(nextSpan)
+    }
+
+    val (safeStart, safeEnd) = spannable.getSafeSpanBoundaries(newStart, newEnd)
+    spannable.setSpan(span, safeStart, safeEnd, Spannable.SPAN_EXCLUSIVE_EXCLUSIVE)
+  }
+
+
   private fun <T>setSpan(spannable: Spannable, type: Class<T>, start: Int, end: Int) {
+    if (type == EnrichedBlockQuoteSpan::class.java || type == EnrichedCodeBlockSpan::class.java) {
```
exploIF: Instead of `EnrichedBlockQuoteSpan::class.java || type == EnrichedCodeBlockSpan::class.java`, shouldn't we check if span is type of continuous? It's available in `EnrichedSpans` config

exploIF (changes requested):

android/src/main/java/com/swmansion/enriched/styles/ParagraphStyles.kt
kacperzolkiewski: But it was created like this inside `setSpan` even before my changes?
[here](https://github.com/software-mansion/react-native-enriched/blob/6aad08e0c585f8fcd88d1d7d19b51a2720f933d2/android/src/main/java/com/swmansion/enriched/styles/ParagraphStyles.kt#L13)


I thought that it is the correct way of creating span.

exploIF: Oh yeah, my bad, it's good :sorry 😔

android/src/main/java/com/swmansion/enriched/styles/ParagraphStyles.kt
kacperzolkiewski: But we can only have one paragraph style applied in one paragraph, so it will always be first?

exploIF: Based on current app logic yes, but I would prefer to have more bulletproof solution. Afaik, there is no technical limitation to have many paragraph spans applied to single line

android/src/main/java/com/swmansion/enriched/styles/ParagraphStyles.kt
kacperzolkiewski: No I cannot merge them without breaking the current logic

android/src/main/java/com/swmansion/enriched/styles/ParagraphStyles.kt
kacperzolkiewski: I created additional method `isContinuous` inside `EnrichedSpans` which handles that check

android/src/main/java/com/swmansion/enriched/styles/ParagraphStyles.kt
kacperzolkiewski: I added a safety check here. Since we don't support multiple paragraph styles of the same type within a single paragraph, I now log a warning if this invalid state is detected.

PR #310, opened 2025-12-08
feat: get HTML value imperatively

qoobes:
# Summary

Adds the ability to use a getHTML method to get the current state of the editor

This is incredibly important to my team since we're adding heavy usage of this library into our app and launching the update in about two weeks. The current performance with onChangeHTML is inadequate for us, so this change is very important.

## Test Plan

Simplest options is to create an enriched input with a ref, and a button below it that triggers a function which calls getHTML() and logs it out to the console. You should see the logged HTML.
## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

### DIFF
--- a/ios/EnrichedTextInputView.mm
+++ b/ios/EnrichedTextInputView.mm
@@ -1082,13 +1082,11 @@ Class<RCTComponentViewProtocol> EnrichedTextInputViewCls(void) {
       NSString *htmlOutput = [parser
           parseToHtmlFromRange:NSMakeRange(0,
                                            textView.textStorage.string.length)];
-      emitter->onRequestHtmlResult({
-          .requestId = static_cast<int>(requestId),
-          .html = [htmlOutput toCppString]});
+      emitter->onRequestHtmlResult({.requestId = static_cast<int>(requestId),
+                                    .html = [htmlOutput toCppString]});
     } @catch (NSException *exception) {
-      emitter->onRequestHtmlResult({
-          .requestId = static_cast<int>(requestId),
-          .html = folly::dynamic(nullptr)});
+      emitter->onRequestHtmlResult({.requestId = static_cast<int>(requestId),
+                                    .html = folly::dynamic(nullptr)});
     }
   }
 }


qoobes:
hi @exploIF @szydlovsky, have you had a chance to review this one yet? Let me know if you have any questions and thanks for the awesome project!

szydlovsky:
Hey @qoobes not yet but it will definitely be checked out in the upcoming days 😄

android/src/main/java/com/swmansion/enriched/EnrichedTextInputViewManager.kt
```diff
@@ -268,6 +272,22 @@ class EnrichedTextInputViewManager : SimpleViewManager<EnrichedTextInputView>(),
...
 
+  override fun requestHTML(view: EnrichedTextInputView?, requestId: Int) {
+    if (view == null) return
+
+    val spannable = view.text as? android.text.Spannable
+    val html = if (spannable != null) {
+      EnrichedParser.toHtml(spannable)
+    } else {
+      "<html>\n<p></p>\n</html>"
+    }
```
exploIF: I tink it will be easier to maintain if we have such logic handled directly in the `EnrichedParser`. We can have `toHtmlWithDefault` method which returns default html if view is text is either null or not spannable

android/src/main/java/com/swmansion/enriched/EnrichedTextInputViewManager.kt
```diff
@@ -268,6 +272,22 @@ class EnrichedTextInputViewManager : SimpleViewManager<EnrichedTextInputView>(),
     view?.addMention(text, indicator, attributes)
   }
 
+  override fun requestHTML(view: EnrichedTextInputView?, requestId: Int) {
```
exploIF, line 273: Can we move this logic directly to `EnrichedTextInputView` class? It's just for consistency, most of the `EnrichedTextInputViewManager` methods are entry points for other methods

android/src/main/java/com/swmansion/enriched/EnrichedTextInputViewManager.kt
```diff
@@ -268,6 +272,22 @@ class EnrichedTextInputViewManager : SimpleViewManager<EnrichedTextInputView>(),
...
+    if (view == null) return
+
+    val spannable = view.text as? android.text.Spannable
```
exploIF:
```suggestion
    val spannable = view.text as? Spannable
```

Also, remember to add missing (after such change) import

android/src/main/java/com/swmansion/enriched/events/OnRequestHtmlResultEvent.kt
```diff
@@ -0,0 +1,28 @@
+package com.swmansion.enriched.events
+
+import com.facebook.react.bridge.Arguments
+import com.facebook.react.bridge.WritableMap
+import com.facebook.react.uimanager.events.Event
+
+class OnRequestHtmlResultEvent(
```
exploIF, line 7: We have experimental support for synchronous events, this is not working fully reliably (as for now). But in general all enriched android events override `experimental_isSynchronous`, could you add such implementation for this method too? You can check how it's done for other event types

src/EnrichedTextInput.tsx
```diff
@@ -229,6 +234,13 @@ export const EnrichedTextInput = ({
     setValue: (value: string) => {
       Commands.setValue(nullthrows(nativeRef.current), value);
     },
+    getHTML: () => {
+      return new Promise<string>((resolve) => {
+        const requestId = nextRequestId++;
+        pendingHtmlRequests.set(requestId, resolve);
```
exploIF: Imo, we should store not only `resolve` but also `reject`, so we have a way to reject promise if something goes wrong

src/EnrichedTextInput.tsx
```diff
@@ -162,6 +164,9 @@ const warnAboutMissconfiguredMentions = (indicator: string) => {
   );
 };
 
+let nextRequestId = 1;
+const pendingHtmlRequests = new Map<number, (html: string) => void>();
```
exploIF: I think it would be better to store this values directly inside component (using `useRef`). Once it's done like that, we should do a cleanup on component unmount -> every pending promise should be rejected

android/src/main/java/com/swmansion/enriched/EnrichedTextInputViewManager.kt
```diff
@@ -268,6 +272,22 @@ class EnrichedTextInputViewManager : SimpleViewManager<EnrichedTextInputView>(),
...
+    val html = if (spannable != null) {
+      EnrichedParser.toHtml(spannable)
```
exploIF: Technically it may throw error. It would be the best to wrap this code in try/catch. If it fails we can still emit `OnRequestHtmlResultEvent`, but with some specific payload, like null. Then on the JS side, if parsed text equals null we already now that error occurred, so we can reject pending promise.

This should be handled for both platforms

src/EnrichedTextInput.tsx
qoobes: Good point yeah

qoobes:
@exploIF Thanks for the review! 

I've implemented the changes you asked for, and verified it still works well in my implementation- let me know if there's anything else i should change

exploIF:
@qoobes I think you forgot to push new changes. I don't see them

src/EnrichedTextInput.tsx
```diff
@@ -162,6 +164,9 @@ const warnAboutMissconfiguredMentions = (indicator: string) => {
   );
 };
 
+let nextRequestId = 1;
```
szydlovsky: Maybe
```suggestion
let nextHtmlRequestId = 1;
```
for clarity

kacperzolkiewski:
Android version also works properly. Nice work 🚀

qoobes:
@exploIF @szydlovsky  sorry! Pushed now

src/EnrichedTextInputNativeComponent.ts
```diff
@@ -64,6 +64,11 @@ export interface OnChangeSelectionEvent {
   text: string;
 }
 
+export interface OnRequestHtmlResultEvent {
+  requestId: Int32;
+  html: UnsafeMixed;
```
IvanIhnatsiuk, line 69: Should it be string?

src/EnrichedTextInput.tsx
```diff
@@ -193,6 +196,23 @@ export const EnrichedTextInput = ({
   ...rest
 }: EnrichedTextInputProps) => {
   const nativeRef = useRef<ComponentType | null>(null);
+  const nextRequestIdRef = useRef(1);
```
IvanIhnatsiuk:
```suggestion
  const nextHtmlRequestID = useRef(1);
```

android/src/main/java/com/swmansion/enriched/EnrichedTextInputViewManager.kt
```diff
@@ -71,6 +71,7 @@ class EnrichedTextInputViewManager : SimpleViewManager<EnrichedTextInputView>(),
      map.put(OnMentionDetectedEvent.EVENT_NAME, mapOf("registrationName" to OnMentionDetectedEvent.EVENT_NAME))
      map.put(OnMentionEvent.EVENT_NAME, mapOf("registrationName" to OnMentionEvent.EVENT_NAME))
      map.put(OnChangeSelectionEvent.EVENT_NAME, mapOf("registrationName" to OnChangeSelectionEvent.EVENT_NAME))
+     map.put(OnRequestHtmlResultEvent.EVENT_NAME, mapOf("registrationName" to OnRequestHtmlResultEvent.EVENT_NAME))
```
exploIF, line 75: Looks like import is missing

src/EnrichedTextInputNativeComponent.ts
exploIF: No, it shouldn't. In case of native failure null is returned which is later handled as a promise rejection on the js side

src/EnrichedTextInput.tsx
exploIF: Let's just remove `ref` suffix

exploIF (approved):
Thank you so much for your contribution!!! I've pushed couple tiny fixes, and it should be ready to be merged :fireworks:

src/EnrichedTextInputNativeComponent.ts
IvanIhnatsiuk: Okay, Then it has to be 

```.ts
export interface OnRequestHtmlResultEvent {
  requestId: Int32;
  html: string | null;
}
```

src/EnrichedTextInput.tsx
IvanIhnatsiuk: Since react-compiler was introduced, I would reccomend to keep the ref suffix because react-compiler uses the ref suffix

https://github.com/facebook/react/commit/83d538e0d04aac2a32d335e6b3963273729a9fe6

src/EnrichedTextInputNativeComponent.ts
exploIF: No, that's not a valid type for iOS codegen. Feel free to check it, it's not building at all with types defined that way

szydlovsky: @IvanIhnatsiuk codegen cannot process null/undefined values correctly. We need to use `unsafeMixed` with nullable values.

src/EnrichedTextInput.tsx
exploIF: @IvanIhnatsiuk but your comment suggested to remove `ref` suffix...

exploIF: Also, based on the RN PR description, it's redundant when using `useRef` directly and ref is not exposed. Linked PR affects only custom hooks

src/EnrichedTextInputNativeComponent.ts
IvanIhnatsiuk: @szydlovsky , @exploIT, if I see it. It seems that this allows you to build a project and pass nullptr values. However, an exception will be thrown during execution indicating that this value is equal to nullptr. My main concern was about typing at runtime and allocating memory while working with huge HTML. But there doesn't seem to be any differences (very interesting)

Issue #204, opened 2025-10-17
Allow setSelection on Ref


temlabs:
When using the focus() method on the ref, it consistently tries to go to the location of the last link (<a>...</a>) if I have used the setValue() method, which is frustrating because that may not be at the end.

It would be great to have control over where the cursor goes, so that we can focus on the end, or focus after some text was inserted using the setValue method.

This is more similar to the native Text component which exposes setSelection on its ref.

temlabs:
(Thanks so much for the amazing work!)

exploIF:
This has been implemented in https://github.com/software-mansion/react-native-enriched/pull/311

temlabs:
Beautiful!

PR #311, opened 2025-12-08
feat: add setSelection method

IslamRustamov:
# Summary

Explain the **motivation** for making this change: here are some points to help you:

- This PR solves https://github.com/software-mansion/react-native-enriched/issues/204
- This PR add a new method called **setSelection** that allows user to programmatically set selection on text input
- Added new types, added function implementation on Android and iOS
- This PR impacts area of available ref methods

## Test Plan

1) Enter text
2) Call ref.current?.setSelection(1, 2)
3) See that text got selected

## Screenshots / Videos

Not applicable

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

### DIFF
--- a/docs/API_REFERENCE.md
+++ b/docs/API_REFERENCE.md
@@ -87,16 +87,16 @@ interface OnChangeHtmlEvent {
 }
 ```
 
-> [!TIP]
-> Specifying `onChangeHtml` may have performance implications, especially with large documents, as it requires continuous HTML parsing.
-> If you only need the HTML content at specific moments (e.g., when saving), consider using the [`getHTML`](#gethtml) ref method instead.
-> When `onChangeHtml` is not provided, the component optimizes performance by avoiding unnecessary HTML parsing.
-
 - `value` is the new HTML.
 
 | Type                                                       | Default Value | Platform |
 |------------------------------------------------------------|---------------|----------|
 | `(event: NativeSyntheticEvent<OnChangeHtmlEvent>) => void` | -             | Both     |
+
+> [!TIP]
+> Specifying `onChangeHtml` may have performance implications, especially with large documents, as it requires continuous HTML parsing.
+> If you only need the HTML content at specific moments (e.g., when saving), consider using the [`getHTML`](#gethtml) ref method instead.
+> When `onChangeHtml` is not provided, the component optimizes performance by avoiding unnecessary HTML parsing.
 
 ### `onChangeMention`
 
@@ -410,6 +410,16 @@ setValue: (value: string) => void;
 Sets the input's value.
 
 - `value: string` - value to set, it can either be `react-native-enriched` supported HTML string or raw text.
+
+### `.setSelection()`
+
+```ts
+setSelection: (start: number, end: number) => void;
+```
+
+Sets the selection at the given indexes.
+- `start: number` - starting index of the selection.
+- `end: number` - first index after the selection's ending index. For just a cursor in place (no selection), `start` equals `end`.
 
 ### `.startMention()`
 


szydlovsky:
Hey @IslamRustamov thanks for the contribution! I'll have to address a few matters first.
We definitely want the API to work exactly the same on both platforms, so we will most likely use the Android approach as other selection topics have been handled this way.
Except for that, I want to know your exact use case for the feature. I'm asking it because iOS might have some inconsistencies - emitted `text` isn't always exactly what the input holds because we sometimes put some zero-width spaces in lines with empty lists, blockquotes, codeblocks (so that they can be drawn), but we don't emit them. This way selection might be different than what you expect because, for example, some zero-width space got selected too on the way. Ideally there could just be a check that extends the selection accordingly if any zero-width spaces were in the range of the `setSelection` method, if you're interested I can just suggest it in the review.

IslamRustamov:
Ah I didn't know that you were using unseen chars in text.

As for use case - nothing in particular, just wanted to contribute something and saw this issue https://github.com/software-mansion/react-native-enriched/issues/204 that required this exact feature.

If you want to develop this feature yourself or don't consider it needed for now - feel free to close this PR, or if you want me to continue working on it then let me know.

szydlovsky:
I see. Honestly, this PR isn't far from being ready. If you just made sure both platforms treat the selection in the same (Android) way it would be great. You can do it similarly on iOS to how `addLink` does that: https://github.com/software-mansion/react-native-enriched/blob/c7a7a2ab7aebd99a80f0dfbb8b4c85cacf2d845f/ios/EnrichedTextInputView.mm#L928-L938
And with the unseen chars on iOS (afaik Android won't have that problem), all that's needed is to check the text in the range to be selected. Then extend the selection by 1 length for each `\u200B` (zero-width-space character). Make sure the selection won't go outside of the input range ;)

We really appreciate meaningful contributions and I don't want to take it from you by basing a PR on your wok.

IslamRustamov:
@szydlovsky I believe that does it. If I understood correctly how you want zero width spaces handled. Tested on lists and it seemed to work.

This is with zero width spaces in mind:
Screenshot 2025-12-11 at 00 23 27

And this is not:
Screenshot 2025-12-11 at 00 24 22

Selection params are **start === 2** and  **end === 6.**

ios/EnrichedTextInputView.mm
```diff
@@ -802,6 +802,11 @@ - (void)handleCommand:(const NSString *)commandName args:(const NSArray *)args {
     CGFloat imgHeight = [(NSNumber*)args[2] floatValue];
     
     [self addImage:uri width:imgWidth height:imgHeight];
+  } else if([commandName isEqualToString:@"setSelection"]) {
+    NSInteger start = [((NSNumber*)args[0]) integerValue];
+    NSInteger end = [((NSNumber*)args[1]) integerValue];
+
```
szydlovsky:
```suggestion
```

ios/EnrichedTextInputView.mm
```diff
@@ -837,6 +842,12 @@ - (void)setValue:(NSString *)value {
   [self anyTextMayHaveBeenModified];
 }
 
+- (void)setCustomSelection:(NSInteger)start end:(NSInteger)end {
+  NSInteger amountOfZeroWidthSpaces = [ZeroWidthSpaceUtils getAmountOfZeroWidthSpacesInRange:NSRange(start, end - start) text:textView.textStorage.string];
+    
+  textView.selectedRange = NSRange(start, end + amountOfZeroWidthSpaces - start);
```
szydlovsky: We rather use `NSMakeRange`. Zero width spaces are countable so it should be `number`, not `amount`.
```suggestion
  NSInteger numberOfZeroWidthSpaces = [ZeroWidthSpaceUtils getNumberOfZeroWidthSpacesInRange:NSMakeRange(start, end - start) text:textView.textStorage.string];
    
  textView.selectedRange = NSMakeRange(start, end + numberOfZeroWidthSpaces - start);
```

ios/utils/ZeroWidthSpaceUtils.mm
```diff
@@ -5,6 +5,24 @@
 #import "UIView+React.h"
 
 @implementation ZeroWidthSpaceUtils
++ (NSInteger)getAmountOfZeroWidthSpacesInRange:(NSRange)range text:(NSString*)text {
```
szydlovsky:
```suggestion
+ (NSInteger)getNumberOfZeroWidthSpacesInRange:(NSRange)range text:(NSString*)text {
```

ios/utils/ZeroWidthSpaceUtils.mm
```diff
@@ -5,6 +5,24 @@
...
+  if (range.length > [text length]) {
```
szydlovsky:
```suggestion
  if (range.length > text.length) {
```

src/EnrichedTextInput.tsx
```diff
@@ -37,6 +37,7 @@ export interface EnrichedTextInputInstance extends NativeMethods {
   focus: () => void;
   blur: () => void;
   setValue: (value: string) => void;
+  setSelection: (start: number, length: number) => void;
```
szydlovsky: Let's keep the same naming convention as in `setLink`
```suggestion
  setSelection: (start: number, end: number) => void;
```

src/EnrichedTextInput.tsx
```diff
@@ -293,6 +294,9 @@ export const EnrichedTextInput = ({
 
       Commands.startMention(nullthrows(nativeRef.current), indicator);
     },
+    setSelection: (start: number, length: number) => {
```
szydlovsky:
```suggestion
    setSelection: (start: number, end: number) => {
```

src/EnrichedTextInput.tsx
```diff
@@ -293,6 +294,9 @@ export const EnrichedTextInput = ({
...
+      Commands.setSelection(nullthrows(nativeRef.current), start, length);
```
szydlovsky:
```suggestion
      Commands.setSelection(nullthrows(nativeRef.current), start, end);
```

src/EnrichedTextInputNativeComponent.ts
```diff
@@ -166,6 +166,11 @@ interface NativeCommands {
   focus: (viewRef: React.ElementRef<ComponentType>) => void;
   blur: (viewRef: React.ElementRef<ComponentType>) => void;
   setValue: (viewRef: React.ElementRef<ComponentType>, text: string) => void;
+  setSelection: (
+    viewRef: React.ElementRef<ComponentType>,
+    start: Int32,
+    length: Int32
```
szydlovsky:
```suggestion
    end: Int32
```

src/EnrichedTextInputNativeComponent.ts
```diff
@@ -166,6 +166,11 @@ interface NativeCommands {
...
+  ) => string;
```
szydlovsky: Does it return anything?
```suggestion
  ) => void;
```

IslamRustamov: No, sorry, it was a leftover of me trying to implement an imperative way to extract text from the input

IslamRustamov:
@szydlovsky made the changes, check please if everything is to your liking

szydlovsky:
@IslamRustamov great, will check it out on Monday, stay tuned!

exploIF (commented):
Looks good to me in general. However, there is one missing thing. Similar to iOS part, we should include zero width spaces in selection range calculations  on Android

kacperzolkiewski:
I added including zero width spaces on Android `setSelection(2, 6)` result:
image

PR #314, opened 2025-12-09
fix(iOS): do not set the selection on the link when applying the initial styles

IvanIhnatsiuk:
# Summary

Added `withSelection` argument for `addLink` method to not set selection when we are applying initial input.

Basically, it solves 2 issues

1. When we initially apply styles, it will always set the selection to the latest link in HTML
2. If we replace the HTML value, it will execute the code in the following order

`applyProcessedStyles -> addLink -> setSelection -> textViewDidChangeSelection -> tryUpdatingActiveStyles`

It is not a big problem, but if we process a really huge HTML that contains a lot of links, it will be a bottleneck

## Test Plan

1. Set the default value like: `<html><h1>Link test</h1><p><a href="test.com">test link</a></p></html>`
2. Open the app
3. Observe that the focus is set correctly

### Before:

https://github.com/user-attachments/assets/38d9593a-7a6d-4dcb-976e-0834893384fa

### After:

https://github.com/user-attachments/assets/7126719e-0e73-42de-b42e-d8788f04ff52

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

### DIFF
--- a/README.md
+++ b/README.md
@@ -238,14 +238,10 @@ See the [API Reference](docs/API_REFERENCE.md) for a detailed overview of all th
 ## Known limitations
 
 - Only one level of lists is supported. We currently do not support nested lists.
-- Inline images are supported only on Android.
-- Codeblocks are supported only on Android.
 - iOS headings can't have the same `fontSize` in their config as input's `fontSize`. Doing so results in incorrect headings behavior.
 
 ## Future Plans
 
-- Adding Codeblocks and Inline Images to iOS input.
-- Making some optimizations around `onChangeHtml` event, maybe some imperative API to get the HTML output.
 - Creating `EnrichedText` text component that supports our HTML output format with all additional interactions like pressing links or mentions.
 - Adding API for custom link detection regex.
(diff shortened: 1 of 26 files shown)


ios/styles/LinkStyle.mm
```diff
@@ -357,7 +357,7 @@ - (void)handleAutomaticLinks:(NSString *)word inRange:(NSRange)wordRange {
       }
     }
     if(addStyle) {
-      [self addLink:word url:regexPassedUrl range:wordRange manual:NO];
+      [self addLink:word url:regexPassedUrl range:wordRange manual:NO withSelection:YES];
```
szydlovsky: These are automatic links. We don't want to use selection with them
```suggestion
      [self addLink:word url:regexPassedUrl range:wordRange manual:NO withSelection:NO];
```

szydlovsky (changes requested):
Just one thing and it works super fine, nice find!

PR #316, opened 2025-12-10
chore: clang setup

exploIF:
# Summary

Scope:
- removed separate clang format workflow
- fixed setup to use local (cross machine consistent) version of clang-format
- integrated lint-clang with current CI workflow
- integrated lint-clang check with lefthook
- run `lint-clang:fix` to format files

## Test Plan

n/a

## Screenshots / Videos

n/a

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |


android/src/main/new_arch/RNEnrichedTextInputViewSpec.cpp
```diff
@@ -1,9 +1,10 @@
 
 /**
- * This code was generated by [react-native-codegen](https://www.npmjs.com/package/react-native-codegen).
```
exploIF, line 3: I will remove this comments in the future. They are not relevant for us

szydlovsky (approved):
🚢

PR #318, opened 2025-12-10
feat(iOS): async image loading

IvanIhnatsiuk:
# Summary

Currently, all images on iOS are loaded in sync way. It fully blocks the main thread and lets's say if we have a large image user has to wait until this image is loaded. 

The best approach for images is to always load them asynchronously. If it's a local image -> it will be immediately loaded, it won't block the main thread.

Additionally, I added the MediaAttachment class that implements the abstract presentation of all attachments. So, if in the future if you would like to add video/block image attachments, it would be easier to extend from this class

## Test Plan
Test case 1:

1. Insert any image manually

Test case 2:

1. set the initial html like:
```
<html>
<p><img src="https://picsum.photos/id/1/200/200" width="200.000000" height="200.000000"/></p>
<p><img src="https://picsum.photos/id/2/200/200" width="200.000000" height="200.000000"/></p>
<p><img src="https://picsum.photos/id/3/200/200" width="200.000000" height="200.000000"/></p>
<p><img src="https://picsum.photos/id/4/200/200" width="200.000000" height="200.000000"/></p>
<p><img src="https://picsum.photos/id/5/200/200" width="200.000000" height="200.000000"/></p>
<p><img src="https://picsum.photos/id/6/200/200" width="200.000000" height="200.000000"/></p>
<p><img src="https://picsum.photos/id/7/200/200" width="200.000000" height="200.000000"/></p>
<p><img src="https://picsum.photos/id/8/200/200" width="200.000000" height="200.000000"/></p>
<p><img src="https://picsum.photos/id/9/200/200" width="200.000000" height="200.000000"/></p>
<p><img src="https://picsum.photos/id/10/200/200" width="200.000000" height="200.000000"/></p>
<p><img src="https://picsum.photos/id/11/200/200" width="200.000000" height="200.000000"/></p>
<p><img src="https://picsum.photos/id/12/200/200" width="200.000000" height="200.000000"/></p>
<p><img src="https://picsum.photos/id/13/200/200" width="200.000000" height="200.000000"/></p>
<p><img src="https://picsum.photos/id/14/200/200" width="200.000000" height="200.000000"/></p>
<p><img src="https://picsum.photos/id/15/200/200" width="200.000000" height="200.000000"/></p>
<p><img src="https://picsum.photos/id/16/200/200" width="200.000000" height="200.000000"/></p>
<p><img src="https://picsum.photos/id/17/200/200" width="200.000000" height="200.000000"/></p>
<p><img src="https://picsum.photos/id/18/200/200" width="200.000000" height="200.000000"/></p>
<p><img src="https://picsum.photos/id/19/200/200" width="200.000000" height="200.000000"/></p>
<p><img src="https://picsum.photos/id/20/200/200" width="200.000000" height="200.000000"/></p>
</html>
```

### Before:

https://github.com/user-attachments/assets/f2a60c27-6d1a-4cd8-ba56-b44d2f9813be

### After:

https://github.com/user-attachments/assets/837aa5c9-aa35-49c7-ada2-a18b9b4116da

https://github.com/user-attachments/assets/1c32ccc6-fc5a-4caa-8225-5770122e502c

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

### DIFF
--- a/ios/EnrichedTextInputView.mm
+++ b/ios/EnrichedTextInputView.mm
@@ -29,7 +29,7 @@ using namespace facebook::react;
   NSMutableSet<NSNumber *> *_activeStyles;
   LinkData *_recentlyActiveLinkData;
   NSRange _recentlyActiveLinkRange;
-  NSString *_recentlyEmittedString;
+  NSString *_recentInputString;
   MentionParams *_recentlyActiveMentionParams;
   NSRange _recentlyActiveMentionRange;
   NSString *_recentlyEmittedHtml;
@@ -75,7 +75,7 @@ Class<RCTComponentViewProtocol> EnrichedTextInputViewCls(void) {
   _recentlyActiveLinkRange = NSMakeRange(0, 0);
   _recentlyActiveMentionRange = NSMakeRange(0, 0);
   recentlyChangedRange = NSMakeRange(0, 0);
-  _recentlyEmittedString = @"";
+  _recentInputString = @"";
   _recentlyEmittedHtml = @"<html>\n<p></p>\n</html>";
   _emitHtml = NO;
   blockEmitting = NO;
@@ -1128,7 +1128,11 @@ Class<RCTComponentViewProtocol> EnrichedTextInputViewCls(void) {
   NSRange linkRange = NSMakeRange(start, end - start);
   if ([self handleStyleBlocksAndConflicts:[LinkStyle getStyleType]
                                     range:linkRange]) {
-    [linkStyleClass addLink:text url:url range:linkRange manual:YES];
+    [linkStyleClass addLink:text
+                        url:url
+                      range:linkRange
+                     manual:YES
+              withSelection:YES];
     [self anyTextMayHaveBeenModified];
   }
 }
@@ -1253,7 +1257,7 @@ Class<RCTComponentViewProtocol> EnrichedTextInputViewCls(void) {
     // mention editing runs if only a selection was done (no text change)
     // otherwise we would double-emit with a second call in the
     // anyTextMayHaveBeenModified method
-    if ([_recentlyEmittedString
+    if ([_recentInputString
             isEqualToString:[textView.textStorage.string copy]]) {
       [mentionStyleClass manageMentionEditing];
     }
@@ -1262,7 +1266,7 @@ Class<RCTComponentViewProtocol> EnrichedTextInputViewCls(void) {
   // typing attributes for empty lines selection reset
   NSString *currentString = [textView.textStorage.string copy];
   if (textView.selectedRange.length == 0 &&
-      [_recentlyEmittedString isEqualToString:currentString]) {
+      [_recentInputString isEqualToString:currentString]) {
     // no string change means only a selection changed with no character changes
     NSRange paragraphRange = [textView.textStorage.string
         paragraphRangeForRange:textView.selectedRange];
@@ -1306,7 +1310,7 @@ Class<RCTComponentViewProtocol> EnrichedTextInputViewCls(void) {
 
   // emptying input typing attributes management
   if (textView.textStorage.string.length == 0 &&
-      _recentlyEmittedString.length > 0) {
+      _recentInputString.length > 0) {
     // reset typing attribtues
     textView.typingAttributes = defaultTypingAttributes;
   }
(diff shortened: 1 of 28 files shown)


kacperzolkiewski:
Nice work 🚀
But I have one comment: 
When we are adding an image for a moment we will see the same placeholder like when the image loading failed.

https://github.com/user-attachments/assets/760096fe-a8af-4fdf-8180-9a09a928b16e


To match the same behaviour like on Android there should not be placeholder but transparent image during loading:

https://github.com/user-attachments/assets/e51a6e9c-dd1c-4619-a138-41b6dc45ef7f

IvanIhnatsiuk:
@kacperzolkiewski 

Done:


https://github.com/user-attachments/assets/483ab230-ca76-447b-94c0-3004015f8c71

kacperzolkiewski (changes requested):
Currently I found two issues: 
The cursor is before image when we provide value through `Set input's value` button:

https://github.com/user-attachments/assets/d0da2cc4-b3d0-4659-8d5c-8e3567fb2e7e


When src is wrong we do not display error placeholder like on Android but only empty space (user does not see that something went wrong):

https://github.com/user-attachments/assets/d1710404-f243-4c95-9975-b9ce5118a130


How it looks on Android:

https://github.com/user-attachments/assets/60f5f7af-b9ee-4fe2-a83a-4614f99c6a14

IvanIhnatsiuk:
@kacperzolkiewski 

1. This issue exists on the main branch, and I guess it should be fixed separately
2. Now it's working the same as on the main branch :)

kacperzolkiewski:
@IvanIhnatsiuk,

1. Yes I agree this issue also occur on the main branch, and should be fixed in another PR. I've just created an issue for that: #329 
2. Yes, I approve. Nice work, right now it displays placeholder properly 🚀

I pushed one minor commit, to display properly placeholder when image url is nil.

ios/EnrichedTextInputView.h
```diff
@@ -11,7 +12,8 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-@interface EnrichedTextInputView : RCTViewComponentView {
+@interface EnrichedTextInputView
+    : RCTViewComponentView <MediaAttachmentDelegate> {
```
szydlovsky, line 16: Any reason why we declare that conformance here instead of in the implementation file? `EnrichedTextInputView.mm`, line 22. Would be nice to have it all in one place

ios/EnrichedTextInputView.mm
```diff
@@ -1535,4 +1535,31 @@ - (void)textViewDidChange:(UITextView *)textView {
   [self anyTextMayHaveBeenModified];
 }
 
+#pragma Media attachments delegate
```
szydlovsky: I'd just go with a `//MARK: - ` comment, pragma does nothing here

ios/EnrichedTextInputView.mm
```diff
@@ -1535,4 +1535,31 @@ - (void)textViewDidChange:(UITextView *)textView {
...
+                       foundRange = range;
+                       *stop = YES;
+                     }
+                   }];
+
+  if (foundRange.location == NSNotFound) {
+    return;
+  }
+
+  [storage edited:NSTextStorageEditedAttributes
```
szydlovsky, line 1583: Just out of curiosity - what does that function do in that case? I think I never used it before and want to understand its impact.

szydlovsky: I'm guessing it refreshes the attachment when it loads, right?

ios/styles/ImageStyle.mm
```diff
@@ -115,27 +116,26 @@ - (ImageData *)getImageDataAt:(NSUInteger)location {
...
-                                                        width:imageData.width
-                                                       height:imageData.height],
-    ImageAttributeName : imageData,
-  } mutableCopy];
+  ImageAttachment *attachment =
+      [[ImageAttachment alloc] initWithImageData:imageData];
+  attachment.delegate = _input;
 
-  // Use the Object Replacement Character for Image.
-  // This tells TextKit "something non-text goes here".
```
szydlovsky, line 128: Can we keep this comment please?

ios/EnrichedTextInputView.mm
IvanIhnatsiuk: right, we need to notify textStorage that this attachment has an updated image

ios/styles/ImageStyle.mm
IvanIhnatsiuk: Sure

ios/EnrichedTextInputView.mm
szydlovsky: Makes sense, thank you!

Issue #290, opened 2025-11-26
Rich Text Editor renders content incorrectly


oreoluwa-bs:
**Describe the bug**
The component’s rendered output does not match the expected formatting.

**To Reproduce**
1. Use the following component code:
```typescript
<EnrichedTextInput
          ref={ref}
          defaultValue={`<html>
          <h3>Hello</h3>
          <br/>
          <ul>
          <li><h3>Item One</h3></li>
          </ul>
          </html>`}
          onChangeHtml={(e) => {
            setHtml(e.nativeEvent.value);
          }}
          style={{
            width: "100%",
            height: "100%",
            fontSize: 20,
            padding: 10,
          }}
        />
``` 

**Expected behavior**
The output should appear as:
<img width="1260" height="2736" alt="Image" src="https://github.com/user-attachments/assets/0f64c18f-3d33-4287-9097-1da7f8568f52" />

**Screenshots**
The output appears as:
<img width="1260" height="2736" alt="Image" src="https://github.com/user-attachments/assets/270700b3-2ac8-4bc3-b073-a3dd2ecda6a7" />

**Device (please complete the following information):**
 - Device: iPhone Air
 - OS: iOS 26
 - Version ^0.1.6

**Additional context**

szydlovsky:
Hey @oreoluwa-bs thanks for the issue! Indeed the formatting is wrong but it probably stems from the fact that we don't support headings and lists together - they are conflicting with one another and the parser most likely didn't catch that. I'll look into it a bit if we can somehow safely handle it but keep in mind: any html that cannot be returned from the library (isn't proper in the library terms) might be displayed incorrectly.

szydlovsky:
Okay, I just did a bit more digging. H3 inside list is improper but won't ruin the display. However the html used in here has spaces (not newlines) between the following lines and that ruins the parsing. This can definitely be fixed and I'll see about it

oreoluwa-bs:
Thanks, that makes sense. The clarification about the spaces breaking parsing is useful. I didn't catch the h3 inside the list.

PR #321, opened 2025-12-11
fix(iOS): remove unnecessary spaces during html preparation

kacperzolkiewski:
# Summary
Fixes: #290 

This PR adds functionality for removing unnecessary whitespaces that breaks our html parsing.

## Test Plan

Run test case from issue: #290 

## Screenshots / Videos

https://github.com/user-attachments/assets/48547d56-a554-498e-bd90-b66f8c6015b0

https://github.com/user-attachments/assets/2c6b298b-278d-4f95-8a76-4a2117214bb9

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

### DIFF
--- a/ios/inputParser/InputParser.mm
+++ b/ios/inputParser/InputParser.mm
@@ -734,7 +734,7 @@
  *
  * APPROACH:
  * This function treats the HTML as having two distinct states:
- * 1. Structure Mode (Depth == 0): We are inside container tags (like
+ * 1. Structure Mode (Depth == 0): We are inside or between container tags (like
  * <blockquote>, <ul>, <codeblock>). In this mode whitespace and newlines are
  * considered layout artifacts and are REMOVED to prevent the parser from
  * creating unwanted spaces.


ios/inputParser/InputParser.mm
```diff
@@ -726,6 +728,75 @@ - (NSString *_Nullable)initiallyProcessHtml:(NSString *_Nonnull)html {
   return fixedHtml;
 }
 
+- (NSString *)removeWhiteSpacesFromHtml:(NSString *)html {
+  NSSet *textTags = [NSSet setWithObjects:@"p", @"h1", @"h2", @"h3", @"h4",
```
exploIF: Is there any way to do this more programmatically? I'm just afraid we will miss it when adding new styles

szydlovsky: Afaik `h4`, `h5`, `h6` isn't there. Also, what is `nil` doing in this array?

ios/inputParser/InputParser.mm
```diff
@@ -726,6 +728,75 @@ - (NSString *_Nullable)initiallyProcessHtml:(NSString *_Nonnull)html {
   return fixedHtml;
 }
 
+- (NSString *)removeWhiteSpacesFromHtml:(NSString *)html {
```
szydlovsky: I'd really love some comment explaining what's the approach with removing the whitespaces because it's not that obvious from the get-go. Also you can add a note stating that if any new tag which can contain text is added, it also needs to be inserted into the `textTags` set.

ios/inputParser/InputParser.mm
```diff
@@ -726,6 +728,75 @@ - (NSString *_Nullable)initiallyProcessHtml:(NSString *_Nonnull)html {
...
+      if (isReadingTag) {
+        [currentTagBuffer appendFormat:@"%C", c];
+        [output appendFormat:@"%C", c];
+        continue;
+      }
+
+      if (textDepth > 0) {
+        [output appendFormat:@"%C", c];
+      } else {
+        if (![whitespaceSet characterIsMember:c]) {
```
szydlovsky: We also should remove newlines that are between text tags. Anyway the next parser step adds proper newlines, and they can also litter the input html. Afaik the set's name is `whitespaceAndNewlineCharacterSet`. I think the funtion name might also need to change ;)

szydlovsky (commented):
Overally good, some smaller changes and we're good to go

ios/inputParser/InputParser.mm
szydlovsky: Not really. But I asked Kacper to add a proper comment

kacperzolkiewski: We need `nil` because `setWithObjects` needs it at the end.
[setWithObjects](https://developer.apple.com/documentation/foundation/nsset/setwithobjects:?changes=_4&language=objc)

szydlovsky: Okay, did not know that, thank you Apple for self-explanatory code 😄

kacperzolkiewski: I add comment with info that we should update `textTags` set while implementing new style. 
But maybe there is a way to iterate over styles and check if it is text style -> if it is then return its tag.

ios/inputParser/InputParser.mm
kacperzolkiewski: done

ios/inputParser/InputParser.mm
kacperzolkiewski: Nice idea

ios/inputParser/InputParser.mm
```diff
@@ -728,14 +728,34 @@ - (NSString *_Nullable)initiallyProcessHtml:(NSString *_Nonnull)html {
...
-  NSSet *textTags = [NSSet setWithObjects:@"p", @"h1", @"h2", @"h3", @"h4",
-                                          @"h5", @"h6", @"li", @"b", @"a", @"s",
-                                          @"mention", @"code", @"u", @"i", nil];
+/**
+ * Prepares HTML for the parser by stripping extraneous whitespace and newlines
+ * from structural tags, while preserving them within text content.
+ *
+ * APPROACH:
+ * This function treats the HTML as having two distinct states:
+ * 1. Structure Mode (Depth == 0): We are inside container tags (like <blockquote>, <ul>, <codeblock>).
```
szydlovsky: I'd add: We are inside `or between` container tags

ios/inputParser/InputParser.mm
```diff
@@ -726,6 +728,98 @@ - (NSString *_Nullable)initiallyProcessHtml:(NSString *_Nonnull)html {
   return fixedHtml;
 }
 
+/**
+ * Prepares HTML for the parser by stripping extraneous whitespace and newlines
+ * from structural tags, while preserving them within text content.
+ *
+ * APPROACH:
+ * This function treats the HTML as having two distinct states:
+ * 1. Structure Mode (Depth == 0): We are inside container tags (like
```
szydlovsky:
```suggestion
 * 1. Structure Mode (Depth == 0): We are inside or between container tags (like
```

PR #323, opened 2025-12-11
fix(iOS): setting single attributed string to text view

IvanIhnatsiuk:
# Summary

This PR fixes many performance bottlenecks for the default value processing. 

### Applying initial styles to the textStorage

Currently, enriched input sets the initial attributed text by modifying textStorage. This is a bad approach since it causes a lot of layoutManager updates because of:
1. We collect all styles
2. Then we apply each style separatly(not wrapped in begin/endEditing).
The general workflow looked like:

ApplyStyleA -> update LayoutManager
ApplyStyleB -> update layoutManager
...
ApplyStyleX -> updateLayoutManager

### Initial pre-processing

As you can see from this stack trace, the applyProcessed styles is called multiple times. Additionally, it always triggers anyTextHaveBeenModified(which is also slow) and will be applied in the didMoveToWindow method.

Places where it was called

1. updateProps(default value changed + stylesPropHasChanged) 4 times (updateViewProps is called twice)
2. didMoveToWindow method
3. textViewDidBeginEditing(since we modified text multiple times)

In general, applyProcessed takes ~700-800ms in the old implementation and ~150m,s and only two calls(updateProps is called twice for some reason 🤔 )

#### Benchmarks old:

Screenshot 2025-12-11 at 22 31 46

#### Benchmarks new:

Screenshot 2025-12-11 at 22 10 05

### Input unmount phase

The problem is that the didMovetoWindow method is called when we remove the actual superView(window becomes nil and triggers this method). It causes an additional anyTextHaveBeenModified call.

### Solution:

I implemented class helpers to process the initial HTML and added helper utils for styles to apply attributes in the attributed string, and then set it as a single update. This also allows us to remove withAttributes from the addAttributes methods, since typing attributes are managed in a text view. This solution simplifies the next development, because devs should not care about this

### How to test?

You can use the following code snippet to test a huge HTML

```.ts
const generateHugeHtml = (repeat = 200) => {
  const parts: string[] = [];
  parts.push('<html>');

  // small helper to make deterministic colors
  const colorAt = (i: number) => {
    const r = (37 * (i + 1)) % 256;
    const g = (83 * (i + 7)) % 256;
    const b = (199 * (i + 13)) % 256;
    const toHex = (n: number) => n.toString(16).padStart(2, '0');
    return `#${toHex(r)}${toHex(g)}${toHex(b)}`;
  };

  for (let i = 0; i < repeat; i++) {
    const col = colorAt(i);
    const imgW = 200 + (i % 5) * 40;
    const imgH = 100 + (i % 3) * 30;

    parts.push(
      // Headings
      `\n<h1>Section ${i + 1}</h1>`,
      `\n<h2>Subsection ${i + 1}.1</h2>`,
      `\n<h3>Topic ${i + 1}.1.a</h3>`,

      // Paragraph with mixed inline styles
      `\n<p>This is a <b>bold</b> and <i>italic</i> paragraph with <u>underline</u>, ` +
        `<s>strike</s>, <code>inline_code_${i}</code>, ` +
        `<a href="[https://example.com/${i}](https://example.com/$%7Bi%7D)">a link ${i}</a>, ` +
        `<mention text="@alex_${i}" indicator="@"></mention>, ` +
        `<mention text="#general_${i}" indicator="#"></mention>, ` +
        `<font color="${col}">colored text ${col}</font> and some plain text to bulk it up.</p>`,

      // Line break
      `\n<br>`,

      // Unordered list
      `<ul>`,
      `<li>bullet A ${i}</li>`,
      `<li>bullet B ${i}</li>`,
      `<li>bullet C ${i}</li>`,
      `</ul>`,

      // Ordered list
      `\n<ol>`,
      `\n<li>step 1.${i}</li>`,
      `\n<li>step 2.${i}</li>`,
      `\n<li>step 3.${i}</li>`,
      `\n</ol>`,

      // Blockquote
      `\n<blockquote>`,
      `\n<p>"Blockquote line 1 for ${i}."</p>`,
      `\n<p>"Blockquote line 2 for ${i}."</p>`,
      `\n</blockquote>`,

      // Code block (escaped characters)
      `\n<codeblock>`,
      `\n<p>for (let k = 0; k < ${i % 7}; k++) { console.log(&quot;block_${i}&quot;); }</p>`,
      `\n</codeblock>`,

      // Image (self-closing)
      `\n<p><img src="[https://picsum.photos/seed/${i}/${imgW}/${imgH}](https://picsum.photos/seed/$%7Bi%7D/$%7BimgW%7D/$%7BimgH%7D)" width="${imgW}" height="${imgH}" /></p>`
    );
  }

  parts.push('\n</html>');
  return parts.join('');
};

const initialHugeHtml = generateHugeHtml();
```

### Other bottlenecks that I found

1. LayoutManager drawLists is veeeery slow
2. anyTextHaveBeenModified method (HTML parsing and managing different colors)
3. UIFont generation for bold/italic styles

## Test Plan

For example:

- Expected vs actual results

### Before:

https://github.com/user-attachments/assets/a0cd278b-0631-4806-9b75-f3393347cd35

### After:

https://github.com/user-attachments/assets/3e3b9612-c1af-4192-a834-63e822e3db67

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |


ios/EnrichedTextInputView.mm
```diff
@@ -218,6 +219,31 @@ - (void)setupPlaceholderLabel {
   _placeholderLabel.hidden = YES;
 }
 
+- (void)mediaAttachmentDidUpdate:(NSTextAttachment *)attachment {
```
IvanIhnatsiuk, line 222: I will remove this one when https://github.com/software-mansion/react-native-enriched/pull/318 will be merged

IvanIhnatsiuk:
### NOTE

needs to be revised after https://github.com/software-mansion/react-native-enriched/pull/324 + use libxml2 to process html

exploIF:
This PR brings interesting idea and potential improvements. However, it's in a draft for some time already. During this time we've provided many improvements and made huge iOS refactor, which means tons of potential conflicts and mismatches. If there is a need and idea to proceed with this logic, please create a new PR

PR #325, opened 2025-12-14
fix(iOS): draw elements only in visible area

IvanIhnatsiuk:
# Summary

Currently, we redraw elements in the whole input range. There is no sense in redrawing elements that are hidden(not visible to the user), and it impacts performance
1. Initial mount
2. Scroll performance

## Test Plan

1. Test scroll/mount on iOS

### Before

https://github.com/user-attachments/assets/43f6d113-f676-47a7-b863-b7797df4e9ca

### After

https://github.com/user-attachments/assets/e87334f1-e074-4bee-98f6-0d59768fa557

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅    |
| Android |    ❌     |

### DIFF
--- a/README.md
+++ b/README.md
@@ -238,14 +238,10 @@ See the [API Reference](docs/API_REFERENCE.md) for a detailed overview of all th
 ## Known limitations
 
 - Only one level of lists is supported. We currently do not support nested lists.
-- Inline images are supported only on Android.
-- Codeblocks are supported only on Android.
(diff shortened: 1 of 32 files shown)


IvanIhnatsiuk:
@szydlovsky could you please review this one?

szydlovsky (approved):

PR #326, opened 2025-12-15
feat(web): initialize Vite + React web example

maksymilianrojek:
# Summary

Initialize Vite + React web example

Changes:
- Created new web example in apps/example-web with React 19 and Vite
- Configured Turbo with build:web task
- Added workspace to monorepo configuration
- Updated ESLint and .gitignore for web build output
- Updated tsconfig files for monorepo setup - extracted common config options to `tsconfig.base.json`

## Test Plan

iOS/Android example app should work as before, no changes.

Running web example:
1. Install dependencies: `yarn install` in root directory
2. Start the dev server: `turbo run dev` in root or `yarn dev` in /apps/example-web directory
3. The app should now be running at `http://localhost:5173`.

## Screenshots / Videos

CleanShot 2025-12-15 at 13 34 50@2x

### DIFF
--- a/android/src/main/java/com/swmansion/enriched/EnrichedTextInputView.kt
+++ b/android/src/main/java/com/swmansion/enriched/EnrichedTextInputView.kt
@@ -30,8 +30,12 @@ import com.swmansion.enriched.events.MentionHandler
 import com.swmansion.enriched.events.OnInputBlurEvent
 import com.swmansion.enriched.events.OnInputFocusEvent
 import com.swmansion.enriched.events.OnRequestHtmlResultEvent
+import com.swmansion.enriched.spans.EnrichedH1Span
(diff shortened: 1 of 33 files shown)


apps/example-web/README.md
```diff
@@ -0,0 +1,60 @@
...
+yarn install
+```
+
+## Step 2: Start the development server
+
+Run the following command from the root of the monorepo or from the `apps/example-web` directory:
+
+```sh
+# From root - using Turbo
+turbo run dev
```
exploIF: I think we can simplify that. Currently in the root `package.json` we have `example` script, we can also add similar `example-web` script, then we can simply have

```suggestion
yarn example-web dev
```

kacperzolkiewski:
It works well 👏

PR #328, opened 2025-12-16
feat: move example to apps folder

maksymilianrojek:
# Summary

Moved example app to `apps` folder.

## Test Plan

There should be no functional changes. Example app should work as before.

### Caution
Once this one is merged, keep in mind to remove any leftovers from the root `example` directory, so uncommitted files are no longer used


kacperzolkiewski:
When I prepare `package.tgz` from the library with:
`yarn prepare`
`yarn pack`

and then I add it to my empty expo app I see those unnecessary files in `node_module`:
image

Can you check if they are generated also on your site? Maybe it is because of my local cache or something

exploIF:
@kacperzolkiewski Thanks for checking. I believe once this is merged we should ensure that `example` directory from the project root has been removed, so not committed files are not used anymore

maksymilianrojek:
The unnecessary files in node_modules were caused by the outdated `yarn.lock`.
I added to empty expo app and it works now:
CleanShot 2025-12-22 at 10 03 28@2x

Issue #329, opened 2025-12-16
[iOS] Wrong cursor position after adding image


kacperzolkiewski:
**Describe the bug**
Cursor position is applied before image, not after when applying image through `Set input's value` button.

**To Reproduce**
1. Go to Example app
2. Add `<html><p><img src="https://picsum.photos/id/1/200/200" width="200" height="200"/></p></html>` through `Set input's value` button
3. See error

**Expected behavior**
Cursor position should be applied after an image

**Screenshots**

https://github.com/user-attachments/assets/857be743-ab7c-4f92-a5f3-f3de94bb9d79

PR #337, opened 2025-12-18
fix: cursor position when inserting image

kacperzolkiewski:
# Summary
Fixes: #329 

## Test Plan
1. Go to Example app
2. Add `<html><p></p></html>` through `Set input's value` button
3. See that cursor is applied after image

## Screenshots / Videos
Before: 

https://github.com/user-attachments/assets/258132ae-1bf4-43e5-b08f-b8ade46ade67

After:

https://github.com/user-attachments/assets/6dede546-471c-489c-81e5-7e36f80377d4

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |


ios/inputParser/InputParser.mm
```diff
@@ -605,9 +605,11 @@ - (void)applyProcessedStyles:(NSArray *)processedStyles
                                                 params:params];
       } else if ([styleType isEqualToNumber:@([ImageStyle getStyleType])]) {
         ImageData *imgData = (ImageData *)stylePair.styleValue;
+        BOOL isLastInText =
+            styleRange.location + styleRange.length == plainTextLength;
         [((ImageStyle *)baseStyle) addImageAtRange:styleRange
                                          imageData:imgData
-                                     withSelection:NO];
+                                     withSelection:isLastInText];
```
IvanIhnatsiuk: @kacperzolkiewski  I'm wondering what happens if we pass the default value like:
```
<html>
<p>Some text</p>
<p><img src="test.com" width="100" height="100" /></p>
</html>
```
I think in this case it will set the selection to the end, but the selection should not be applied during the initial HTML processing or any HTML processing

Would it be better to add a check in the setValue method directly?

ios/attachments/ImageAttachment.mm
```diff
@@ -16,8 +16,10 @@ - (instancetype)initWithImageData:(ImageData *)data {
 
 - (void)loadAsync {
   NSURL *url = [NSURL URLWithString:self.uri];
-  if (!url)
+  if (!url) {
+    self.image = [UIImage systemImageNamed:@"file"];
```
kacperzolkiewski, line 20: It should be added here: https://github.com/software-mansion/react-native-enriched/pull/318#issuecomment-3660669405

But it was deleted by force push

ios/inputParser/InputParser.mm
kacperzolkiewski: Yes my idea was wrong. Right now I set cursor position at the end always when we set input value by seValue. 
But when for example props changed I apply cursor position on previous remembered position.

ios/EnrichedTextInputView.mm
```diff
@@ -529,6 +529,8 @@ - (void)updateProps:(Props::Shared const &)props
     stylePropChanged = YES;
   }
 
+  NSRange prevSelectedRange = textView.selectedRange;
```
szydlovsky: Why don't we put it inside of the `if (stylePropChanged` block? I'm afraid such code out of the context could break something in the future. Let's keep it in its context.

ios/inputParser/InputParser.mm
```diff
@@ -679,6 +679,10 @@ - (NSString *_Nullable)initiallyProcessHtml:(NSString *_Nonnull)html {
     fixedHtml = [fixedHtml stringByReplacingOccurrencesOfString:@"<br></p>"
                                                      withString:@"<br>"];
 
+    // remove Object Replacement Character as it has no meaning in HTML
+    fixedHtml = [fixedHtml stringByReplacingOccurrencesOfString:@"\uFFFC"
+                                                     withString:@""];
+
```
szydlovsky: I think we should also remove these characters when producing a html (same as we do with zero width spaces). It might even be not needed here if we fix it there

ios/EnrichedTextInputView.mm
```diff
@@ -676,6 +678,8 @@ - (void)updateProps:(Props::Shared const &)props
   if (isFirstMount && newViewProps.autoFocus) {
     [textView reactFocus];
   }
+
+  textView.selectedRange = prevSelectedRange;
```
IvanIhnatsiuk: So, now we always process the default value and in a case when the current value is:
```
<html>
<p>Some text |<p> <- | shows caret position
</html>
```
We replace the default value with
```
<html>
</html>
```
Will you get out of the range exception?

ios/EnrichedTextInputView.mm
IvanIhnatsiuk: Could you explain why do we need to control selection if props changed?

szydlovsky: Sure @IvanIhnatsiuk, the `stylePropChanged` flag tells us some prop from style config has changed and we want to re-build the whole input in such case. For the user, it should just be seen as the said config changing (e.g., some color  of a list bullet). But this operation completely cleans and refills the input with content, hence we want to bring selection back to where it was before this operation.

ios/EnrichedTextInputView.mm
szydlovsky: If the selection change is moved inside of the `if (stylePropChanged)` block, everything should be fine.

ios/EnrichedTextInputView.mm
IvanIhnatsiuk: Ah, I see it's done in the same way as for Android

ios/EnrichedTextInputView.mm
IvanIhnatsiuk: Okay, but now we are in a situation like:

We changed the default value and one style:

1. We applied a new attrbiutedstring to the textView when we processed the default value
2. Then we applied attributedstring again during the stylePropChanged.

Would it be better to apply new attrbiuted string only once?

ios/EnrichedTextInputView.mm

ios/inputParser/InputParser.mm
kacperzolkiewski: Yes, you are right. Fixed

ios/EnrichedTextInputView.mm
```diff
@@ -539,6 +539,7 @@ - (void)updateProps:(Props::Shared const &)props
     NSString *currentHtml = [parser
         parseToHtmlFromRange:NSMakeRange(0,
                                          textView.textStorage.string.length)];
+    NSRange prevSelectedRange = textView.selectedRange;
```
szydlovsky, line 543: And a comment explaining that we want to preserve the selection, please 😄

ios/EnrichedTextInputView.mm
szydlovsky: It has a potential for an optimization but I think it's not in this PR's scope

ios/EnrichedTextInputView.mm
kacperzolkiewski: done

exploIF (approved):

Issue #332, opened 2025-12-17
Padding seems to be applied as margin


rossfishkind:
When adding padding to the editor, it seems to treat it as margin, separating the content from the edges, and insetting the scroll bar. You can see in the screenshot that the scroll bar is inset to the content, when it should be at the outer bounds, closest to the screen edge. This behavior is expected with margin, however, padding is expected to be internal. The other issue with this is that the outside bounds are not touchable for scrolling - not that it's a popular touch point, but nonetheless.

To double-check this, I replaced the editor with an RN `TextInput`, and it had the expected behavior.

Here is my code using `EnrichedTextInput`:
```tsx
export default function WriteContent() {
	const richEditorRef = useRef<EnrichedTextInputInstance>(null);
	const insets = useSafeAreaInsets();
	const { styles: themeStyles } = useTheme();
	const body = useEditorStore((state) => state.body);
	const { height: keyboardHeight } = useReanimatedKeyboardAnimation();

	const containerStyle = useAnimatedStyle(() => {
		return {
			paddingBottom: Math.abs(keyboardHeight.value),
		};
	}, [insets.bottom]);

	return (
		<Animated.View style={[styles.container, containerStyle]}>
			<EnrichedTextInput
				ref={richEditorRef}
				defaultValue={toEnrichedHtml(body)}
				placeholder="Start your story…"
				placeholderTextColor={themeStyles.text.placeholder}
				style={styles.editor}
				onFocus={() => {
					requestAnimationFrame(() => richEditorRef.current?.focus?.());
				}}
			/>
		</Animated.View>
	);
}

const styles = StyleSheet.create({
	container: { flex: 1 },
	editor: {
		fontSize: 18,
		fontFamily: "Lora",
		lineHeight: 27,
		padding: 24,
	},
});
```

Screenshot:
<img width="320" alt="Image" src="https://github.com/user-attachments/assets/1939f77d-6b87-4d6d-9e44-ea7c4ce1bcbf" />

I am on version `0.2.0`

Thank you team! Amazing library!

exploIF:
Thanks for reporting, do you see this issue on both platforms or is it iOS only?

rossfishkind:
@exploIF sure thing! I have only tested iOS so far, but I can get a simulator up and running for android to check.

larsmunkholm:
This is on Android:

https://github.com/user-attachments/assets/4c643ecf-9680-4bcb-a61b-8b1684db3609

[...]

PR #339, opened 2025-12-18
fix(iOS): proper layout calculation

IvanIhnatsiuk:
# Summary
This PR simplifies and significantly optimizes text measurement + corrects textKit measurements. 

Closes https://github.com/software-mansion/react-native-enriched/issues/332

### Problems:

1. We had a bug that contentInsets were incorrectly applied, and padding was treated as margin.
2. We had asynchronous layout calculations that caused a small delay in the initial mount.
3. We had some workarounds to make content scrollable by setting textContentSize manually.
4. Scroll position is lost if we navigate back to the screen with the editor

All these problems were because we used the `setContentView` method, which is not a UIKit method(it was a RN method) that required proper children caculation. However, in our scenario, we just need to know the current content size, and UIKit can calculate the UIScrollView bounds and text content automatically.

#### Problem Description

The issue was related to our `InputTextView`.

`InputTextView` always had **zero bounds (`{0,0}`)** during layout, which meant that **TextKit did not know the actual size of the text container**. As a result, TextKit attempted to measure `contentSize` using an invalid (zero-width) container, which led to incorrect text measurement and layout behavior.

This behavior can be verified by placing a breakpoint in the method responsible for computing `contentSize` inside `InputTextView`: at that moment, TextKit had no knowledge of the real container size.

---

#### Root Cause

- `updateLayoutMetrics` did not properly apply geometry to `InputTextView`
- The view’s `frame`/`bounds` remained zero
- `textContainer.size` did not reflect the actual width of the component
- TextKit was asked to measure text **before receiving the final container size**

Because of this, `contentSize` was calculated incorrectly, which caused issues such as:
- incorrect height calculation
- broken initial layout
- scrolling not activating properly

---

### Benchmarks 

for the text content like: 
```.ts
const html =
  '<html>' +
  Array(1000)
    .fill(
      `<h1><u>Test</u></h1>
      <ul><li>First item</li></ul>
      <h2>Test h2</h2>
      <p>This is a <b>bold</b> statement with a <a href="https://www.example.com">link</a> and an image below:</p>
      <ol>
        <li>First ordered item</li>
        <li>Second ordered item with <i>italic</i> text</li>
      </ol>
      <blockquote>This is a blockquote.</blockquote>
      <p>Here is some inline code: <code>const x = 10;</code></p>
      `
    )
    .join('') +
  '</html>';
```

| Scenario                      | Before (ms) | After (ms) |
|------------------------------|-------------|------------|
| Measure size on initial mount | 408         | 141        |
| Single character change       | 764         | 0.007      |

## Test Plan

1. Set a huge default value using 
```.ts
const html =
  '<html>' +
  Array(100)
    .fill(`<h1><u>Test</u></h1><ul><li>First item</li></ul><h2>Test h2</h2>`)
    .join('') +
  '</html>';
```
2. Test that scrolling/typing is working well

For example:

- Expected vs actual results

#### Initial layout + scrolling

https://github.com/user-attachments/assets/2af11a21-75a5-4557-9e7f-93b75683afb6

### Long single paragraph

https://github.com/user-attachments/assets/ee18d450-4ffc-4caf-947b-980f9e0fe762

## New line insertion:

Please watch these videos at 0.25x speed.

### Before

https://github.com/user-attachments/assets/d2c10745-204f-4fb8-b608-1b662ad5414a 

### After

https://github.com/user-attachments/assets/0eea53ba-5cc3-4ad0-a6eb-f68f0afa388d 

### Content size layout measurement *(Before)*

Uploading Screen Recording 2025-12-21 at 18.55.16.mov…

### Error that content size is ambiguous

Screenshot 2025-12-21 at 18 56 22

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |


szydlovsky:
Hey @IvanIhnatsiuk the changes look promising but we need to make sure absolutely no regressions are introduced. 
Can you please check the reproductions from these three PR's:

- https://github.com/software-mansion/react-native-enriched/pull/209
- https://github.com/software-mansion/react-native-enriched/pull/224
- https://github.com/software-mansion/react-native-enriched/pull/299

We cannot proceed with the PR no matter how big the performance gains if any of these re-emerge.

If it comes to the measuring part, I'll check it on my end.

IvanIhnatsiuk:
@szydlovsky Sure and it's obviously clear to me 😄 . I will test all of these scenarios and comeback with update 🙂

IvanIhnatsiuk:
@szydlovsky 
1. It is unclear what I should test
2. Works as expected

https://github.com/user-attachments/assets/27cce4f5-d633-4572-b29e-34ba32ca98d5

3. Again, it works:

https://github.com/user-attachments/assets/22095bb0-b631-4d8a-8e2d-1e5707f10910

szydlovsky:
@IvanIhnatsiuk okay, we anyway want to test it out really thoroughly. Might be as late as at the beginning of the new year so we'll need some patience here 🙏

example/ios/Podfile.lock
```diff
@@ -1920,6 +1920,93 @@ PODS:
     - ReactCommon/turbomodule/core
     - SocketRocket
     - Yoga
+  - react-native-safe-area-context (5.6.2):
```
IvanIhnatsiuk: Please, let me know if you want to keep separate screens in the example app or I should remove all of this logic.

NOTE:

Initially, I added it to simplify testing on your side

exploIF: It should be removed. If we considered extending example app, we would like to this separately

IvanIhnatsiuk:
@exploIF kind reminder 🙂

exploIF:
I've noticed one strange behavior (not something crucial but would be good to verify what's happening):
- with example app setup
- specify default value which consists of 5 paragraphs (one liners)
- notice that there is no bouncing when trying to scroll such input, however there is bouncing when we text is 4 or 6 paragraphs 😕 



https://github.com/user-attachments/assets/041fcade-289c-45a8-8468-17e41a28f6b3

IvanIhnatsiuk:
> I've noticed one strange behavior (not something crucial but would be good to verify what's happening):
> 
> * with example app setup
> * specify default value which consists of 5 paragraphs (one liners)
> * notice that there is no bouncing when trying to scroll such input, however there is bouncing when we text is 4 or 6 paragraphs 😕
> 
>  Simulator.Screen.Recording.-.iPhone.17.Pro.-.2026-01-13.at.14.46.46.mov

@exploIF fixed

PR #349, opened 2026-01-06
chore: avoid double text copy in selection change - iOS

IvanIhnatsiuk:
# Summary

Text copying is an expensive operation and may take up to 5ms in large texts. In the selection change, we did this twice. To improve performance, we can copy it only once

## Test Plan

1. Run the example app
2. Type some text
3. Change selection

## Screenshots / Videos

n/a

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |


kacperzolkiewski (approved):
LGTM 🚀

Issue #346, opened 2026-01-05
[iOS] Platform font scaling is not respected


wongk:
**Describe the bug**
When the user has increased their font size within the iOS system settings (Accessibility > Display & Text Size > Larger Text > Increase the font size), this is not respected in the editor input area. This works as expected on Android. See attached screenshots.

**To Reproduce**
1. Increase the system font size
2. Type some text into the editor

**Expected behavior**
The editor font size should be scaled in the same way as other controls.

**Screenshots**

<img width="544" height="1036" alt="Image" src="https://github.com/user-attachments/assets/e995008c-6ff8-4009-9366-f8c0bd444c76" />

Here is Android for comparison:

<img width="532" height="1040" alt="Image" src="https://github.com/user-attachments/assets/4c162304-5710-498a-98ee-f56410314581" />

**Device (please complete the following information):**
 - Device: iPhone 16 Pro sim
 - OS: iOS 18.5
 - Version: 0.20.1

wongk:
The text outside of the editor shows the default size with font scaling applied.

PR #352, opened 2026-01-07
fix: font scaling when changing font size from system settings

kacperzolkiewski:
# Summary
Fixes: #346 

This PR fixes font scaling when changing font size from system settings.

## Test Plan
1. type something in the editor
2. apply styling 
3. go to system settings and change font size
4. come back to app and see that font react on font size

## Screenshots / Videos

https://github.com/user-attachments/assets/d9a77b9f-acbd-471d-a766-b391cd585d6d

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

### DIFF
--- a/ios/styles/HeadingStyleBase.mm
+++ b/ios/styles/HeadingStyleBase.mm
@@ -102,8 +102,8 @@
                            BOOL *_Nonnull stop) {
                 if ([self styleCondition:value:range]) {
                   UIFont *newFont = [(UIFont *)value
-                      setSize:[[[self typedInput]->config
-                                scaledPrimaryFontSize] floatValue]];
+                      setSize:[[[self typedInput]->config scaledPrimaryFontSize]
+                                  floatValue]];
                   if ([self isHeadingBold]) {
                     newFont = [newFont removeBold];
                   }
@@ -123,8 +123,7 @@
     NSMutableDictionary *newTypingAttrs =
         [[self typedInput]->textView.typingAttributes mutableCopy];
     UIFont *newFont = [currentFontAttr
-        setSize:[[[self typedInput]->config scaledPrimaryFontSize]
-                    floatValue]];
+        setSize:[[[self typedInput]->config scaledPrimaryFontSize] floatValue]];
     if ([self isHeadingBold]) {
       newFont = [newFont removeBold];
     }


ios/styles/H3Style.mm
```diff
@@ -9,7 +9,8 @@ + (BOOL)isParagraphStyle {
   return YES;
 }
 - (CGFloat)getHeadingFontSize {
-  return [((EnrichedTextInputView *)input)->config h3FontSize];
+  CGFloat rawSize = [((EnrichedTextInputView *)input)->config h3FontSize];
+  return [[UIFontMetrics defaultMetrics] scaledValueForValue:rawSize];
```
IvanIhnatsiuk: Make sure to add this logic in H4, H5, H6 styles as well 🙂

ios/EnrichedTextInputView.mm
```diff
@@ -1702,6 +1705,48 @@ - (void)textViewDidChange:(UITextView *)textView {
...
+
+    NSString *currentHtml = [parser
+        parseToHtmlFromRange:NSMakeRange(0,
+                                         textView.textStorage.string.length)];
+    NSString *initiallyProcessedHtml =
+        [parser initiallyProcessHtml:currentHtml];
+    [parser replaceWholeFromHtml:initiallyProcessedHtml];
+
+    textView.selectedRange = prevSelectedRange;
+    recentlyChangedRange = NSMakeRange(0, textView.textStorage.string.length);
```
szydlovsky: Unsure if it's necessary; essentially `recentlyChangedRange` only influences `handleWordModificationBasedChanges`, which in turn just handles manual and automatic links extending/removal/detection. Changing font size trait doesn't impact links in no way so I think this `recentlyChangedRange` modification is not needed here, but correct me if I'm wrong.

ios/styles/H1Style.mm
```diff
@@ -9,7 +9,8 @@ + (BOOL)isParagraphStyle {
   return YES;
 }
 - (CGFloat)getHeadingFontSize {
-  return [((EnrichedTextInputView *)input)->config h1FontSize];
+  CGFloat rawSize = [((EnrichedTextInputView *)input)->config h1FontSize];
+  return [[UIFontMetrics defaultMetrics] scaledValueForValue:rawSize];
```
szydlovsky: I think it would be more elegant if we used `[UIFontMetrics defaultMetrics] scaledValueForValue:` already in the end of the config getter for all the headings. This way we keep the logic more centralized.

ios/styles/HeadingStyleBase.mm
```diff
@@ -139,7 +145,15 @@ - (void)removeTypingAttributes {
...
-  return font != nullptr && font.pointSize == [self getHeadingFontSize];
+  if (font == nullptr) {
+    return NO;
+  }
+
+  if (self.lastAppliedFontSize > 0.0) {
+    return font.pointSize == self.lastAppliedFontSize;
+  }
+
+  return font.pointSize == [self getHeadingFontSize];
```
szydlovsky, line 155: Let's add a comment that describes why we need that; when the traits already change, the `getHeadginFontSize` will return the new font size and no headings would be properly detected, so that's why we have to use the latest applied font size rather than that value.

ios/EnrichedTextInputView.mm
```diff
@@ -1702,6 +1705,48 @@ - (void)textViewDidChange:(UITextView *)textView {
...
+ * This method detects the change and performs a "Hard Refresh" of the content.
+ */
+- (void)traitCollectionDidChange:(UITraitCollection *)previousTraitCollection {
+  [super traitCollectionDidChange:previousTraitCollection];
+
+  if (previousTraitCollection.preferredContentSizeCategory !=
+      self.traitCollection.preferredContentSizeCategory) {
+    UIFont *baseFont = [config primaryFont];
+    UIFont *scaledFont =
+        [[UIFontMetrics defaultMetrics] scaledFontForFont:baseFont];
```
szydlovsky: All of the changes from the PR that use the `scaledFontForFont:` method can be avoided by just making sure that the config properly takes that into account. If we now get the `pointSize` from the `scaledFont` and apply it to `[config setPrimaryFontSize]`, then we invalidate all of the needed fonts (primary font, monospaced font, ordered list marker font). This way, all the other usages of the `scaledFontForFont:` can be just not used (except for the usage in specific `H1/2/3/4/5/6Style.mm` classes). After that we can just update the `defaultTypingAttributes` and we're good.

ios/EnrichedTextInputView.mm
kacperzolkiewski: Yes you are right, deleted

ios/config/InputConfig.mm
```diff
@@ -477,4 +477,16 @@ - (void)setCodeBlockBorderRadius:(CGFloat)newValue {
   _codeBlockBorderRadius = newValue;
 }
 
+- (void)invalidateFonts {
+  _primaryFontNeedsRecreation = YES;
+  _monospacedFontNeedsRecreation = YES;
+  _olMarkerFontNeedsRecreation = YES;
+}
+
+- (NSNumber *)getScaledPrimaryFontSize {
```
szydlovsky: Can we please rename it to `scaledPrimaryFontSize`? We're not using `get` word in the config.

kacperzolkiewski: fixed

PR #353, opened 2026-01-07
feat: link detection regex

exploIF:
# Summary

Adds handling of user-defined regular expressions that match and automatically adds links.

### General (JS side)

- Added `linkRegex` prop that takes a `RegExp` value. Not defining this prop (`undefined`) results in a default link regex. Setting a `null` completely disables automatic links.
- The prop is firstly properly normalized and any unsupported regex elements get a proper warning.

### iOS 

- Added the logic behind the user-defined regex handling to the existing automatic links logic.
- Improved both the user-defined as well as the default regular expressions to be memoized and not recreated each time automatic links handling occurs (so on any character change effectively).
- Cleaned up the automatic links code.

### Android

- Added the logic behind the user-defined regex handling to the existing automatic links logic.

## Test Plan

Run the example app and see the currently used link regex:

`const LINK_REGEX =
  /^(?:enriched:\/\/\S+|(?:https?:\/\/)?(?:www\.)?swmansion\.com(?:\/\S*)?)$/i;`
 
 Either play with this one or add your own and see that the links are properly added and removed when they are no longer matching the regex.

### iOS:

https://github.com/user-attachments/assets/fb9614e1-affa-46f4-80e0-b78c78f1f087

### Android:

https://github.com/user-attachments/assets/f3f2f7fa-c4c4-4ba8-adbb-dbe169f94cb4

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

### DIFF
--- a/ios/styles/LinkStyle.mm
+++ b/ios/styles/LinkStyle.mm
@@ -13,9 +13,6 @@ static NSString *const AutomaticLinkAttributeName =
 
 @implementation LinkStyle {
   EnrichedTextInputView *_input;
-  NSRegularExpression *_fullRegex;
-  NSRegularExpression *_wwwRegex;
-  NSRegularExpression *_bareRegex;
 }
 
 + (StyleType)getStyleType {
@@ -26,28 +23,52 @@ static NSString *const AutomaticLinkAttributeName =
   return NO;
 }
 
++ (NSRegularExpression *)fullRegex {
+  static NSRegularExpression *regex;
+  static dispatch_once_t onceToken;
+  dispatch_once(&onceToken, ^{
+    regex =
+        [NSRegularExpression regularExpressionWithPattern:
+                                 @"http(s)?:\\/\\/"
+                                 @"www\\.[-a-zA-Z0-9@:%._\\+~#=]{1,256}\\.[a-"
+                                 @"z]{2,6}\\b([-a-zA-Z0-9@:%_\\+.~#?&//=]*)"
+                                                  options:0
+                                                    error:nullptr];
+  });
+  return regex;
+}
+
++ (NSRegularExpression *)wwwRegex {
+  static NSRegularExpression *regex;
+  static dispatch_once_t onceToken;
+  dispatch_once(&onceToken, ^{
+    regex =
+        [NSRegularExpression regularExpressionWithPattern:
+                                 @"www\\.[-a-zA-Z0-9@:%._\\+~#=]{1,256}\\.[a-"
+                                 @"z]{2,6}\\b([-a-zA-Z0-9@:%_\\+.~#?&//=]*)"
+                                                  options:0
+                                                    error:nullptr];
+  });
+  return regex;
+}
+
++ (NSRegularExpression *)bareRegex {
+  static NSRegularExpression *regex;
+  static dispatch_once_t onceToken;
+  dispatch_once(&onceToken, ^{
+    regex =
+        [NSRegularExpression regularExpressionWithPattern:
+                                 @"[-a-zA-Z0-9@:%._\\+~#=]{1,256}\\.[a-z]{2,"
+                                 @"6}\\b([-a-zA-Z0-9@:%_\\+.~#?&//=]*)"
+                                                  options:0
+                                                    error:nullptr];
+  });
+  return regex;
+}
+
 - (instancetype)initWithInput:(id)input {
   self = [super init];
   _input = (EnrichedTextInputView *)input;
-
-  // init default regexes
-  _fullRegex = [NSRegularExpression
-      regularExpressionWithPattern:@"http(s)?:\\/\\/"
-                                   @"www\\.[-a-zA-Z0-9@:%._\\+~#=]{1,256}\\.[a-"
-                                   @"z]{2,6}\\b([-a-zA-Z0-9@:%_\\+.~#?&//=]*)"
-                           options:0
-                             error:nullptr];
-  _wwwRegex = [NSRegularExpression
-      regularExpressionWithPattern:@"www\\.[-a-zA-Z0-9@:%._\\+~#=]{1,256}\\.[a-"
-                                   @"z]{2,6}\\b([-a-zA-Z0-9@:%_\\+.~#?&//=]*)"
-                           options:0
-                             error:nullptr];
-  _bareRegex = [NSRegularExpression
-      regularExpressionWithPattern:@"[-a-zA-Z0-9@:%._\\+~#=]{1,256}\\.[a-z]{2,"
-                                   @"6}\\b([-a-zA-Z0-9@:%_\\+.~#?&//=]*)"
-                           options:0
-                             error:nullptr];
-
   return self;
 }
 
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/swmansion/enriched/EnrichedTextInputView.kt
```diff
@@ -429,6 +435,35 @@ class EnrichedTextInputView : AppCompatEditText {
     ) or if (flag == InputType.TYPE_NULL) 0 else flag
   }
 
+  fun setLinkRegex(config: ReadableMap?) {
+    val patternStr = config?.getString("pattern")
```
IvanIhnatsiuk, line 439: Should we recheck the text after changing this since non link text may become a link and vice versa?

ios/EnrichedTextInputView.mm
```diff
@@ -708,6 +708,15 @@ - (void)updateProps:(Props::Shared const &)props
     [config setMentionIndicators:newIndicators];
   }
 
+  // linkRegex
+  LinkRegexConfig *oldRegexConfig =
+      [[LinkRegexConfig alloc] initWithLinkRegexProp:oldViewProps.linkRegex];
+  LinkRegexConfig *newRegexConfig =
+      [[LinkRegexConfig alloc] initWithLinkRegexProp:newViewProps.linkRegex];
+  if (![newRegexConfig isEqualToConfig:oldRegexConfig]) {
+    [config setLinkRegexConfig:newRegexConfig];
```
IvanIhnatsiuk, line 717: Should you add stylePropChanged = YES here to check the text for links with the new regular expression?

ios/styles/LinkStyle.mm
```diff
@@ -26,6 +29,25 @@ + (BOOL)isParagraphStyle {
 - (instancetype)initWithInput:(id)input {
   self = [super init];
   _input = (EnrichedTextInputView *)input;
+
+  // init default regexes
+  _fullRegex = [NSRegularExpression
```
IvanIhnatsiuk: Could you combine all of them and put in dispatch_once?

```.c
+ (NSRegularExpression *)fullURLRegex {
  static NSRegularExpression *regex;
  static dispatch_once_t onceToken;
  dispatch_once(&onceToken, ^{
    regex =
        [NSRegularExpression regularExpressionWithPattern:
                                 @"http(s)?://"
                                 @"www\\.[-a-zA-Z0-9@:%._\\+~#=]{1,256}\\.[a-z]"
                                 @"{2,6}\\b([-a-zA-Z0-9@:%_\\+.~#?&//=]*)"
                                                  options:0
                                                    error:nil];
  });
  return regex;
}
```

android/src/main/java/com/swmansion/enriched/styles/ParametrizedStyles.kt
```diff
@@ -145,10 +142,9 @@ class ParametrizedStyles(
     val spannable = view.text as Spannable
     val (word, start, end) = result
 
-    // TODO: Consider using more reliable regex, this one matches almost anything
-    val urlPattern =
-      android.util.Patterns.WEB_URL
-        .matcher(word)
+    val regex = view.linkRegex ?: return
```
kacperzolkiewski: Nit: We can move it even before this line:
`val spannable = view.text as Spannable`
because there is no need to create variables: `spannable` and `(word, start, end)` when `linkRegex` is `null`?

ios/config/InputConfig.mm
```diff
@@ -477,4 +479,41 @@ - (void)setCodeBlockBorderRadius:(CGFloat)newValue {
...
 
+- (LinkRegexConfig *)linkRegexConfig {
+  return _linkRegexConfig;
+}
+
+- (void)setLinkRegexConfig:(LinkRegexConfig *)newValue {
+  _linkRegexConfig = newValue;
+
+  // try initializing the native regular expression if it applies
+
```
kacperzolkiewski:
```suggestion
```

ios/config/InputConfig.mm
```diff
@@ -477,4 +479,41 @@ - (void)setCodeBlockBorderRadius:(CGFloat)newValue {
...
+                               : 0);
+  NSRegularExpression *userRegex =
+      [NSRegularExpression regularExpressionWithPattern:_linkRegexConfig.pattern
+                                                options:options
+                                                  error:&regexInitError];
+
+  if (regexInitError) {
+    RCTLogWarn(@"[EnrichedTextInput]: Couldn't parse the user-defined link "
+               @"regex, falling back to a default regex.");
+    _parsedLinkRegex = nullptr;
```
kacperzolkiewski, line 508: On Android when there is error during regex parsing we are setting regex to default value?
https://github.com/software-mansion/react-native-enriched/pull/353/files#diff-1052dedc3f4bb2fe0123daeab806a12c8706b80b2dcc952fbe5e095c296905d2R463

szydlovsky: @kacperzolkiewski and I'm doing the exact same. Since iOS uses 3 different regexes (see `LinkStyle.mm`), if the regex config suggests user-defined regex but `parsedLinkRegex` prop is `nullptr`, then we fallback to the default ones

kacperzolkiewski: Yeah, right now I see it. when writing the comment I hadn't gotten there yet :D

ios/utils/LinkRegex.h
```diff
@@ -0,0 +1,19 @@
+#pragma once
```
kacperzolkiewski: Do we need this if we have `LinkRegexConfig.h`?

ios/utils/LinkRegexConfig.mm
```diff
@@ -0,0 +1,37 @@
...
+  LinkRegexConfig *copy = [[[self class] allocWithZone:zone] init];
+  copy->_pattern = [_pattern copy];
+  copy->_caseInsensitive = _caseInsensitive;
+  copy->_dotAll = _dotAll;
+  copy->_isDefault = _isDefault;
+  copy->_isDisabled = _isDisabled;
+  return copy;
+}
+
+- (BOOL)isEqualToConfig:(LinkRegexConfig *)otherObj {
```
kacperzolkiewski, line 30: `LinkRegexConfig.h` does not have `isEqualToConfig` function. It has `isEqualToLinkRegex`?

ios/utils/LinkRegex.h
szydlovsky: Nope, my mistake

ios/EnrichedTextInputView.mm
szydlovsky: There's no need. We decided to not check the existing links, because the regex only applies to the automatic detection.

szydlovsky: A link that's already added is out of the scope of the link detection feature.

ios/styles/LinkStyle.mm
szydlovsky: @IvanIhnatsiuk is it really necessary? The regex creation will anyway run only once for a component, because `LinkStyle` gets initialised only once.

IvanIhnatsiuk: @szydlovsky  I remember that in the Discord channel, we have a dev who used a lot of enriched views per screen + it will be a single regex pass through the whole text

android/src/main/java/com/swmansion/enriched/EnrichedTextInputView.kt
exploIF: No we shouldn't. As for now we don't plan to support updating link regex during runtime

PR #356, opened 2026-01-08
chore: lock Ruby and some gems versions

szydlovsky:
# Summary

We've seen some constant lockfile differences (especially `hermes engine` hashes in `Podfile.lock`). 
This PR: 
- locks the ruby version to a fresh, stable `3.3.10`
- locks the cocoapods to `1.16.2`
- changes version requirements for `xcodeproj` gem to `>= 1.27.0` for the newer cocoapods

These together should fix the issues with different pod hashes, provided all the devs use `bundle install && bundle exec pod install` for the Pods ;)

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

### DIFF
--- a/apps/example/Gemfile
+++ b/apps/example/Gemfile
@@ -1,12 +1,12 @@
 source 'https://rubygems.org'
 
 # You may use http://rbenv.org/ or https://rvm.io/ to install and use this version
-ruby "3.4.8"
+ruby "3.3.10"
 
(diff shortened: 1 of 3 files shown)


szydlovsky:
Decided it's not the greatest idea to lock the versions, which will risk people giving up on launching the example app, over some funny hash changing from time to time.

PR #331, opened 2025-12-17
[DO NOT MERGE] feat: set up project for web + native

maksymilianrojek:
> [!WARNING]
> Do not merge - this PR is a collective PR for all web-related changes. It will be kept until the web implementation is complete. 

# Summary

- updated typescript configuration
- updated pre-commit hook to run typescript check with correct config in package and each example
- updated package.json: added "sideEffects" field to support tree-shaking, updated "exports" to support native and web build

## Test Plan

- both example and example-web works correctly
- react-native app with imported `react-native-enriched` works correctly (check freshly setup app)
- web app with imported `react-native-enriched` works correctly (check freshly setup app)

## Screenshots / Videos

Tested on fresh Expo app. The correct entry point is used for web and native: `index.tsx` for web, `index.native.tsx` for native. I checked the bundle using Expo Atlas to make sure the tree-shaking is working as expected - it does.
CleanShot 2025-12-18 at 10 30 51@2x

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| Web |    ✅     |

PR #357, opened 2026-01-09
feat: add types to web implementation, extract common types

maksymilianrojek:
# Summary

Changes:
- added types to web implementation
- extracted common types to /common/types.ts file

## Test Plan

- both example and example-web works correctly (no lint or typecheck issues)

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| Web |    ✅    |

This PR will be merged to a collective [PR](https://github.com/software-mansion/react-native-enriched/pull/331) where all web-related changes will be kept until the implementation is complete.


src/web/EnrichedTextInput.tsx
```diff
@@ -1,3 +1,189 @@
...
+  onChangeSelection?: (
+    e: SyntheticEvent<HTMLElement, OnChangeSelectionEvent>
+  ) => void;
+  /**
+   * Unused for web, but kept for parity with native
+   */
+  androidExperimentalSynchronousEvents?: boolean;
+}
+
+export const EnrichedTextInput = forwardRef(
```
exploIF: I'm fine with keeping it as it is for now, but in the future we can drop support for React < 19. Minimal supported react-native version is 0.79 which uses React 19, so we should be fine with supporting React >= 19 only

src/web/EnrichedTextInput.tsx
```diff
@@ -1,3 +1,189 @@
...
+  (
+    props: EnrichedTextInputProps,
+    ref: ForwardedRef<EnrichedTextInputInstance>
+  ) => {
+    const {
+      autoFocus,
+      editable = true,
```
exploIF: Maybe we should have a common definition of default props and override it on the platform specific component level later if needed?

src/index.native.tsx
```diff
@@ -1,9 +1,12 @@
...
+  OnChangeSelectionEvent,
+} from './native/EnrichedTextInputNativeComponent';
 export type {
   OnChangeTextEvent,
   OnChangeHtmlEvent,
+  OnChangeMentionEvent,
   OnChangeStateEvent,
-  OnLinkDetected,
   OnMentionDetected,
-  OnChangeSelectionEvent,
```
szydlovsky, line 9: Any reason why we essentially omit these two from the web editor? I think we're trying to achieve API parity, even if we use no-ops in some cases (which might be true for `OnChangeSelectionEvent` on web), I think we still want the API interface to be identical.

src/common/types.ts
```diff
@@ -0,0 +1,73 @@
+export interface OnChangeTextEvent {
+  value: string;
+}
+
+export interface OnChangeHtmlEvent {
+  value: string;
+}
+
+export interface OnChangeStateEvent {
```
exploIF: We've just merged a PR which slightly changes this behavior. It would be good to update https://github.com/software-mansion/react-native-enriched/tree/%40maksymilianrojek/set-up-web-input and adjust to new definition soon. It can be handled in separate PR

src/web/EnrichedTextInput.tsx
```diff
@@ -1,3 +1,187 @@
...
+  ul?: {
+    bulletColor?: string;
+    bulletSize?: number;
+    marginLeft?: number;
+    gapWidth?: number;
+  };
+}
+
+export interface EnrichedTextInputProps {
+  ref?: RefObject<EnrichedTextInputInstance | null>;
```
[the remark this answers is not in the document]
szydlovsky: @maksymilianrojek bumping the Copilot comment, do we really need the ref prop here?

src/index.native.tsx
maksymilianrojek: On the web those two are in './web/EnrichedTextInput' file, so they are exported using `export * from './web/EnrichedTextInput';`. I see it now that it might be confusing

src/common/types.ts
maksymilianrojek: I've updated the #331 PR

src/web/EnrichedTextInput.tsx
maksymilianrojek: Good call, I've extracted all defaults into a separate object so it can be shared between web and native implementation.

src/web/EnrichedTextInput.tsx
maksymilianrojek: I've removed the `forwardRef` 👍

src/index.native.tsx
szydlovsky: Yeah, now I see native side is using codegen types in these. I wonder if the users will run into typing issues if they had a shared codebase for both web and native and there would be essentially two possible definitions for these interfaces..

szydlovsky: @maksymilianrojek any resolutions to that one? We can definitely proceed with the PR once this matter is dealt with.

maksymilianrojek: I've added common types to the `types.ts`, added handlers to the native side to "translate" native events to the common event type. https://github.com/software-mansion/react-native-enriched/pull/357/changes/c30730407b9818a9ad6d32652d2e8b771ca28a33

PR #358, opened 2026-01-09
fix: padding applied as margin

### LINKED ISSUE #332: Padding seems to be applied as margin

kacperzolkiewski:
# Summary
Fixes: #332 

1. Implemented `updateLayoutMetrics` to correctly apply `padding` for `_placeholderLabel` constraints.
2. Switched from `boundingRectWithSize` to `[textView sizeThatFits:]` to ensure measurement matches actual rendering logic exactly.
3. Removed `_performRelayout` that caused scroll flickering.
4. Added debounced `ensureCursorIsVisible` to fix cursor visibility when styles (like H1) change the text height dynamically.
5. Fix position of lists markers

## Test Plan
1. Run example code from #332 and check if the padding is not applied as margin.
2. Run example app and play with different styles - check If height of the view is correct

## Screenshots / Videos
Before:

https://github.com/user-attachments/assets/d94bff71-aa6f-4aa3-858b-af991a618cb0

After:

https://github.com/user-attachments/assets/717993c3-98ea-406c-85eb-0dbf330375f9

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

### DIFF
--- a/ios/EnrichedTextInputView.mm
+++ b/ios/EnrichedTextInputView.mm
@@ -38,6 +38,9 @@ using namespace facebook::react;
   UIColor *_placeholderColor;
   BOOL _emitFocusBlur;
   BOOL _emitTextChange;
+  NSLayoutConstraint *_placeholderTopConstraint;
+  NSLayoutConstraint *_placeholderLeadingConstraint;
+  NSLayoutConstraint *_placeholderWidthConstraint;
 }
 
 // MARK: - Component utils
@@ -246,14 +249,16 @@ Class<RCTComponentViewProtocol> EnrichedTextInputViewCls(void) {
   _placeholderLabel = [[UILabel alloc] initWithFrame:CGRectZero];
   _placeholderLabel.translatesAutoresizingMaskIntoConstraints = NO;
   [textView addSubview:_placeholderLabel];
+  _placeholderLeadingConstraint = [_placeholderLabel.leadingAnchor
+      constraintEqualToAnchor:textView.leadingAnchor];
+  _placeholderWidthConstraint = [_placeholderLabel.widthAnchor
+      constraintEqualToAnchor:textView.widthAnchor];
+  _placeholderTopConstraint =
+      [_placeholderLabel.topAnchor constraintEqualToAnchor:textView.topAnchor];
   [NSLayoutConstraint activateConstraints:@[
-    [_placeholderLabel.leadingAnchor
-        constraintEqualToAnchor:textView.leadingAnchor],
-    [_placeholderLabel.widthAnchor
-        constraintEqualToAnchor:textView.widthAnchor],
-    [_placeholderLabel.topAnchor constraintEqualToAnchor:textView.topAnchor],
-    [_placeholderLabel.bottomAnchor
-        constraintEqualToAnchor:textView.bottomAnchor]
+    _placeholderLeadingConstraint,
+    _placeholderWidthConstraint,
+    _placeholderTopConstraint,
   ]];
   _placeholderLabel.lineBreakMode = NSLineBreakByTruncatingTail;
   _placeholderLabel.text = @"";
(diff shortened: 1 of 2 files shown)


ios/EnrichedTextInputView.mm
```diff
@@ -1535,46 +1529,45 @@ - (void)anyTextMayHaveBeenModified {
...
+// Debounced scroll to cursor - coalesces multiple requests into one per runloop
 // tick
-- (void)scheduleRelayoutIfNeeded {
-  // Cancel any previously scheduled invocation to debounce
-  [NSObject cancelPreviousPerformRequestsWithTarget:self
-                                           selector:@selector(_performRelayout)
-                                             object:nil];
-  // Schedule on next runloop cycle
-  [self performSelector:@selector(_performRelayout)
+- (void)ensureCursorIsVisible {
```
IvanIhnatsiuk: It's not the best idea to control the scroll and the cursor manually. Here are bugs that I found

1. Apply some huge padding(50 for the big input and 20 for small messages like) and then text will jump:

https://github.com/user-attachments/assets/6d98b1e0-640b-4869-bf1f-946eb813b6a7

2. If I select the whole text, it will automatically scroll up, this doesn't happen with the default TextView


https://github.com/user-attachments/assets/543fda14-6633-44e3-8af4-aac32d9f2889

3. If I type any word in visible area, it will auto-scroll to the top with padding applied


https://github.com/user-attachments/assets/56b72ae2-c9e1-4cfa-b586-480e1cd4d72f

4. It doesn't allow for selecting the whole text via cursor


https://github.com/user-attachments/assets/365e9c30-9c58-4970-b1e4-e8142dcc1ccc

IvanIhnatsiuk:
In general, this fix is pretty much  the same as what I did in #339 

@kacperzolkiewski , @szydlovsky 

Can we proceed with my PR instead, since this one won't react to the screen rotation and dynamic size changes + it will calculate size when width and height are not actually changed?

A small example:

" - indicates text view bounds
| - caret position

"Some text | "

If I type a text in the caret position, it won't produce a new line, it won't increase the text view width(if we set it as a static value), and there is no sense in **recalculating** width and height for the shadow node.

ios/EnrichedTextInputView.mm
kacperzolkiewski: Yes, you are right. Thank you for testing and bringing it up. I deleted scrolling manually.

kacperzolkiewski:
Hi @IvanIhnatsiuk, 
@exploIF will look at your PR within few days.

Issue #248, opened 2025-11-14
[Feature Request] Checkbox Support


bbereti:
Checkbox support would be really nice.
Thanks for your work

kacperzolkiewski:
Hi @bbereti,
Checkbox support was added in #366 and #363 PRs. 
It will be available in next release.

PR #363, opened 2026-01-12
feat: checkbox list - android

exploIF:
# Summary
Handles: #248 
This PR introduces checkbox list style on Android.

## Test Plan

Ensure that checklist style on Android works fine

## Screenshots / Videos

https://github.com/user-attachments/assets/a3c77384-19a8-4752-ad28-e838ae838037

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

### DIFF
--- a/android/src/main/java/com/swmansion/enriched/textinput/styles/ListStyles.kt
+++ b/android/src/main/java/com/swmansion/enriched/textinput/styles/ListStyles.kt
@@ -58,40 +58,30 @@ class ListStyles(
     name: String,
     start: Int,
     end: Int,
-    isChecked: Boolean?,
+    isChecked: Boolean? = false,
   ) {
     val (safeStart, safeEnd) = spannable.getSafeSpanBoundaries(start, end)
 
-    if (name == EnrichedSpans.UNORDERED_LIST) {
-      val span = EnrichedUnorderedListSpan(view.htmlStyle)
-      spannable.setSpan(span, safeStart, safeEnd, Spanned.SPAN_EXCLUSIVE_EXCLUSIVE)
-      return
+    when (name) {
+      EnrichedSpans.UNORDERED_LIST -> {
+        val span = EnrichedUnorderedListSpan(view.htmlStyle)
+        spannable.setSpan(span, safeStart, safeEnd, Spanned.SPAN_EXCLUSIVE_EXCLUSIVE)
+      }
+
+      EnrichedSpans.ORDERED_LIST -> {
+        val index = getOrderedListIndex(spannable, safeStart)
+        val span = EnrichedOrderedListSpan(index, view.htmlStyle)
+        spannable.setSpan(span, safeStart, safeEnd, Spanned.SPAN_EXCLUSIVE_EXCLUSIVE)
+      }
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/swmansion/enriched/textinput/spans/EnrichedCheckboxListSpan.kt
```diff
@@ -0,0 +1,94 @@
...
+  var isChecked: Boolean,
+  private val htmlStyle: HtmlStyle,
+) : MetricAffectingSpan(),
+  LineHeightSpan,
+  LeadingMarginSpan,
+  EnrichedParagraphSpan {
+  override val dependsOnHtmlStyle: Boolean = true
+
+  private val checkboxDrawable =
+    CheckboxDrawable(htmlStyle.ulCheckboxBoxSize, htmlStyle.ulCheckboxBoxColor, isChecked).apply {
```
IvanIhnatsiuk, line 26: Each time you add a checklist span, it will create a new drawable. What will happen if I have 100 checkboxes on one screen + it's impossible to customize it via providing a custom image?

android/src/main/java/com/swmansion/enriched/textinput/styles/ListStyles.kt
```diff
@@ -68,9 +71,30 @@ class ListStyles(
       val index = getOrderedListIndex(spannable, safeStart)
       val span = EnrichedOrderedListSpan(index, view.htmlStyle)
       spannable.setSpan(span, safeStart, safeEnd, Spanned.SPAN_EXCLUSIVE_EXCLUSIVE)
+      return
+    }
+
+    if (name == EnrichedSpans.CHECKBOX_LIST) {
```
IvanIhnatsiuk: Would it be better to use when(name) here?

android/src/main/java/com/swmansion/enriched/textinput/styles/ListStyles.kt
```diff
@@ -57,6 +58,7 @@ class ListStyles(
     name: String,
     start: Int,
     end: Int,
+    isChecked: Boolean?,
```
kacperzolkiewski: Maybe we can set a default value for that: `isChecked: Boolean? = null`
and then we can remove created extra `setSpan` method that uses this function with `isChecked = null`?

android/src/main/java/com/swmansion/enriched/textinput/spans/EnrichedCheckboxListSpan.kt
exploIF: Yes, that's true, it's impossible to customize by providing custom image. We may consider adding such improvement in the future, as for now it's not planned.

IvanIhnatsiuk: Got it. What about the first question? I just don't understand the point of why we need to create a new drawable for each EnrichedCheckboxListSpan when it's static.

PR #366, opened 2026-01-13
feat(iOS): Add checkbox list

kacperzolkiewski:
# Summary
Fixes: #248 
Add support for checkbox list. 

## Test Plan
Play with checkbox list in example app.

## Screenshots / Videos

https://github.com/user-attachments/assets/6aaebe78-b793-420a-84bd-19a81f1ae851

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

### DIFF
--- a/ios/EnrichedTextInputView.mm
+++ b/ios/EnrichedTextInputView.mm
@@ -264,64 +264,9 @@ Class<RCTComponentViewProtocol> EnrichedTextInputViewCls(void) {
   textView.autoresizingMask =
       UIViewAutoresizingFlexibleWidth | UIViewAutoresizingFlexibleHeight;
   textView.adjustsFontForContentSizeCategory = YES;
-
-  TextBlockTapGestureRecognizer *blockTap =
-      [[TextBlockTapGestureRecognizer alloc]
-          initWithTarget:self
-                  action:@selector(onTextBlockTap:)];
-
-  blockTap.textView = textView;
-  blockTap.input = self;
-  blockTap.cancelsTouchesInView = YES;
-  blockTap.delaysTouchesBegan = YES;
-  blockTap.delaysTouchesEnded = YES;
-
-  for (UIGestureRecognizer *gr in textView.gestureRecognizers) {
-    [gr requireGestureRecognizerToFail:blockTap];
-  }
-  [textView addGestureRecognizer:blockTap];
-}
-
-- (void)onTextBlockTap:(TextBlockTapGestureRecognizer *)gr {
-  if (gr.state != UIGestureRecognizerStateEnded)
-    return;
-  if (![self->textView isFirstResponder]) {
-    [self->textView becomeFirstResponder];
-  }
-
-  switch (gr.tapKind) {
-
-  case TextBlockTapKindCheckbox: {
-    CheckboxListStyle *checkboxStyle =
-        (CheckboxListStyle *)stylesDict[@([CheckboxListStyle getStyleType])];
-
-    if (checkboxStyle) {
-      NSUInteger charIndex = (NSUInteger)gr.characterIndex;
-      [checkboxStyle toggleCheckedAt:charIndex];
-      [self anyTextMayHaveBeenModified];
-
-      NSString *fullText = textView.textStorage.string;
-      NSRange paragraphRange =
-          [fullText paragraphRangeForRange:NSMakeRange(charIndex, 0)];
-      NSUInteger endOfLineIndex = NSMaxRange(paragraphRange);
-
-      // If the paragraph ends with a newline, step back by 1 so the cursor
-      // stays on the current line instead of jumping to the next one.
-      if (endOfLineIndex > 0 && endOfLineIndex <= fullText.length) {
-        unichar lastChar = [fullText characterAtIndex:endOfLineIndex - 1];
-        if ([[NSCharacterSet newlineCharacterSet] characterIsMember:lastChar]) {
-          endOfLineIndex--;
-        }
-      }
-
-      textView.selectedRange = NSMakeRange(endOfLineIndex, 0);
-    }
-    break;
-  }
-
-  default:
-    break;
-  }
+  [textView addGestureRecognizer:[[TextBlockTapGestureRecognizer alloc]
+                                     initWithInput:self
+                                            action:@selector(onTextBlockTap:)]];
 }
 
 // MARK: - Props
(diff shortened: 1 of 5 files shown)


ios/utils/LayoutManagerExtension.mm
```diff
@@ -345,52 +320,151 @@ - (void)drawLists:(EnrichedTextInputView *)typedInput
...
+    CGFloat checkEndY = insetRect.origin.y + insetRect.size.height * 0.35;
+
+    [checkmarkPath moveToPoint:CGPointMake(checkStartX, checkStartY)];
+    [checkmarkPath addLineToPoint:CGPointMake(checkMidX, checkMidY)];
+    [checkmarkPath addLineToPoint:CGPointMake(checkEndX, checkEndY)];
+    [checkmarkPath setLineWidth:strokeWidth * 1.5];
+    [[UIColor whiteColor] setStroke];
+    [checkmarkPath setLineCapStyle:kCGLineCapRound];
+    [checkmarkPath setLineJoinStyle:kCGLineJoinRound];
+    [checkmarkPath stroke];
```
IvanIhnatsiuk: It's not critical, but.

Let's imagine we have 1000 checklist items, and I will scroll through them quickly(using the scroll indicator). In this case layout manager will redraw 1000  items + allocate memory for them. This is unnecessary work that we can just avoid by moving this to config or storing it elsewhere

kacperzolkiewski: Fixed

ios/EnrichedTextInputView.mm
```diff
@@ -251,6 +264,64 @@ - (void)setupTextView {
...
+  blockTap.textView = textView;
+  blockTap.input = self;
+  blockTap.cancelsTouchesInView = YES;
+  blockTap.delaysTouchesBegan = YES;
+  blockTap.delaysTouchesEnded = YES;
+
+  for (UIGestureRecognizer *gr in textView.gestureRecognizers) {
+    [gr requireGestureRecognizerToFail:blockTap];
+  }
+  [textView addGestureRecognizer:blockTap];
```
szydlovsky: All of these could be done in some custom `init` on the `TextBlockTapGestureRecognizer`; one that would take `input` and of course the `action`, assign the input to a property, ran the `initWithTarget` base class method and then did all these operations. I think it would be much more clear.

ios/EnrichedTextInputView.mm
```diff
@@ -251,6 +264,64 @@ - (void)setupTextView {
...
+}
+
+- (void)onTextBlockTap:(TextBlockTapGestureRecognizer *)gr {
```
szydlovsky: Can we move this method somewhere lower; I think it should be in the same place as `traitCollectionDidChange` method - so in the section of delegate methods and likewise.

ios/EnrichedTextInputView.mm
```diff
@@ -251,6 +264,64 @@ - (void)setupTextView {
...
+      // If the paragraph ends with a newline, step back by 1 so the cursor
+      // stays on the current line instead of jumping to the next one.
+      if (endOfLineIndex > 0 && endOfLineIndex <= fullText.length) {
+        unichar lastChar = [fullText characterAtIndex:endOfLineIndex - 1];
+        if ([[NSCharacterSet newlineCharacterSet] characterIsMember:lastChar]) {
+          endOfLineIndex--;
+        }
+      }
+
+      textView.selectedRange = NSMakeRange(endOfLineIndex, 0);
```
szydlovsky: Let's add some comment describing why this seleciton fix is needed and what case was happening without it.

ios/config/InputConfig.mm
```diff
@@ -527,4 +539,111 @@ - (NSNumber *)scaledPrimaryFontSize {
...
+    [boxPath stroke];
+
+    // Draw Checkmark
+    UIBezierPath *checkPath = [UIBezierPath bezierPath];
+    CGFloat startX = insetRect.origin.x + insetRect.size.width * 0.25;
+    CGFloat startY = insetRect.origin.y + insetRect.size.height * 0.5;
+    CGFloat midX = insetRect.origin.x + insetRect.size.width * 0.45;
+    CGFloat midY = insetRect.origin.y + insetRect.size.height * 0.65;
+    CGFloat endX = insetRect.origin.x + insetRect.size.width * 0.75;
+    CGFloat endY = insetRect.origin.y + insetRect.size.height * 0.35;
```
szydlovsky, line 627: Smart, but crazy at the same time to manually draw a checmark 😮

ios/inputParser/InputParser.mm
```diff
@@ -748,7 +825,7 @@ - (NSString *_Nullable)initiallyProcessHtml:(NSString *_Nonnull)html {
                                          inString:fixedHtml
                                           leading:YES
                                          trailing:NO];
-    fixedHtml = [self stringByAddingNewlinesToTag:@"<li>"
+    fixedHtml = [self stringByAddingNewlinesToTag:@"<li"
```
szydlovsky: Okay so I'm not sure about this one. It's there to specifically add newlines after all `<li>`. If we remove the `>`, it will start adding newlines in the middle of `li` tag content. I think we should rather separately do it for:
- `<li>`
- `<li checked>`

You can test it out by inputting a html like this;
`<html><ul data-type=\"checkbox\"><li checked>something</li></ul></html>`
to `setHtml` and I think there will be something wrong with newlines.

ios/styles/CheckboxListStyle.mm
```diff
@@ -0,0 +1,318 @@
...
+         [_input->config checkboxListBoxSize];
+}
+
+- (instancetype)initWithInput:(id)input {
+  self = [super init];
+  _input = (EnrichedTextInputView *)input;
+  return self;
+}
+
+- (void)applyStyle:(NSRange)range {
```
szydlovsky, line 32: Let's add a comment why it's a no-op.

ios/config/InputConfig.mm
kacperzolkiewski: Do you have some other idea to draw it?

ios/inputParser/InputParser.mm
kacperzolkiewski: Actually it worked: 

https://github.com/user-attachments/assets/a8c5a77d-3100-414a-b849-12fb72a9ec7a


But I added separate handling for `<li checked>` as you mention. It will be more secure like that

ios/EnrichedTextInputView.mm
kacperzolkiewski: I created custom `- (instancetype)initWithInput:(id)input action:(SEL)action` where I separated all of this logic

ios/config/InputConfig.mm
szydlovsky: @kacperzolkiewski nope, just interesting 👍

PR #372, opened 2026-01-14
chore: cleanup android hardcoded characters

exploIF:
# Summary

Replaces hardcoded characters with `EnrichedConstants`

### DIFF
--- a/android/src/main/java/com/swmansion/enriched/utils/EnrichedConstants.kt
+++ b/android/src/main/java/com/swmansion/enriched/utils/EnrichedConstants.kt
@@ -3,7 +3,9 @@ package com.swmansion.enriched.utils
 object EnrichedConstants {
   // Zero Width Space
   const val ZWS = '\u200B'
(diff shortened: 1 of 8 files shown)


android/src/main/java/com/swmansion/enriched/utils/EnrichedConstants.kt
```diff
@@ -0,0 +1,9 @@
+package com.swmansion.enriched.utils
+
+object EnrichedConstants {
+  // Zero Width Space
+  const val ZWS = '\u200B'
+
+  // Object Replacement Character
+  const val ORC = '\uFFFC'
```
IvanIhnatsiuk, line 9: Would it be better to declare here ZWS_STRING and ORC_STRING to avoid `toString` + add new line char/string?

```.kt
const val ZWS_STRING = "" + ZWS or ZWS.toString()
const val ORC_STRING = "" + ORC or ORC.toString()
```

PR #378, opened 2026-01-15
chore: refactor android native structure, rename codegen spec

exploIF:
# Summary

- refactor native android file structure - that way it's easier in the future to maintain many different components
- rename codegen spec to `ReactNativeEnrichedSpec` - previous naming was fine for `input` only library, but as we plan to bring `text` component we should use something more generic.

## Test Plan

Ensure that library still works fine and there is no regression. This PR should not affect how library works.

## Screenshots / Videos

n/a

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |


kacperzolkiewski (approved):
LGTM 🚀

Issue #295, opened 2025-11-27
EnrichedTextInputProps support onSubmitEditing and returnKey and blurOnSubmit


duochizhacai:
I hope it can have the same or similar props as TextInput for text submission.

IslamRustamov:
@szydlovsky I can give this a try if deemed needed. 

Though I think **blurOnSubmit** is deprecated and **submitBehavior** should be implemented.

exploIF:
@IslamRustamov this is something that we would definitely like to bring. If you have enough time and motivation it would be super cool. If you decide to work on it and need any help feel free to ping me ;)

As it comes to API, I think we should follow react-native conventions (but let's skip `blurOnSubmit` as it's deprecated). Ideally we should have:
- [returnKeyLabel](https://reactnative.dev/docs/textinput#returnkeylabel-android) - Android only
- [returnKeyType](https://reactnative.dev/docs/textinput#returnkeytype)
- [submitBehavior](https://reactnative.dev/docs/textinput#submitbehavior) - we should follow multiline input rules
- [onSubmitEditing](https://reactnative.dev/docs/textinput#onsubmitediting)

PR #379, opened 2026-01-18
feat: add returnKeyType, submitBehavior, onSubmitEditing, returnKeyLabel props

IslamRustamov:
# Summary
Fixes: #295 
- This PR solves this issue https://github.com/software-mansion/react-native-enriched/issues/295;
- This PR adds new props such as **returnKeyType**, **submitBehavior**, **onSubmitEditing**, **returnKeyLabel** (Android only);
- For the implementation I took most of the code from the original **TextInput** component of react-native;
- This PR impacts the main files with the implementation of text input.

## Test Plan

To verify that the props work set the corresponding props on the text input and see how they work.

Note that **returnKeyLabel** is visible only in landscape mode.

## Screenshots / Videos

Custom **returnKeyLabel**:

Screenshot_1768764380

Custom **returnKeyType**s:

Screenshot 2026-01-18 at 22 31 07
Screenshot 2026-01-18 at 22 31 22
Screenshot 2026-01-18 at 22 31 37

I advise you to test such props as **submitBehavior**, **onSubmitEditing** yourself by pulling this branch and testing them manually.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

### IMPORTANT NOTES

Some of the things were probably implemented poorly so I advice you to not hesitate and jump in with proposals/your own changes in this PR since I might not quickly return to this task.

1. I didn't use built-in type for **returnKeyType** in NativeProps because codegen generates a mess when you pass a `type SomeType = "one" | "two" | "three"` and I can't comprehend how to use. If you have a better idea how to deal with it - please inform me about that;
2. Android multiline problem one - **returnKeyType** basically doesn't work;
3. Android multiline problem two - for some reason listener of "Done" button on keyboard doesn't work on multiline inputs, so I had to write a somewhat hack in TextWatcher to handle the press of "Done" button;
4. Problem 3 is also present on iOS. Because of multiline input I have to check for "\n" input to understand that we pressed "Done". So, if user tries to copy-paste "\n" - then it might blur the input if we set **submitBehavior** to blurAndSubmit (which is an unlikely scenario but this a somewhat hack).

### DIFF
--- a/src/EnrichedTextInput.tsx
+++ b/src/EnrichedTextInput.tsx
@@ -192,6 +192,9 @@ export const EnrichedTextInput = ({
   onChangeSelection,
   onKeyPress,
   onSubmitEditing,
+  returnKeyType,
+  returnKeyLabel,
+  submitBehavior,
   contextMenuItems,
   androidExperimentalSynchronousEvents = false,
   useHtmlNormalizer = false,
@@ -463,6 +466,9 @@ export const EnrichedTextInput = ({
       contextMenuItems={nativeContextMenuItems}
       onContextMenuItemPress={handleContextMenuItemPress}
       onSubmitEditing={onSubmitEditing}
+      returnKeyType={returnKeyType}
+      returnKeyLabel={returnKeyLabel}
+      submitBehavior={submitBehavior}
       androidExperimentalSynchronousEvents={
         androidExperimentalSynchronousEvents
       }


ios/utils/KeyboardUtils.mm
```diff
@@ -0,0 +1,34 @@
...
+
+  id value = uiReturnKeyTypes[returnKeyType];
+
+  if (value) {
+    UIReturnKeyType returnKey = (UIReturnKeyType)[value integerValue];
+
+    return returnKey;
+  }
+
+  return UIReturnKeyDefault;
```
kacperzolkiewski: I think that it would be efficient to use `if else` than creating a dictionary on every  `getUIReturnKeyTypeFromReturnKeyType` usage:
```suggestion
+ (UIReturnKeyType)getUIReturnKeyTypeFromReturnKeyType:(NSString *)returnKeyType {
    if ([returnKeyType isEqualToString:@"done"]) return UIReturnKeyDone;
    if ([returnKeyType isEqualToString:@"go"]) return UIReturnKeyGo;
    if ([returnKeyType isEqualToString:@"next"]) return UIReturnKeyNext;
    if ([returnKeyType isEqualToString:@"search"]) return UIReturnKeySearch;
    if ([returnKeyType isEqualToString:@"send"]) return UIReturnKeySend;
    if ([returnKeyType isEqualToString:@"emergency-call"]) return UIReturnKeyEmergencyCall;
    if ([returnKeyType isEqualToString:@"google"]) return UIReturnKeyGoogle;
    if ([returnKeyType isEqualToString:@"join"]) return UIReturnKeyJoin;
    if ([returnKeyType isEqualToString:@"route"]) return UIReturnKeyRoute;
    if ([returnKeyType isEqualToString:@"yahoo"]) return UIReturnKeyYahoo;
    
    return UIReturnKeyDefault;
}
```

ios/EnrichedTextInputView.mm
```diff
@@ -1213,6 +1226,31 @@ - (void)focus {
...
+                                    withString:@""] copy];
+
+      emitter->onSubmitEditing({
+          .text = [stringToBeEmitted toCppString],
+      });
+    }
+  }
+
+  return shouldSubmit;
+}
```
kacperzolkiewski: It would be cleaner to reactor this function into two separate functions. 
1. `- (BOOL)textInputShouldSubmitOnReturn {...}` - that will return `shouldSubmit`
2. `- (void)emitOnSubmitEdittingEvent {...} `that will contain emitting event

And then emit the event not here inside `textInputShouldSubmitOnReturn` but inside 
```
- (bool)textView:(UITextView *)textView
    shouldChangeTextInRange:(NSRange)range
            replacementText:(NSString *)text
```
 when `shouldSubmit` is `true`

ios/EnrichedTextInputView.mm
```diff
@@ -1738,6 +1776,17 @@ - (void)textViewDidEndEditing:(UITextView *)textView {
 - (bool)textView:(UITextView *)textView
     shouldChangeTextInRange:(NSRange)range
             replacementText:(NSString *)text {
+  if ([text isEqualToString:@"\n"]) {
```
kacperzolkiewski, line 2008:
```suggestion
  // Check if the user pressed "Enter"
  if ([text isEqualToString:@"\n"]) {
```

kacperzolkiewski:
Unfortunately on Android this PR makes that when we press enter in input nothing happens:

https://github.com/user-attachments/assets/b4fc4fc0-8841-4cac-ae61-33593548c84a


Even when I pass prop: `submitBehavior="newline"`

kacperzolkiewski:
About second point:
`Android multiline problem one - returnKeyType basically doesn't work;`

I tested it on `TextInput` from `react-native` and it looks like it does not work also in their implementation.

https://github.com/user-attachments/assets/afc79be2-a4e9-4b45-a667-38455d8644ac

So we can drop support for this prop on Android

kacperzolkiewski:
If I add at the same time `returnKeyType` and `returnKeyLabel` e.g.
```
returnKeyType="next"
returnKeyLabel="Next"
```
 and press on label there is a crash on Android:

https://github.com/user-attachments/assets/15cfc025-1141-4c9a-a02e-2c34b9d85c40

When we will drop support for `returnKeyType` for Android there should not be this crash

IslamRustamov:
@kacperzolkiewski I've made requested changes

\+ regarding the crash on Android when we press on action label. This crash is happening right now on main branch too. It is happening because for some reason this text input thinks that there is some other text input that it should focus on after the action label is pressed and since there is no other text inputs - it crashes with "focus search returned a view that wasn't able to take focus!". I've added commit that sets up ime action label in **prepareComponent** to unspecified action so that the crash would be avoided, but ideally there should be an ability to focus on next text input on ime action label press if there are any. (probably should be done separately, not in this PR)

\+ not sure if it's a big deal or not but on Android when we pass "blurAndSubmit" for example and we hit Enter - we also receive a text changed event, not just submit editing event. Since I can't think of anything else but a hacky way to handle this (creating a boolean variable **ignoreAfterTextChanged** that sets to true in **onTextChanged** and then sets to false in **afterTextChanged**) I'm not sure what would be the best approach here.

kacperzolkiewski:
@IslamRustamov, 
Thank you for your work 🙏
I pushed update for the Android implementation, with the fix for receiving also a text changed event while we passed "blurAndSubmit"  and also removed hacky way of checking if there is `\n` sign inserted in input.
I override `onEditorAction` function and I manage (`shouldSubmit`, `shouldBlur`) there, so we do not need to check newline manually and also it does not fire `onTextChange` because it is not called - system recognizes that it is submit action no newline insertion.

What do you think about that?

IslamRustamov:
I think that's the most correct way to do that. 

I initially tried to use the same thing but due to lack of proper knowledge of Android development I didn't set up onEditorAction correctly and it just didn't fire for me.

src/EnrichedTextInput.tsx
```diff
@@ -185,6 +191,7 @@ export const EnrichedTextInput = ({
   onEndMention,
   onChangeSelection,
   onKeyPress,
+  onSubmitEditing,
```
pkaramon, line 194: Shouldn't we also explicitly destructure  `returnKeyType`, `submitBehaviour`, `returnKeyLabel` so we are consistent with the rest of the props? Currently those 3 props are now passed through `...rest`.

kacperzolkiewski: Yes I think we should. Nice catch!

Issue #212, opened 2025-10-23
[Feature Request] Support keyboard image/GIF insertion (Gboard, iOS keyboard, etc.)


shuo-hiwintech:
**Is your feature request related to a problem? Please describe.**
Currently, `react-native-enriched` does not support inserting images or GIFs directly from the keyboard (like Gboard on Android or the native keyboard on iOS 14+). When users try to insert an image from their keyboard, nothing happens. This is frustrating because:
- Modern keyboards (Gboard, SwiftKey, iOS native keyboard) provide built-in image/GIF pickers
- Users expect this functionality to work in rich text inputs
- It's a standard feature in most modern messaging and social apps
- Copy-pasting images from the clipboard also doesn't work

**Describe the solution you'd like**
I would like `react-native-enriched` to support the Commit Content API (Android) and `UIPasteboard` image handling (iOS) to allow users to:

1. **Insert images from keyboard**: When users select an image/GIF from their keyboard (Gboard, iOS keyboard, etc.), the component should receive the image data
2. **Paste images from clipboard**: When users copy an image and paste it into the input, it should be captured
3. **Provide a callback**: Add an `onImageInserted` or `onRichContentInserted` callback that provides:
   - The image URI/path
   - MIME type (image/png, image/jpeg, image/gif, etc.)
   - Metadata if available

**Example API:**
```typescript
<EnrichedTextInput
  onRichContentInserted={(content) => {
    console.log('URI:', content.uri);
    console.log('Type:', content.mimeType);
    // Handle image upload
  }}
/>
```

**Describe alternatives you've considered**
1. **Custom patch (current workaround)**: I've created a patch that implements this for Android using `InputConnectionCompat.OnCommitContentListener` and iOS using `UIPasteboard`, but it would be better to have official support.

2. **Separate image picker button**: While we can add a button to open the image picker, this doesn't provide the seamless UX that keyboard-based image insertion offers.

3. **Using a different library**: Other libraries don't provide the rich HTML formatting features that `react-native-enriched` offers.

### Technical Implementation Suggestions

**Android (API 25+):**
```kotlin
override fun onCreateInputConnection(outAttrs: EditorInfo): InputConnection? {
    val ic = super.onCreateInputConnection(outAttrs)
    EditorInfoCompat.setContentMimeTypes(outAttrs, arrayOf("image/*"))
    
    return InputConnectionCompat.createWrapper(ic, outAttrs) { inputContentInfo, flags, _ ->
[...]

Issue #164, opened 2025-09-23
Support pasting images


exploIF:
**Is your feature request related to a problem? Please describe.**
It would be super good to have a way to support pasting images.

**Describe the solution you'd like**
Rendering pasted image automatically it's not a good solution. We will have to handle different image sizes, aspect ratios. etc.

Ideally we can provide an event (callback prop) which will be emitted every time user pastes image. Then it's up to the library user (developer) to decide what to do with it. It can be just inserting Image manually or uploading file as an attachment in the chat like apps.

arunabhverma:
Agree with the `onPaste` event approach. When a user pastes an image, the library should emit the event (including image details) and let the developer decide how to handle it.

PR #384, opened 2026-01-20
feat: add support for pasting images

kacperzolkiewski:
# Summary
Fixes: #164 
Fixes: #212

This PR adds:

- support for pasting Images/GIFs into the input with emitting the `OnPasteImages` event.
- support for displaying GIFs

## Test Plan
Run example app and try to paste different images/GIFs into the input.

## Screenshots / Videos

https://github.com/user-attachments/assets/c20ae5de-d10e-4c0d-b3c9-dd540dbcf647

https://github.com/user-attachments/assets/7df6991f-37c5-4de6-acde-1860e54e5b64

https://github.com/user-attachments/assets/17d78680-16ee-4932-a762-cbaee7a6aecd

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

### DIFF
--- a/ios/interfaces/ImageAttachment.mm
+++ b/ios/interfaces/ImageAttachment.mm
@@ -9,6 +9,10 @@
     return nil;
 
   _imageData = data;
+
+  // Assign an empty image to reserve layout space within the text system.
+  // The actual image is not drawn here; it is rendered and overlaid by a
+  // separate ImageView.
   self.image = [UIImage new];
 
   [self loadAsync];
@@ -24,6 +28,11 @@
 
   dispatch_async(dispatch_get_global_queue(QOS_CLASS_USER_INITIATED, 0), ^{
     NSData *bytes = [NSData dataWithContentsOfURL:url];
+
+    // We pass all image data (including static formats like PNG or JPEG)
+    // through the GIF parser. It safely acts as a universal parser, returning
+    // a single-frame UIImage for static formats and an animated UIImage for
+    // GIFs.
     UIImage *img = bytes ? [UIImage animatedImageWithAnimatedGIFData:bytes]
                          : [UIImage systemImageNamed:@"photo"];
 


src/spec/EnrichedTextInputNativeComponent.ts
```diff
@@ -189,6 +189,15 @@ interface TargetedEvent {
   target: Int32;
 }
 
+export interface OnPasteImagesEvent {
+  images: {
+    uri: string;
+    type: string;
+    width: Float;
+    height: Float;
```
exploIF, line 204: Could you create a separate interface that defines pasted image object?

apps/example/src/App.tsx
```diff
@@ -299,6 +301,18 @@ export default function App() {
     setSelection(sel);
   };
 
+  const handlePasteImagesEvent = (e: OnPasteImagesEvent) => {
+    console.log('Pasted images:', e.images);
+
+    for (let image of e.images) {
```
exploIF: Minor, could you replace it with `forEeach`? Let's use more JS syntactic sugar :D

android/src/main/java/com/swmansion/enriched/textinput/utils/RichContentReceiver.kt
```diff
@@ -0,0 +1,153 @@
...
+    Thread {
+      val results = mutableListOf<PastedImage>()
+
+      for (i in 0 until clip.itemCount) {
+        val item = clip.getItemAt(i)
+        val uri = item.uri
+
+        if (uri == null) {
+          continue
+        }
```
exploIF:
```suggestion
        val uri = item.uri ?: continue

```

android/src/main/java/com/swmansion/enriched/textinput/utils/RichContentReceiver.kt
```diff
@@ -0,0 +1,153 @@
...
+      Log.e("RichContentReceiver", "Failed to save URI: $uri", e)
+      null
+    }
+  }
+
+  private fun getMimeTypeFromUri(uri: Uri): String {
+    var mimeType = context.contentResolver.getType(uri)
+
+    if (mimeType == null) {
+      val str = uri.toString().lowercase()
```
exploIF: I think it's better to use https://developer.android.com/reference/android/webkit/MimeTypeMap#getFileExtensionFromUrl(java.lang.String)

android/src/main/java/com/swmansion/enriched/textinput/utils/RichContentReceiver.kt
```diff
@@ -0,0 +1,153 @@
...
+          mimeType.contains("png") -> "png"
+          mimeType.contains("webp") -> "webp"
+          mimeType.contains("heic") -> "heic"
+          mimeType.contains("tiff") -> "tiff"
+          else -> "jpg"
+        }
+
+      // Create Temp File
+      val fileName = "${UUID.randomUUID()}.$ext"
+      val file = File(context.cacheDir, fileName)
```
exploIF: How about https://developer.android.com/reference/java/io/File#createTempFile(java.lang.String,%20java.lang.String)?

android/src/main/java/com/swmansion/enriched/textinput/utils/RichContentReceiver.kt
```diff
@@ -0,0 +1,153 @@
...
+  private fun saveUriToTempFile(
+    context: Context,
+    uri: Uri,
+    mimeType: String,
+  ): PastedImage? {
+    return try {
+      val resolver = context.contentResolver
+
+      // Guess Extension
+      val ext =
```
exploIF: https://developer.android.com/reference/android/webkit/MimeTypeMap#getExtensionFromMimeType(java.lang.String)

android/src/main/java/com/swmansion/enriched/textinput/utils/RichContentReceiver.kt
```diff
@@ -0,0 +1,153 @@
...
+      resolver.openInputStream(uri).use { input ->
+        if (input == null) return null
+        FileOutputStream(file).use { output ->
+          input.copyTo(output)
+        }
+      }
+
+      // Decode Dimensions
+      val options = BitmapFactory.Options()
+      options.inJustDecodeBounds = true
```
exploIF:
```suggestion
      val options = BitmapFactory.Options().apply { inJustDecodeBounds = true}
```

android/src/main/java/com/swmansion/enriched/textinput/events/OnPasteImagesEvent.kt
```diff
@@ -0,0 +1,47 @@
...
+import com.facebook.react.bridge.WritableArray
+import com.facebook.react.bridge.WritableMap
+import com.facebook.react.uimanager.events.Event
+
+data class PastedImage(
+  val uri: String,
+  val type: String,
+  val width: Double,
+  val height: Double,
+)
```
exploIF: Could you put it inside OnPasteImagesEvent companion object?

src/spec/EnrichedTextInputNativeComponent.ts
kacperzolkiewski: as always with codegen I think I cannot:
image

image

android/src/main/java/com/swmansion/enriched/textinput/utils/RichContentReceiver.kt
kacperzolkiewski: Thanks, those are the functions that I was looking for!

src/spec/EnrichedTextInputNativeComponent.ts
exploIF: In such scenario, can we use this type just for the codegen, but export adjusted type?

ios/inputTextView/InputTextView.mm
```diff
@@ -75,6 +75,64 @@ - (void)paste:(id)sender {
   NSArray<NSString *> *pasteboardTypes = pasteboard.pasteboardTypes;
   NSRange currentRange = typedInput->textView.selectedRange;
 
+  NSMutableArray<NSDictionary *> *foundImages = [NSMutableArray new];
```
szydlovsky, line 81: Code for handling image pasting looks fine but let's add at least some comments explaining what's happening; specifically that we first try to find some images and if we do, we use them from the pasteboard instead of looking for html/text further 👍

ios/extensions/ImageExtension.h
```diff
@@ -0,0 +1,35 @@
+#import <UIKit/UIKit.h>
```
szydlovsky, line 1: What's happening in these two files `UIImage+animatedGif` header/implementation is basically an extension to the `UIImage`. Can we rename them to `ImageExtension` and put in the `extensions` folder?

ios/interfaces/UIImage+animatedGIF.h
```diff
@@ -0,0 +1,35 @@
+#import <UIKit/UIKit.h>
+
+/**
+        UIImage (animatedGIF)
+
+    This category adds class methods to `UIImage` to create an animated
+   `UIImage` from an animated GIF.
+*/
+@interface UIImage (animatedGIF)
```
szydlovsky: If we move it to extensions I'd rather we named it this way too:

```suggestion
@interface UIImage (ImageExtension)
```

ios/extensions/ImageExtension.h
```diff
@@ -0,0 +1,35 @@
...
+    To handle this mismatch, I add each source image (from the GIF) to
+   `animation` a varying number of times to match the ratios between the frame
+   durations in the GIF.
+
+    For example, suppose the GIF contains three frames.  Frame 0 has duration 3.
+   Frame 1 has duration 9.  Frame 2 has duration 15.  I divide each duration by
+   the greatest common denominator of all the durations, which is 3, and add
+   each frame the resulting number of times.  Thus `animation` will contain
+   frame 0 3/3 = 1 time, then frame 1 9/3 = 3 times, then frame 2 15/3 = 5
+   times.  I set `animation.duration` to (3+9+15)/100 = 0.27 seconds.
```
szydlovsky, line 30: Unsure about the duration of the animation. Wouldn't it make more sense to have a baseline of 60 frames per second (as it usually is with iPhones, not 100fps) and then operate using that? In that case, we'd have a duration of 3+9+15/60 which is roughly  0.45 seconds. I just don't get where this 100 comes from tbh.

szydlovsky: But at the same time I see it's an implementation from some outside source. I'm okay with leaving it that way but maybe we could ask some LLM what it thinks about the duration and maybe have it change it?

ios/interfaces/ImageAttachment.mm
```diff
@@ -17,17 +18,17 @@ - (instancetype)initWithImageData:(ImageData *)data {
 - (void)loadAsync {
   NSURL *url = [NSURL URLWithString:self.uri];
   if (!url) {
-    self.image = [UIImage systemImageNamed:@"file"];
+    self.storedAnimatedImage = [UIImage systemImageNamed:@"photo"];
```
szydlovsky, line 25: I get it we change the default img if loading failed but why store it as the `storedAnimatedImage`?

ios/interfaces/ImageAttachment.mm
```diff
@@ -17,17 +18,17 @@ - (instancetype)initWithImageData:(ImageData *)data {
...
   dispatch_async(dispatch_get_global_queue(QOS_CLASS_USER_INITIATED, 0), ^{
     NSData *bytes = [NSData dataWithContentsOfURL:url];
-    UIImage *img = bytes ? [UIImage imageWithData:bytes]
-                         : [UIImage systemImageNamed:@"file"];
+    UIImage *img = bytes ? [UIImage animatedImageWithAnimatedGIFData:bytes]
+                         : [UIImage systemImageNamed:@"photo"];
 
     dispatch_async(dispatch_get_main_queue(), ^{
-      self.image = img;
+      self.storedAnimatedImage = img;
```
szydlovsky, line 40: And also here. Do I see it right that `animatedImageWithAnimatedGifData` will just properly handle both animated and non-animated images? And I see we also store the image rather in the `storedAnimatedImage` than the in `image`, why is that so? It's pretty complicated and not really explained. From the current implementation I understand that we just don't use `self.image` in here, right?

kacperzolkiewski: Right now `image` is like a placeholder in text so we keep it empty. And `storedAnimatedImage` will be drawn by ImageViews. If we would store images in `image` property it would draw wrongly two frames -> one from `Image` in text and the other from `storedAnimatedImage` in IMageViews

ios/interfaces/ImageAttachment.mm
kacperzolkiewski: I assume this is to ensure it displays correctly in `UIImageView`. I did run into some issues with `Image` before, but I can’t quite remember what the exact problem was.

ios/extensions/ImageExtension.h
kacperzolkiewski: Answer from Gemini:

> The logic described in your comment is the standard, correct way to adapt a variable-duration GIF for UIImage. Because UIImage forces all frames to have the same duration, normalizing them using the Greatest Common Divisor (GCD) is the only way to preserve the relative timing accuracy.

src/spec/EnrichedTextInputNativeComponent.ts
kacperzolkiewski: fixed

ios/extensions/ImageExtension.h
kacperzolkiewski: of course

ios/inputTextView/InputTextView.mm
kacperzolkiewski: I add some comment

PR #389, opened 2026-01-26
fix: focus events types

exploIF:
# Summary

As per [discussion](https://github.com/software-mansion/react-native-enriched/discussions/387) this fixes incorrect types for `onBlur` and `onFocus` 

## Test Plan

Ensure types for `onFocus` and `onBlur` are defined correctly.

## Screenshots / Videos

n/a

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

### DIFF
--- a/src/EnrichedTextInput.tsx
+++ b/src/EnrichedTextInput.tsx
@@ -30,12 +30,16 @@ import type {
   MeasureOnSuccessCallback,
   NativeMethods,
   NativeSyntheticEvent,
+  TargetedEvent,
   TextStyle,
   ViewProps,
   ViewStyle,
 } from 'react-native';
 import { normalizeHtmlStyle } from './utils/normalizeHtmlStyle';
 import { toNativeRegexConfig } from './utils/regexParser';
+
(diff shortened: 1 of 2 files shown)


src/EnrichedTextInput.tsx
```diff
@@ -144,8 +148,8 @@ export interface EnrichedTextInputProps extends Omit<ViewProps, 'children'> {
   style?: ViewStyle | TextStyle;
   scrollEnabled?: boolean;
   linkRegex?: RegExp | null;
-  onFocus?: () => void;
-  onBlur?: () => void;
+  onFocus?: (e: FocusEvent) => void;
```
kacperzolkiewski, line 151: Don't we want to keep the convention where events are defined here directly:
`e: NativeSyntheticEvent<TargetedEvent>`

exploIF: I'm not sure. In this specific scenario I wanted to have consistent implementation with how RN types blur/focus events, and they do expose `FocusEvent` and `BlurEvent`

Issue #374, opened 2026-01-14
[Feature request] EnrichedText


exploIF:
**Is your feature request related to a problem? Please describe.**
Create an `EnrichedText` component that will be used for rendering `EnrichedTextInput` output.

moraj-gloitel:
We need EnrichedText component.

PR #400, opened 2026-01-27
feat(android): EnrichedText

exploIF:
# Summary

Implements `EnrichedText` component which can be used for displaying text created by `EnrichedTextInput`.
Tackles: https://github.com/software-mansion/react-native-enriched/issues/374

## Test Plan

Verify that `EnrichedText` component is rendered properly, styles can be customized and any valid html or raw text is displayed properly

## Screenshots / Videos

https://github.com/user-attachments/assets/280bfc98-a2f4-4762-aafa-d5b00e7827c8

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |


android/src/main/java/com/swmansion/enriched/text/EnrichedTextView.kt
```diff
@@ -0,0 +1,207 @@
...
+    if (family != fontFamily) {
+      fontFamily = family
+      typefaceDirty = true
+    }
+  }
+
+  fun setFontWeight(weight: String?) {
+    val fontWeight = parseFontWeight(weight)
+
+    if (fontWeight != fontStyle) {
```
kacperzolkiewski:
```suggestion
    if (fontWeight != this.fontWeight) {
```

android/src/main/java/com/swmansion/enriched/text/EnrichedTextView.kt
```diff
@@ -0,0 +1,207 @@
...
+
+    movementMethod = EnrichedTextMovementMethod.getInstance()
+    setPadding(0, 0, 0, 0)
+    setFontSize(EnrichedConstants.TEXT_DEFAULT_FONT_SIZE)
+  }
+
+  private fun updateValue() {
+    val text = value ?: return
+    val style = enrichedStyle ?: return
+    if (!valueDirty) return
```
kacperzolkiewski, line 137:
```suggestion
	if (!valueDirty) return
    val text = value ?: return
    val style = enrichedStyle ?: return
```

kacperzolkiewski: Nit: We do not need to check for `text` and `style` if `valueDirty == false`

android/src/main/java/com/swmansion/enriched/text/EnrichedTextViewManager.kt
```diff
@@ -0,0 +1,143 @@
...
+  override fun setPadding(
+    view: EnrichedTextView?,
+    left: Int,
+    top: Int,
+    right: Int,
+    bottom: Int,
+  ) {
+    super.setPadding(view, left, top, right, bottom)
+
+    view?.setPadding(left, top, right, bottom)
```
kacperzolkiewski, line 86: Do we need to call `setPadding` twice with `super `and `view.` here?

android/src/main/java/com/swmansion/enriched/text/MeasurementStore.kt
```diff
@@ -0,0 +1,166 @@
...
+
+    val isHtml = text.startsWith("<html>") && text.endsWith("</html>")
+    if (!isHtml) return text
+
+    try {
+      val style = props?.getMap("htmlStyle") ?: return text
+      val enrichedStyle = EnrichedTextStyle.fromReadableMap(context as ReactContext, fontSize, style)
+      val factory = EnrichedTextSpanFactory()
+      val parsed = EnrichedParser.fromHtml(text, enrichedStyle, factory)
+      return parsed.trimEnd('\n')
```
kacperzolkiewski, line 114: I think that we should put deletion of last newline:`.trimEnd('\n')` into `fromHtml` function because everywhere where we are using `fromHtml` we also delete last newline after parsing.

src/EnrichedText.tsx
```diff
@@ -0,0 +1,131 @@
...
+    },
+    setNativeProps: (nativeProps: object) => {
+      nullthrows(nativeRef.current).setNativeProps(nativeProps);
+    },
+    focus: () => {
+      nullthrows(nativeRef.current).focus();
+    },
+    blur: () => {
+      nullthrows(nativeRef.current).blur();
+    },
```
kacperzolkiewski, line 115: Should't we omit those methods? `EnrichedText` is only renderer so those methods have no effect on it?

android/src/main/java/com/swmansion/enriched/text/EnrichedTextView.kt
```diff
@@ -0,0 +1,207 @@
...
+      return
+    }
+
+    try {
+      val parsed = EnrichedParser.fromHtml(text, style, spannableFactory)
+      val withoutLastNewLine = parsed.trimEnd('\n')
+      setText(withoutLastNewLine, BufferType.SPANNABLE)
+      observeAsyncImages()
+    } catch (e: Exception) {
+      this.text = text
```
kacperzolkiewski: Maybe if there will be some problem during parsing we would like to use `GumboParser` like inside `EnrichedTextInput`? It would cover more cases of different html.

.maestro/scripts/run-tests.sh
pkaramon, line 1: I think that in CONTRIBUTING.md section on E2E testing might need a slight update as well.

.maestro/enrichedText/subflows/capture_or_assert_screenshot.yaml
pkaramon, line 1: Nitpick: We could extract the common denominator from enrichedText and enrichedInput `capture_or_assert_screenshot.yaml` and put it in `.maestro/subflows`, then subflows for each component would call that general subflow and pass as argument `elementId`(to specify on which component we should crop) and they would also add a prefix to the screenshot path.

pkaramon:
Not sure how hard this would be, but it’d be great if styles were preserved when copying from EnrichedText into EnrichedTextInput.

https://github.com/user-attachments/assets/17ac2c6c-422e-4aa0-8ce2-d91fadc41162

.maestro/enrichedText/flows/mention_press.yaml
```diff
@@ -0,0 +1,23 @@
+appId: swmansion.enriched.example
+---
+# Validates that link press events are triggered correctly
```
pkaramon:
```suggestion
# Validates that mention press events are triggered correctly
```

.maestro/enrichedText/flows/custom_styles_display.yaml
```diff
@@ -0,0 +1,25 @@
+appId: swmansion.enriched.example
+---
+# Validates that custom styles are displayed correctly
+- launchApp
+
```
pkaramon, line 5: Maybe consider adding a `android-only` tags to those tests, as those tests currently do not pass on iOS

src/utils/normalizeHtmlStyle.ts
```diff
@@ -1,8 +1,28 @@
...
+    ...DEFAULT_HTML_STYLE.a,
+    pressColor: 'darkblue',
+  },
+  mention: {
+    ...DEFAULT_HTML_STYLE.mention,
+    pressColor: 'darkblue',
+    pressBackgroundColor: 'yellow',
+  },
+};
+
```
pkaramon: I think we should consider moving this into `./defaultHtmlStyle`, as web will need it as well. I would also make the name all caps, so it's consistent with `DEFAULT_HTML_STYLE`

src/index.tsx
```diff
@@ -17,4 +18,9 @@ export type {
   EnrichedTextInputInstance,
   ContextMenuItem,
   OnChangeMentionEvent,
+  EnrichedTextHtmlStyle,
 } from './types';
+export type {
+  OnMentionPressEvent,
+  OnLinkPressEvent,
+} from './spec/EnrichedTextNativeComponent';
```
pkaramon: I would recreate those types in `types.tsx` as we do for the rest of the event types. The reason is that web will be able to then use those types as well.

kacperzolkiewski:
> Not sure how hard this would be, but it’d be great if styles were preserved when copying from EnrichedText into EnrichedTextInput.
> 
>  Screen.Recording.2026-04-16.at.11.58.59.mov

@pkaramon nice idea, but it will be better to add it in separate PR ;)

android/src/main/java/com/swmansion/enriched/text/EnrichedTextView.kt
kacperzolkiewski: Ok we will add support for `GumboParser` also in separate PR

pkaramon:
When `selectable` is set to true clicking on mentions and links does not work.

https://github.com/user-attachments/assets/a33d2702-4764-46be-8dc5-a151e47c6828

I've asked AI about this and apparently: 

`TextView` only uses **one** `MovementMethod` at a time.
- **Links/mentions** rely on a link-style method (here: `EnrichedTextMovementMethod`, based on `LinkMovementMethod`).
- **`selectable={true}`** maps to `setTextIsSelectable(true)`, which installs the **selection** movement method (`ArrowKeyMovementMethod`), **replacing** the link handler.

.maestro/scripts/run-tests.sh
kacperzolkiewski: Yes, updated

kacperzolkiewski:
> When `selectable` is set to true clicking on mentions and links does not work.
> 
>  Screen.Recording.2026-04-16.at.14.41.08.mov 
> 
> I've asked AI about this and apparently:
> 
> `TextView` only uses **one** `MovementMethod` at a time.
> 
> * **Links/mentions** rely on a link-style method (here: `EnrichedTextMovementMethod`, based on `LinkMovementMethod`).
> * **`selectable={true}`** maps to `setTextIsSelectable(true)`, which installs the **selection** movement method (`ArrowKeyMovementMethod`), **replacing** the link handler.

@pkaramon can you test it right now? I pushed the fix for that

pkaramon:
I've found a tiny bug/inconsistency. I think we should display an image placeholder in `EnrichedText` the same as we do in `EnrichedTextInput`:

https://github.com/user-attachments/assets/1e29638c-1a3a-490d-8b95-9e4137cab54c

kacperzolkiewski:
> I've found a tiny bug/inconsistency. I think we should display an image placeholder in `EnrichedText` the same as we do in `EnrichedTextInput`:
> 
>  Screen.Recording.2026-04-17.at.08.35.23.mov

@pkaramon as we discussed on Slack, it’s not visible because it defaults to white on Android.
My idea is to make the image placeholder use the text color instead. I’d prefer to handle this in a separate PR.

src/EnrichedText.tsx
exploIF: We extend `NativeMethods` which has `focus` and `blur` defined as required members.

That's the same for RN TextInput - https://github.com/facebook/react-native/blob/77d3df8cab3dccfd4fefad6a80603fadcfc45106/packages/react-native/Libraries/Text/Text.d.ts#L227

android/src/main/java/com/swmansion/enriched/text/EnrichedTextViewManager.kt
exploIF: We do exactly the same for input. If I remember correctly, this is required in order to properly measure component size

android/src/main/java/com/swmansion/enriched/text/MeasurementStore.kt
exploIF: Let's handle that separately, as this will affect input component as well

src/EnrichedText.tsx
kacperzolkiewski: Yes, but we can omit those methods when extending `NativeMethods`?

exploIF: What's the benefit here? Are you sure this won't break any RN internals? It looks like RN expects that each native component defines methods provided by HostInstance

exploIF:
Looks like `ellipsizeMode` is not working properly. I think it's a regression as it worked fine at some point

exploIF (commented):
Recent changes looks good to me 🔥

kacperzolkiewski (approved):
LGTM🚀

PR #403, opened 2026-01-29
feat(ios): add textAlignment

kacperzolkiewski:
# Summary

This PR:
 1. Adds textAlignment support on iOS
 2. Applies a minimal NSKernAttributeName to zero-width spaces (\u200B).
-  ZWS has zero advance width, NSLayoutManager produces a zero-width usedRect for ZWS-only lines, which under right/center alignment collapses the line height. The minimal kerning gives the glyph a non-zero advance, preserving line geometry while remaining invisible.

## Test Plan
Run example app and play with `textAlignment` when different styles are toggled.
Check if paragraphs/lists with different `textAlignment` than `left` are properly parsed and contain `style="text-align:center|right|justify"` inside HTML

## Screenshots / Videos

https://github.com/user-attachments/assets/cf9be29d-8808-45a9-b5d3-53b6d28fab86

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |


MartycFly321:
@kacperzolkiewski  Awesome feature! Would love to see it implemented.
I remember from the foreground color PR that it only gets approved if it works for IOS AND Android:

https://github.com/software-mansion/react-native-enriched/pull/266#issuecomment-3569936267 
_"Hey @IvanIhnatsiuk, it's super nice to get such verbose contribution from the community! While the feature is potentially interesting for us we cannot really proceed with it for now; we want the API and new features to be working on both mobile platforms from the get-go. Either way, great work!"_

ios/inputParser/InputParser.mm
```diff
@@ -1442,4 +1470,110 @@ - (NSDictionary *)prepareCheckboxListStyleValue:(NSValue *)rangeValue
...
+  case NSTextAlignmentRight:
+    return @"right";
+  case NSTextAlignmentJustified:
+    return @"justify";
+  default:
+    return nil;
+  }
+}
+
+- (NSString *)prepareStyleAttrStr:(NSInteger)location
```
szydlovsky: I'd rename it to something like `prepareCssStyleString`, `styleAttr` might mean anything in this context

ios/inputParser/InputParser.mm
```diff
@@ -219,19 +221,24 @@ - (NSString *)parseToHtmlFromRange:(NSRange)range {
           [result appendString:@"\n</ul>"];
         }
 
+        NSString *styleAttr = [self prepareStyleAttrStr:currentRange.location
```
szydlovsky: If we're gonna rename the method (see other comment near its declaration), for the sake of clarity I'd rename the variable to `cssStyleString` or something along these lines

ios/inputParser/InputParser.mm
```diff
@@ -486,6 +495,9 @@ - (NSString *)parseToHtmlFromRange:(NSRange)range {
 - (NSString *)tagContentForStyle:(NSNumber *)style
                       openingTag:(BOOL)openingTag
                         location:(NSInteger)location {
+  NSString *styleAttr = [self prepareStyleAttrStr:location
```
szydlovsky: Same here, we might use some more informative name.

ios/inputParser/InputParser.mm
```diff
@@ -595,8 +607,7 @@ - (NSString *)tagContentForStyle:(NSNumber *)style
     }
   } else if ([style isEqualToNumber:@([BlockQuoteStyle getStyleType])] ||
              [style isEqualToNumber:@([CodeBlockStyle getStyleType])]) {
-    // blockquotes and codeblock use <p> tags the same way lists use <li>
```
szydlovsky: Let's keep that comment here, it explains our format for anyone interested.

ios/utils/AlignmentUtils.mm
```diff
@@ -0,0 +1,175 @@
...
+
+  if ([alignStr isEqualToString:@"left"]) {
+    alignment = NSTextAlignmentLeft;
+  } else if ([alignStr isEqualToString:@"center"]) {
+    alignment = NSTextAlignmentCenter;
+  } else if ([alignStr isEqualToString:@"right"]) {
+    alignment = NSTextAlignmentRight;
+  } else if ([alignStr isEqualToString:@"justify"]) {
+    alignment = NSTextAlignmentJustified;
+  }
```
szydlovsky: This looks exactly the same as what is done in `InputParams` `alignmentFromStyleParams` method. Maybe let's add a helper that extracts that in this utils file instead of copying the logic?

ios/utils/AlignmentUtils.mm
```diff
@@ -0,0 +1,175 @@
...
+  [textView.textStorage endEditing];
+
+  // Update Typing Attributes
+  NSMutableDictionary *typingAttrs = [textView.typingAttributes mutableCopy];
+  NSMutableParagraphStyle *typingStyle =
+      [typingAttrs[NSParagraphStyleAttributeName] mutableCopy];
+  typingStyle.alignment = alignment;
+  typingAttrs[NSParagraphStyleAttributeName] = typingStyle;
+  textView.typingAttributes = typingAttrs;
```
szydlovsky: We have to make the typingAttributes updating optional; otherwise, when setting html via parser, we might have e.g. a line somewhere in the middle that has a different alignment, but we'll start typing with it at the end of the input. The parser already adds styles without their typing attributes for this exact reason so let's fix it with alignment as well.

szydlovsky (commented):
Overall solid, left a few comments. 
I wonder if we can just make the alignment just another style? It would simplify a few things because this implementation kinda looks all over the place. Either way I'm open for discussion 😄

ios/utils/AlignmentUtils.mm
kacperzolkiewski: Nice catch! done

kacperzolkiewski:
Currently, the cursor is not properly aligned on empty lines when active formatting (such as alignment or headings) is applied. Fixing this requires a refactor of the attribute application logic, so it will be addressed in a separate PR.

https://github.com/user-attachments/assets/f0dcb730-62a8-4474-9b97-a3a3ec4f848b

.maestro/enrichedInput/flows/paragraph_styles_alignment_visual.yaml
```diff
@@ -0,0 +1,46 @@
...
+- inputText: 'Left aligned'
+- pressKey: Enter
+
+# Centre-aligned
+- tapOn:
+    id: 'toolbar-align-center'
+- inputText: 'Centre aligned'
+- pressKey: Enter
+
+# Heading 4
```
szydlovsky:
```suggestion
# Heading 6
```

apps/example/src/screens/DevScreen.tsx
szydlovsky, line 1: Changes to this file are unused.

ios/enrichedInputTextView/EnrichedInputTextView/EnrichedInputTextView.mm
```diff
@@ -18,6 +20,48 @@ - (void)layoutSubviews {
...
+
+  // Determine whether the paragraph contains no visible characters.
+  BOOL isEmpty = NO;
+  if (paraRange.length == 0) {
+    isEmpty = YES;
+  }
+
+  if (!isEmpty) {
+    return rect;
+  }
```
szydlovsky: We can simplify it quite a bit I think:
```suggestion
  // Non-empty paragraph gets its caret drawn the usual way.
  if (paraRange.length != 0) {
    return rect;
  }
```

ios/utils/TextListsUtils.h
szydlovsky, line 1: quite a nit but I think it should be `TextListsUtils` (plural form), as we're anyway modifying `.textLists` prop

ios/utils/TextListUtils.h
```diff
@@ -0,0 +1,40 @@
...
+// Appends value to the array. If exclusivePrefix is non-nil, any existing
+// entry whose markerFormat starts with that prefix is evicted first, ensuring
+// only one value from the family is present at a time.
++ (NSArray<NSTextList *> *_Nonnull)
+      textListsByAdding:(NSString *_Nonnull)value
+    withExclusivePrefix:(NSString *_Nullable)prefix
+                toArray:(NSArray<NSTextList *> *_Nullable)existing;
+
+// Returns a new array with every entry whose markerFormat equals value removed
+// or whose markerFormat starts with prefix
```
szydlovsky: this description kinda doesn't make sense, shouldn't it be:

```suggestion
// Returns a new array without entries whose markerFormat equals value removed
// or whose markerFormat starts with prefix
```

ios/utils/TextListUtils.h
```diff
@@ -0,0 +1,40 @@
...
++ (NSArray<NSTextList *> *_Nonnull)
+    textListsByRemoving:(NSString *_Nonnull)value
+             withPrefix:(NSString *_Nullable)prefix
+              fromArray:(NSArray<NSTextList *> *_Nullable)existing;
+
+// Returns a new array with every entry whose markerFormat starts with prefix
+// removed.
```
szydlovsky: Same here, should be:

```suggestion
// Returns a new array without entries whose markerFormat starts with prefix
```

ios/utils/TextListUtils.h
```diff
@@ -0,0 +1,40 @@
...
+
+// Returns YES if any entry's markerFormat equals value exactly.
++ (BOOL)textLists:(NSArray<NSTextList *> *_Nullable)textLists
+    containsValue:(NSString *_Nonnull)value;
+
+// Returns YES if any entry's markerFormat starts with prefix.
++ (BOOL)textLists:(NSArray<NSTextList *> *_Nullable)textLists
+    containsPrefix:(NSString *_Nullable)prefix;
+
+// Returns the first entry whose markerFormat starts with prefix, or nil.
```
szydlovsky:
```suggestion
// Returns the first entry with a markerFormat that starts with prefix, otherwise nil.
```

ios/utils/ParagraphAttributesUtils.mm
```diff
@@ -234,19 +241,41 @@ + (BOOL)handleResetTypingAttributesOnBackspace:(NSRange)range
...
-    typedInput->textView.typingAttributes = typedInput->defaultTypingAttributes;
+    [self resetTypingAttributes:typedInput preserveAlignment:savedAlignment];
     return YES;
   }
 
   return NO;
 }
 
++ (void)resetTypingAttributes:(EnrichedTextInputView *)input
+            preserveAlignment:(NSTextAlignment)alignment {
```
szydlovsky: Nit, maybe a rename:

```suggestion
+ (void)resetTypingAttributes:(EnrichedTextInputView *)input
            preservingAlignment:(NSTextAlignment)alignment {
```

ios/inputAttributesManager/InputAttributesManager.mm
```diff
@@ -136,7 +143,29 @@ - (void)manageTypingAttributesWithOnlySelection:(BOOL)onlySelectionChanged {
...
+          defaultPStyle ? [defaultPStyle mutableCopy]
+                        : [[NSMutableParagraphStyle alloc] init];
+
+      // preserve alignment
+      paraStyle.alignment = savedAlignment;
+      NSString *markerFormat =
+          [AlignmentUtils alignmentToMarker:savedAlignment];
+      paraStyle.textLists =
+          @[ [[NSTextList alloc] initWithMarkerFormat:markerFormat options:0] ];
+
```
szydlovsky: We have the method in `ParagraphAttributesUtils`: `resetTypingAttributes` that preserves alignment. Let's use it here to not copy the logic too much.

ios/htmlParser/HtmlParser.mm
```diff
@@ -1348,9 +1379,96 @@ + (NSString *)tagContentForStyle:(NSNumber *)style
...
+  case NSTextAlignmentCenter:
+    return @"center";
+  case NSTextAlignmentRight:
+    return @"right";
+  case NSTextAlignmentJustified:
+    return @"justify";
+  default:
+    return nil;
+  }
+}
```
szydlovsky: We could also move it to `AlignmentUtils` but I leave the choice to you.

ios/htmlParser/HtmlParser.mm
```diff
@@ -1348,9 +1379,96 @@ + (NSString *)tagContentForStyle:(NSNumber *)style
...
+  if (match) {
+    // rangeAtIndex:1 corresponds to the capture group
+    // (left|center|right|justify)
+    NSString *value =
+        [[params substringWithRange:[match rangeAtIndex:1]] lowercaseString];
+    return [AlignmentUtils stringToAlignment:value];
+  }
+
+  return NSTextAlignmentNatural;
+}
```
szydlovsky: Same here, it could land in `AlignmentUtils` but if you feel like we want to keep it here, then I'm also okay with that.

ios/htmlParser/HtmlParser.mm
szydlovsky, line 1: I found one regression, not critical but nice to fix:
If we have a specially-aligned (other than default) paragraph, and then regular paragraph, their paragraph tags in emitted html don't have a newline between them:

Image

Image

Probably some change in parser has taken out some newline.

kacperzolkiewski: fixed

szydlovsky (approved):
Looks solid 💣

PR #406, opened 2026-01-30
fix: remove unnecessary native onChangeStateDeprecated code

kacperzolkiewski:
# Summary

Remove native `onChangeStateDeprecated` code and handle this event on JS side.

## Test Plan

Run example app and toggle some style - check if `onChangeStateDeprecated` and `onChangeState` callbacks are called properly.

## Screenshots / Videos

n/a

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

### DIFF
--- a/ios/inputParser/InputParser.mm
+++ b/ios/inputParser/InputParser.mm
@@ -1315,7 +1315,7 @@
       [styleArr addObject:@([MentionStyle getStyleType])];
       // extract html expression into dict using some regex
       NSMutableDictionary *paramsDict = [[NSMutableDictionary alloc] init];
-      NSString *pattern = @"(\\w+)=\"([^\"]*)\"";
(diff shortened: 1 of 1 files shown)


src/EnrichedTextInput.tsx
```diff
@@ -356,6 +356,36 @@ export const EnrichedTextInput = ({
     pendingHtmlRequests.current.delete(requestId);
   };
 
+  const onChangeStateWithDeprecated = (
```
exploIF, line 359: Minor one, let's call `onChangeState` first.

Also, let's add a comment above `onChangeStateDeprecated` to remove it in enriched 0.5.0

exploIF (approved):
🚀

src/EnrichedTextInput.tsx
kacperzolkiewski: Fixed

PR #409, opened 2026-02-01
feat: add insertValue method

### LINKED ISSUE #255: [Feature Request] Add insertText(text) API to insert plain text at cursor
EnrichedTextInput currently provides formatting commands (bold, image, mention, link, etc.), but there is no API to insert plain text at the current cursor position.

A simple method like:

```insertText(text: string): void```

would allow developers to implement common features such as inserting emoji (😄), autocomplete suggestions, custom keyboard input, and snippet insertion.
Using setValue() is not a solution because it breaks formatting, history, and cursor state.

This feature requires only minimal native changes and would greatly improve integration with custom input components.

### PR DESCRIPTION
**Pull request by IslamRustamov**
# Summary
Fixes: #255 
- This PR solves https://github.com/software-mansion/react-native-enriched/issues/255
- This PR adds **insertValue** method that allows users to insert new text at specific parts of text input
- I added new method **insertValue** with 3 arguments (new text to insert, start index of insertion, end index of insertion), added iOS and Android implementation respectfully
- This PR impacts the main files with implementation of text input methods

## Test Plan

Highly recommend to test this PR manually by adding a new button, which invokes a function that calls **insertValue** method.

## Screenshots / Videos

Not applicable.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅      |
| Android |    ✅     |
| Web     |    ✅     |

## Checklist

- [X] E2E tests are passing

NOTE:

There are several questions I would like to ask: 

1) Do we need to call **getActualIndex** on **start** and **end** when we insert text with specified indexes?
2) Does **start** and **end** arguments have to be non nullables?
3) The **handleCommand** in iOS is turning into huge if/else, maybe something should be done about it (not in this PR).

Do please share your suggestions, fixes, test cases and everything else that I may have missed. And don't hesitate to push whatever you want to this PR if you feel like it's needed.

### CODE REVIEW

**Comment by IslamRustamov** [comment]
@kacperzolkiewski check it out at your convenience please (there are some questions in the end of the description)

**Comment by kacperzolkiewski** [comment]
Hi @IslamRustamov, 

To answer your questions:
1. No, we should not call `getActualIndex` when inserting text at specified indices. Because if we would use `getActualIndex` with `start` and  `end` from `onChangeSelection` it won't work properly because `onChangeSelection` returns `start` `end` with ZWS
2. We discussed this with the team and agreed it is the right approach. If a user needs to insert text at the cursor position, they can get the `start` and `end` values from the `onChangeSelection` callback.
3. If you would like to work on this, please go ahead.

**Comment by IslamRustamov** [comment]
> 2. We discussed this with the team and agreed it is the right approach. If a user needs to insert text at the cursor position, they can get the `start` and `end` values from the `onChangeSelection` callback.

Some misunderstanding here sorry, so just to be sure, we want them to be optional or not?

**Comment by kacperzolkiewski** [comment]
@IslamRustamov,
we want them to not be optional :D

**Comment by IslamRustamov** [comment]
i guess its ready for review then, please take a look

File: ios/EnrichedTextInputView.mm [hunk 1]
```diff
@@ -1202,6 +1208,41 @@ - (void)setCustomSelection:(NSInteger)visibleStart end:(NSInteger)visibleEnd {
   textView.selectedRange = NSMakeRange(actualStart, actualEnd - actualStart);
 }
 
+- (void)insertValue:(NSString *)value
```
kacperzolkiewski, line 1353: I think that we should check if `end >= start`.
Because if I call:
`ref?.current?.insertValue('Hello world!', 10, 0)`
The app crashes

kacperzolkiewski:
On Android if we insert text in paragraph style it does not inherit styling:

https://github.com/user-attachments/assets/d1328581-a188-4fc7-b65e-b10c0c8a3afc

On iOS it works properly.

ios/EnrichedTextInputView.mm
IslamRustamov: What do you expect to happen if `end < start`? Just not invoke the function, swap `end` and `start` values or something else

IslamRustamov:
> On Android if we insert text in paragraph style it does not inherit styling:
> 
>  Screen.Recording.2026-02-02.at.16.33.10.mov 
> On iOS it works properly.

this problem is also relevant for pasting functionality. Seems like there is something wrong with **mergeSpannables** which does not correctly set needed span to the inserted text (pasted or inserted) at the end of the string. ~~Though this bug doesn't exist when we paste text at the end string for styles like italic and bold~~

I currently don't see that **mergeSpannables** actually merging spans (I mean, not adding a h1 span to the newly inserted text at the end for example).

I believe that I should copy code of the **handleCustomPaste** and use it in **insertValue**, so it will all come down to modifying **mergeSpannables** in such a way that this problem is solved for both pasting and inserting texts (suggestions on what should be done inside **mergeSpannables** are welcome).

ios/EnrichedTextInputView.mm
kacperzolkiewski: Sorry for late response!
Yes, I think that not invoking a function would be correct.

kacperzolkiewski:
> > On Android if we insert text in paragraph style it does not inherit styling:
> > Screen.Recording.2026-02-02.at.16.33.10.mov
> > On iOS it works properly.
> 
> this problem is also relevant for pasting functionality. Seems like there is something wrong with **mergeSpannables** which does not correctly set needed span to the inserted text (pasted or inserted) at the end of the string. ~Though this bug doesn't exist when we paste text at the end string for styles like italic and bold~
> 
> I currently don't see that **mergeSpannables** actually merging spans (I mean, not adding a h1 span to the newly inserted text at the end for example).
> 
> I believe that I should copy code of the **handleCustomPaste** and use it in **insertValue**, so it will all come down to modifying **mergeSpannables** in such a way that this problem is solved for both pasting and inserting texts (suggestions on what should be done inside **mergeSpannables** are welcome).

@IslamRustamov,
Let’s not worry about this right now. It’s not an issue with your implementation but rather with how spans are merged. It should be addressed in a separate PR.

ios/EnrichedTextInputView.mm
kacperzolkiewski: I applied changes for that

kacperzolkiewski (approved):
Thanks for your work!

temlabs:
Really looking forward to this one! Any indication as to what release we might get this in?

minhnhat165:
I hope the release comes out soon; I really need this.

android/src/main/java/com/swmansion/enriched/textinput/EnrichedTextInputView.kt
```diff
@@ -436,6 +436,39 @@ class EnrichedTextInputView :
     setSelection(actualStart, actualEnd)
   }
 
+  fun insertValue(
```
kacperzolkiewski, line 473: Unfortunately I tested it once again and on Android we need to fix merging spans first before merging this PR:

https://github.com/user-attachments/assets/8fb5716f-01b4-42ff-93c5-3bff8b7fbf93

temlabs: Are you able to provide any guidance on when merging spans might be fixed, and this insertValue method might be rolled out as a result?

This is the single most important feature for our use at the moment.

kacperzolkiewski: Hi @temlabs,

I've just started working on fixing merging spans. It should be ready in the near future.

src/web/EnrichedTextInput.tsx
```diff
@@ -406,6 +412,61 @@ export const EnrichedTextInput = ({
       measureInWindow: () => {},
       measureLayout: () => {},
       setNativeProps: () => {},
+      insertValue: (value: string, start: number, end: number) => {
```
hejsztynx: that file is pretty long already, that logic should be extracted to a separate file

hejsztynx:
1.

There are some inconsistencies how active inline styles are resolved during an insertion. It seems like:

- web: it takes the styles from the selection's end
- ios: it takes the styles from the selection's start
- android: it requires the whole selection's range to be covered by the inline style 

Maybe we could make it more consistent?

web:

https://github.com/user-attachments/assets/faaab779-c04e-4e48-a687-40d8a056ef63

iOS:

https://github.com/user-attachments/assets/67715720-aa60-47df-859e-ce87f2f6b1e0

2.

The text alignment seems to not be properly applied on iOS and Android.

Web (fine):

https://github.com/user-attachments/assets/6419a6ec-4bbd-4c28-97d8-dbddc77e6edd

iOS:

https://github.com/user-attachments/assets/196b6e74-294f-47b3-a428-aa65d06372e3

Android:

https://github.com/user-attachments/assets/b96defa4-ad01-4ce3-85f4-32ef6186381f

3.

Looks like not in this PR scope, but on Android, changes within an inline code style, strips the bold and italics, but keeps the active style state stale

https://github.com/user-attachments/assets/a3503de7-70ca-43da-9f34-622bbd603b5f

hejsztynx (changes requested):

kacperzolkiewski:
@hejsztynx first and second point fixed. 
Right now when inserting text it inherits styles from the start of the selection as it was on iOS.

hejsztynx (approved):
Works really well!

src/web/utils/insertBlockContent.ts
```diff
@@ -0,0 +1,313 @@
...
+ * - Incoming wrappers of the same type as an ancestor (a code block pasted into a code block)
+ *   are unwrapped so their lines merge into the existing wrapper rather than nesting/breaking it.
+ * - A transparent (plain) edge that joins the line adopts the line's block style - including its
+ *   wrapper (code block/blockquote/list) and alignment — while a trailing plain block with no
+ *   text after it stays a plain paragraph.
+ * - A block that can't live in the current wrapper (a heading inside a code block, a code block
+ *   inside a blockquote, …) lifts the line out by splitting the wrapper, leaving siblings wrapped.
+ *
+ * Returns false without meaningfully mutating when the structure can't accept the reconstructed
+ * blocks, so the caller can fall back to TipTap's default insertion.
```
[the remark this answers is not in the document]
hejsztynx: not correct, as `tr` will not be dispatched if it early returns with `false`

PR #417, opened 2026-02-13
fix: revert "fix(iOS): proper layout calculation #339"

szydlovsky:
Reverts https://github.com/software-mansion/react-native-enriched/pull/339 due to a iOS layout bug we found;
The input sometimes becomes unscrollable when it's inside a scrollable bottom sheet (`@gorhom/bottom-sheet`), with a `KeyboardAwareScrollView` (`react-native-keyboard-controller`) turned into a proper bottom-sheet-scrollview component. 

The bug doesn't appear when testing a commit before the PR, but does happen on the exact merge commit of the said PR.

### DIFF
--- a/apps/example/ios/Podfile.lock
+++ b/apps/example/ios/Podfile.lock
@@ -2064,7 +2064,7 @@ EXTERNAL SOURCES:
 
 SPEC CHECKSUMS:
   FBLazyVector: c12d2108050e27952983d565a232f6f7b1ad5e69
-  hermes-engine: 57506c1404e8b80c5386de18743e0d02c8d4f681
+  hermes-engine: c50a496dd5a3974cde1b7d14a944bee1ee64ca0c
   RCTDeprecation: 3280799c14232a56e5a44f92981a8ee33bc69fd9
   RCTRequired: 9854a51b0f65ccf43ea0b744df4d70fce339db32
   RCTSwiftUI: 96986e49a4fdc2c2103929dee2641e1b57edf33d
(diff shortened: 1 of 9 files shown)


IvanIhnatsiuk:
@szydlovsky Could you give more context on what goes wrong?

1. Is it full height input placed in the bottom sheet scroll view? If so, then the scroll breaks because it renders a UIScrollView inside another UIScrollView. Have you tried to give an exact height for one of these scroll views?
2. Could you check the layout in iOS native hierarchy and try to get the layout height value of the input?

NOTE: If you revert these changes, then please test this flow:

1. Open screen A with the enriched input
2. Type any text to make it scrollable and scroll to the end of the 
3. Navigate to screen B
4. Navigate back to screen A and check the scroll position

szydlovsky:
@IvanIhnatsiuk The bug I mentioned is pretty difficult to reproduce. I thought the issue might have been the input just hijacking the scrolling gesture, so I decided to use `scrollEnabled={ false }` and found some likely related issue.

See the following scenario:

We want an input that can be any size, and should always expand instead of scrolling (exactly what's needed in case of input being inside a scroll that is inside a bottom sheet), so naturally `scrollEnabled={ false }` and don't add any height constraints in the input's `style`.
We expect the input to jest freely change sizes.
However, it just stays as a one-liner.

Here's an example from the current main, example app with just removed `maxHeight` style prop and added `scrollEnabled={ false }`:

https://github.com/user-attachments/assets/80bebe49-8a88-44d2-a4a5-ec6a2023348d

And here's the same setup but from the branch that reverses the PR:

https://github.com/user-attachments/assets/ae662c6f-07dc-4a72-a4ed-9e80cae81eef

I believe it's enough for having the PR reverted for now and outweighs just not preserving scroll position on navigation.

EDIT: if `scrollEnabled={ false }` is set, `maxHeight` also doesn't work properly. So we basically lose all the autogrow when the scroll is disabled.

IvanIhnatsiuk:
@szydlovsky okay 🙂

Issue #259, opened 2025-11-17
Text is crammed together in the editor with no whitespace


wongk:
**Describe the bug**
There is no whitespace between elements in the editor. Everything is crammed together very tightly. It looks bad.

**To Reproduce**
See the screenshot.

**Expected behavior**
Even though there are no APIs to control the spacing around different elements in the editor, there should at least be some sensible defaults that looks reasonable in common use cases.

**Screenshots**
<img width="544" height="1036" alt="Image" src="https://github.com/user-attachments/assets/8509f416-28d2-484d-9fd6-c85e260d8c97" />

**Device (please complete the following information):**
 - Device: iPhone 16 Pro simulator
 - OS: iOS 18.5
 - Version 0.1.5

tusharshuklaa:
I also think there should be a way to add styling to regular text (paragraph tag?), like setting a default color, font size, line height, etc.
Looking forward to this bug being fixed.

Thanks for this amazing library

jamesedmonston:
We're pulling in some markup that doesn't necessarily have `<br/>` in between paragraphs, e.g.:

```
<p>line 1</p>
<p>line 2</p>
<p>line 3</p>
```

Currently there doesn't seem to be a way to space these elements out, without rewriting the HTML and forcing breaks?

Anybody solved this at all?

kacperzolkiewski:
Hi @wongk, 
Thanks for reporting this issue!
This PR should resolve your problem: #420

jamesedmonston:
@kacperzolkiewski is there any plan to get this working on the `EnrichedText` component? I can only seem to get it working on the input component.

Also, ideally I'd still like to see a proper `gap`/`margin` prop that allows forcing space _between_ elements – is that on the roadmap at all? To get a good amount of space between element I need to increase `lineHeight` a lot, but that makes paragraphs look a little silly.

Thanks!

<img width="1198" height="2442" alt="Image" src="https://github.com/user-attachments/assets/12ab23aa-4c38-4460-9e80-472b459f64d9" />

jamesedmonston:
@kacperzolkiewski for more context, I got GPT to patch in support for spacing between elements; this is the kind of thing I'd be looking for – independent spacing _and_ line height:

https://github.com/user-attachments/assets/58ae073d-f952-4aba-ab03-799008785dad

Thank you!

kacperzolkiewski:
Hi @jamesedmonston, 

Thank you for the suggestions!
[...]

PR #420, opened 2026-02-17
feat(iOS): add support for lineHeight

kacperzolkiewski:
# Summary
Fixes: #427 
Fixes: #259 
Add lineHeight support on iOS.

## Test Plan

Go to example app and set `lineHeight` to input style props, check if `lineHeight` is applied.

## Screenshots / Videos
`lineHeight: 2`

https://github.com/user-attachments/assets/583a068f-97b7-4457-b0d0-1767d10365e1

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

### DIFF
--- a/ios/EnrichedTextInputView.mm
+++ b/ios/EnrichedTextInputView.mm
@@ -1932,6 +1932,14 @@ Class<RCTComponentViewProtocol> EnrichedTextInputViewCls(void) {
     return NO;
   }
 
+  // Tapping near a link causes iOS to re-derive typingAttributes from
+  // character attributes after textViewDidChangeSelection returns, undoing
+  // the cleanup in manageSelectionBasedChanges. Strip them again here, right
+  // before insertion, so new text never inherits link styling.
+  if (linkStyle != nullptr) {
+    [linkStyle manageLinkTypingAttributes];
+  }
+
   return YES;
 }
 
--- a/android/src/main/java/com/swmansion/enriched/textinput/styles/ParagraphStyles.kt
+++ b/android/src/main/java/com/swmansion/enriched/textinput/styles/ParagraphStyles.kt
@@ -108,6 +108,9 @@ class ParagraphStyles(
     spannable.setSpan(span, safeStart, safeEnd, Spannable.SPAN_EXCLUSIVE_EXCLUSIVE)
   }
 
+  // Removes spans of the given type in the specified range.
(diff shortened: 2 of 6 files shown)


ios/EnrichedTextInputView.mm
```diff
@@ -290,6 +290,28 @@ - (void)setupPlaceholderLabel {
...
+// MARK: - Paragraph style helpers
+
+- (void)applyLineHeightToTypingAttributes:
+    (NSMutableDictionary *)typingAttributes {
+  NSMutableParagraphStyle *paragraphStyle =
+      [[NSMutableParagraphStyle alloc] init];
+  CGFloat lineHeightMultiplier = [config primaryLineHeight];
+  if (lineHeightMultiplier > 0) {
+    CGFloat baseFontSize = [[config scaledPrimaryFontSize] floatValue];
+    CGFloat targetLineHeight = baseFontSize * lineHeightMultiplier;
```
szydlovsky: I kinda don't get that; why do we use `lineHeight` prop as the line height multiplier, instead of just using it as a direct value (maybe multiplied by the scale that font is being multiplied by for the accessibility purposes).

kacperzolkiewski: Yes, I looked on web implementation of `lineHeight` not RN. Fixed.

ios/EnrichedTextInputView.mm
```diff
@@ -731,8 +752,7 @@ - (void)updateProps:(Props::Shared const &)props
         [config primaryColor];
     defaultTypingAttributes[NSStrikethroughColorAttributeName] =
         [config primaryColor];
-    defaultTypingAttributes[NSParagraphStyleAttributeName] =
-        [[NSParagraphStyle alloc] init];
+    [self applyLineHeightToTypingAttributes:defaultTypingAttributes];
```
szydlovsky: We need to do it differently because `primaryLineHeight` in the config doesn't influence the new html generating when `stylePropeChanged` refreshes the html. What we could do is just both add the proper line height to typing attributes as well as apply it over the whole input content here.

kacperzolkiewski: Yes you are right, Thanks ;) 
Fixed

exploIF (approved):
Works well 🙌

PR #422, opened 2026-02-17
feat(iOS): add support for adding custom context menu items

kacperzolkiewski:
# Summary

Add support for `contextMenuItems` prop that allows to add custom buttons in the context native menu.

## Test Plan

1. Run example app
2. Add custom button/buttons through `contextMenuItems` prop
3. Check if they are visible properly and they react on press action properly

## Screenshots / Videos

https://github.com/user-attachments/assets/6064afe8-3cd5-4eb3-be25-a7076405d4ef

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

### DIFF
--- a/ios/EnrichedTextInputView.mm
+++ b/ios/EnrichedTextInputView.mm
@@ -1909,7 +1909,30 @@ Class<RCTComponentViewProtocol> EnrichedTextInputViewCls(void) {
          .selectedText = [selectedText toCppString],
          .selectionStart = static_cast<int>(selectedRange.location),
          .selectionEnd =
-             static_cast<int>(selectedRange.location + selectedRange.length)});
+             static_cast<int>(selectedRange.location + selectedRange.length),
+         .styleState = {
+             .bold = GET_STYLE_STATE([BoldStyle getStyleType]),
+             .italic = GET_STYLE_STATE([ItalicStyle getStyleType]),
+             .underline = GET_STYLE_STATE([UnderlineStyle getStyleType]),
+             .strikeThrough =
+                 GET_STYLE_STATE([StrikethroughStyle getStyleType]),
+             .inlineCode = GET_STYLE_STATE([InlineCodeStyle getStyleType]),
+             .h1 = GET_STYLE_STATE([H1Style getStyleType]),
+             .h2 = GET_STYLE_STATE([H2Style getStyleType]),
+             .h3 = GET_STYLE_STATE([H3Style getStyleType]),
+             .h4 = GET_STYLE_STATE([H4Style getStyleType]),
+             .h5 = GET_STYLE_STATE([H5Style getStyleType]),
+             .h6 = GET_STYLE_STATE([H6Style getStyleType]),
+             .codeBlock = GET_STYLE_STATE([CodeBlockStyle getStyleType]),
+             .blockQuote = GET_STYLE_STATE([BlockQuoteStyle getStyleType]),
+             .orderedList = GET_STYLE_STATE([OrderedListStyle getStyleType]),
+             .unorderedList =
+                 GET_STYLE_STATE([UnorderedListStyle getStyleType]),
+             .link = GET_STYLE_STATE([LinkStyle getStyleType]),
+             .image = GET_STYLE_STATE([ImageStyle getStyleType]),
+             .mention = GET_STYLE_STATE([MentionStyle getStyleType]),
+             .checkboxList =
+                 GET_STYLE_STATE([CheckboxListStyle getStyleType])}});
   }
 }
 
--- a/src/EnrichedTextInput.tsx
+++ b/src/EnrichedTextInput.tsx
@@ -194,9 +194,6 @@ export const EnrichedTextInput = ({
   const nextHtmlRequestId = useRef(1);
   const pendingHtmlRequests = useRef(new Map<number, HtmlRequest>());
 
-  // Track latest style state for context menu callbacks
-  const latestStyleStateRef = useRef<OnChangeStateEvent | null>(null);
-
   // Store onPress callbacks in a ref so native only receives serializable data
   const contextMenuCallbacksRef = useRef<
     Map<string, ContextMenuItem['onPress']>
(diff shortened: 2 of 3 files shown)


src/EnrichedTextInput.tsx
```diff
@@ -78,6 +80,16 @@ export interface EnrichedTextInputInstance extends NativeMethods {
...
 }
 
+export interface ContextMenuItem {
+  text: string;
+  onPress: (
+    text: string,
+    selection: { start: number; end: number },
+    styleState: OnChangeStateEvent
+  ) => void;
+  visible: boolean;
```
exploIF: I think this should be optional (`true` by default)

src/EnrichedTextInput.tsx
```diff
@@ -176,6 +190,44 @@ export const EnrichedTextInput = ({
   const nextHtmlRequestId = useRef(1);
   const pendingHtmlRequests = useRef(new Map<number, HtmlRequest>());
 
+  // Track latest style state for context menu callbacks
+  const latestStyleStateRef = useRef<OnChangeStateEvent | null>(null);
+
+  // Store onPress callbacks in a ref so native only receives serializable data
+  const contextMenuCallbacksRef = useRef<Array<ContextMenuItem['onPress']>>([]);
```
exploIF: I'm not sure about accessing callbacks using array indices. Maybe we should store them inside map and use context menu text as a key? Context item label should be unique

src/EnrichedTextInput.tsx
```diff
@@ -176,6 +190,44 @@ export const EnrichedTextInput = ({
...
+  if (contextMenuItems) {
+    contextMenuCallbacksRef.current = contextMenuItems.map(
+      (item) => item.onPress
+    );
+  } else {
+    contextMenuCallbacksRef.current = [];
+  }
```
exploIF: I don't think we should iterate over items on every render, rather only when `contextMenuItems` changes

src/EnrichedTextInput.tsx
```diff
@@ -176,6 +190,44 @@ export const EnrichedTextInput = ({
...
+      const { index, selectedText, selectionStart, selectionEnd } =
+        e.nativeEvent;
+      const callback = contextMenuCallbacksRef.current[index];
+      if (callback) {
+        callback(
+          selectedText,
+          { start: selectionStart, end: selectionEnd },
+          latestStyleStateRef.current ?? ({} as OnChangeStateEvent)
+        );
+      }
```
exploIF:
```suggestion
      callback?.(
        selectedText,
        { start: selectionStart, end: selectionEnd },
        latestStyleStateRef.current ?? ({} as OnChangeStateEvent)
      );
```

docs/API_REFERENCE.md
```diff
@@ -23,6 +23,41 @@ Tells input to automatically capitalize certain characters.
...
+Item type:
+
+```ts
+interface ContextMenuItem {
+  text: string;
+  onPress: (
+    text: string,
+    selection: { start: number; end: number },
+    styleState: OnChangeStateEvent
+  ) => void;
```
exploIF: Imo it would be better to pass arguments as an object with properties (selectedText, selection, stylesState). I guess it will be common case to use only some of this params (eg. selection, but not use selectedText at all). If we pass it as object it's much easier to skip unused properties

ios/EnrichedTextInputView.mm
```diff
@@ -850,6 +851,31 @@ - (void)updateProps:(Props::Shared const &)props
...
 
+  // contextMenuItems
+  bool contextMenuChanged = newViewProps.contextMenuItems.size() !=
+                            oldViewProps.contextMenuItems.size();
+  if (!contextMenuChanged) {
+    for (size_t i = 0; i < newViewProps.contextMenuItems.size(); i++) {
+      if (newViewProps.contextMenuItems[i].text !=
+              oldViewProps.contextMenuItems[i].text ||
+          newViewProps.contextMenuItems[i].visible !=
+              oldViewProps.contextMenuItems[i].visible) {
```
exploIF: Isn't it enough to do filtering on the JS side?

src/EnrichedTextInput.tsx
kacperzolkiewski: Great idea, fixed

ios/EnrichedTextInputView.mm
kacperzolkiewski: Yes, it will be enough. Fixed

docs/API_REFERENCE.md
```diff
@@ -23,6 +23,41 @@ Tells input to automatically capitalize certain characters.
...
+}
+```
+
+- `text` is the title displayed in the menu.
+- `onPress` is the callback invoked when the item is tapped.
+- `visible` controls whether the item is shown. Defaults to `true`.
+
+| Type                | Default Value | Platform |
+| ------------------- | ------------- | -------- |
+| `ContextMenuItem[]` | -             | iOS      |
```
exploIF:
```suggestion
| `ContextMenuItem[]` | []             | iOS      |
```

src/EnrichedTextInput.tsx
```diff
@@ -176,6 +194,48 @@ export const EnrichedTextInput = ({
   const nextHtmlRequestId = useRef(1);
   const pendingHtmlRequests = useRef(new Map<number, HtmlRequest>());
 
+  // Track latest style state for context menu callbacks
+  const latestStyleStateRef = useRef<OnChangeStateEvent | null>(null);
```
exploIF: I'm not sure about this approach. I feel like it does not cover edge case where `onPress` is called fast enough that style state is still not computed (up to date). 

I was thinking if we could:
- on the native side track if style state has been validated or if it's still being computed
- emit `onPress` event once style state is validated and item has been pressed
- that way we can be sure that styles state is not stale

kacperzolkiewski: Fixed, I emit right now styleState directly from native side without unnecessary refs

PR #441, opened 2026-02-26
feat: add removeLink method

kacperzolkiewski:
# Summary

Add `removeLink(start, end)` method to API

## Test Plan

1. Run example app
2. call `removeLink()` on some link range
3. link should be removed

## Screenshots / Videos
iOS:

https://github.com/user-attachments/assets/171316e7-3bcd-45ab-967c-a1667531ac23

Android:

https://github.com/user-attachments/assets/d980bce6-1e6a-4ab6-9e49-045b9c9ac10f

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |


android/src/main/java/com/swmansion/enriched/textinput/EnrichedTextInputView.kt
```diff
@@ -801,6 +802,25 @@ class EnrichedTextInputView : AppCompatEditText {
...
+  fun removeLink(
+    start: Int,
+    end: Int,
+  ) {
+    val spannable = text as Spannable
+    val textLength = spannable.length
+
+    val clampedStart = start.coerceAtLeast(0)
+    val clampedEnd = end.coerceAtMost(textLength)
+    if (clampedStart > clampedEnd) return
```
exploIF: Why not just support such case and flip start with end?

android/src/main/java/com/swmansion/enriched/textinput/EnrichedTextInputView.kt
```diff
@@ -801,6 +802,25 @@ class EnrichedTextInputView : AppCompatEditText {
     parametrizedStyles?.setLinkSpan(start, end, text, url)
   }
 
+  fun removeLink(
```
exploIF, line 825: I think most of this implementation should be moved into `ParametrizedStyles` which is responsible for handling links related logic

android/src/main/java/com/swmansion/enriched/textinput/EnrichedTextInputView.kt
exploIF: `coerceIn` should do the job, I think

ios/EnrichedTextInputView.mm
```diff
@@ -1542,6 +1546,29 @@ - (void)addLinkAt:(NSInteger)start
...
+  NSInteger textLength = (NSInteger)textView.textStorage.length;
+  if (start < 0) {
+    start = 0;
+  }
+  if (end > textLength) {
+    end = textLength;
+  }
+  if (start > end) {
+    return;
+  }
```
exploIF: Same here, let's just handle this

android/src/main/java/com/swmansion/enriched/textinput/EnrichedTextInputView.kt
kacperzolkiewski: fixed

ios/EnrichedTextInputView.mm

PR #447, opened 2026-02-27
test: add e2e tests setup

pkaramon:
# Summary

Added setup for writing e2e using Maestro framework:
+ added testIds for controls in the example app
+ added button for "Test screen", so we can customize the layout of the example app for tests
+ separated the app into two screen: Dev screen(the same as before but with a "Test button"), Test screen(a special screen made specifically for testing)
+ documented testing in CONTRIBUTING.md
+ added new scripts `test:e2e:android`, `test:e2e:ios`
+ added scripts for automatically creating required emulators/simulators
+ changed bundle identifier to `swmansion.enriched.example`, so it's the same as applicationId on android
+ added example tests

## Test Plan

1. Install maestro
2. Run all tests on android, ios; run individual test using; test `--update-screenshots` using `yarn test:e2e:ios` `yarn:test:e2e:android`.

## Screenshots / Videos

Added test button:

Screenshot 2026-03-05 at 11 18 31

Screenshot 2026-03-05 at 13 16 07

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

### DIFF
--- a/.maestro/scripts/run-tests-all.sh
+++ b/.maestro/scripts/run-tests-all.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 # run-tests-all.sh - run e2e tests on both iOS and Android sequentially.
 #
 # Usage:
@@ -7,13 +7,13 @@
 # All options (--update-screenshots, --rebuild, flow files, etc.) are
 # forwarded to run-tests.sh for each platform.
 
-set -eu
+set -euo pipefail
 
 SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
 
-echo === iOS ===
+echo "=== iOS ==="
 "$SCRIPT_DIR/run-tests.sh" --platform ios "$@"
 
 echo ""
-echo "== android ==="
+echo "=== Android ==="
 "$SCRIPT_DIR/run-tests.sh" --platform android "$@"
--- a/.maestro/scripts/run-tests.sh
+++ b/.maestro/scripts/run-tests.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 # run-tests.sh - set up a device, build the example app, and run Maestro flows.
 #
 # Usage:
@@ -17,7 +17,7 @@
 #   ./run-tests.sh --platform android --update-screenshots .maestro/flows/core_controls_smoke.yaml
 #   ./run-tests.sh --platform ios --rebuild
 
-set -eu
+set -euo pipefail
 
 MIN_MAESTRO_VERSION="2.2.0"
 
@@ -69,7 +69,7 @@ app_installed() {
 }
 
 if [ -n "$REBUILD" ] || ! app_installed; then
-  [ -n "$REBUILD" ] && echo "=== --rebuild requested, building and installing ==="
+  [ -n "$REBUILD" ] && echo "=== rebuild requested, building and installing ==="
   [ -z "$REBUILD" ] && echo "=== App ($BUNDLE_ID) not found, building and installing ==="
   if [ "$PLATFORM" = ios ]; then
     yarn example ios --udid "$DEVICE_ID"
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -80,9 +80,12 @@ We use [Maestro](https://maestro.mobile.dev/) for end-to-end testing. Flows live
 
 #### Prerequisites
 
-- **Maestro CLI** (v2.2.0+) - follow the [Getting Started guide](https://github.com/mobile-dev-inc/maestro?tab=readme-ov-file#getting-started)
-- **iOS** - Xcode
-- **Android** - Android SDK with SDK Command-line Tools, SDK Platform-Tools, Emulator.
+- **Maestro CLI** (v2.2.0+) - follow the [Getting Started guide](https://github.com/mobile-dev-inc/maestro?tab=readme-ov-file#getting-started). After installing, ensure `~/.maestro/bin` is in your `PATH`.
+- **iOS** - Xcode. Ensure `xcrun` is available (it ships with Xcode Command Line Tools).
+- **Android** - Android SDK with SDK Command-line Tools, SDK Platform-Tools, Emulator. Set up `ANDROID_HOME` (typically `$HOME/Library/Android/sdk` on macOS) and ensure the following are in your `PATH`:
+  - `$ANDROID_HOME/cmdline-tools/latest/bin`
+  - `$ANDROID_HOME/platform-tools`
+  - `$ANDROID_HOME/emulator`
 
 The target devices are:
 
@@ -103,7 +106,7 @@ Each command sets up the device and runs all Maestro flows. The script automatic
 
 ```sh
 # Both platforms sequentially
-yarn test:e2e:mobile
+yarn test:e2e
 
 # Single platform
 yarn test:e2e:ios
@@ -126,7 +129,7 @@ Some flows compare a screenshot of the editor against a saved baseline in `.maes
 
 ```sh
 # Update baselines on both platforms
-yarn test:e2e:mobile --update-screenshots
+yarn test:e2e --update-screenshots
 
 # Single platform
 yarn test:e2e:ios --update-screenshots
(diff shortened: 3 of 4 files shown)


apps/example/src/App.tsx
```diff
@@ -349,22 +350,41 @@ export default function App() {
...
         </View>
         <Button
           title="Set input's value"
           onPress={openValueModal}
           style={styles.valueButton}
+          testID="set-value-button"
+        />
+        <Button
+          title={isTestScreen ? 'Dev Screen' : 'Test Screen'}
+          onPress={() => setIsTestScreen((prev) => !prev)}
```
exploIF: I think it would be more readable when we had two separate screens (we can still toggle them using state):
- one for development work - already existing screen
- one for running tests - new screen with helper buttons, inputs, everything we need for testing purposes.

Each of this screen variants should be a different component, so we have clear separation

CONTRIBUTING.md
```diff
@@ -74,6 +74,55 @@ Remember to add tests for your change if possible. Run the unit tests by:
 yarn test
 ```
 
+### E2E tests
+
+We use [Maestro](https://maestro.mobile.dev/) for end-to-end testing. Test flows are located in `.maestro/flows/`, and shared logic lives in `.maestro/subflows/`.
```
exploIF: We should specify minimum maestro version required for running those tests. It looks like it works fine in maestro 2.2.0, but not when using 2.1.0

CONTRIBUTING.md
```diff
@@ -74,6 +74,55 @@ Remember to add tests for your change if possible. Run the unit tests by:
...
+
+Install the Maestro CLI by following the [Getting Started guide](https://github.com/mobile-dev-inc/maestro?tab=readme-ov-file#getting-started).
+
+#### Required devices
+
+| Platform | Device    | OS version                        |
+| -------- | --------- | --------------------------------- |
+| iOS      | iPhone 17 | iOS 26.2                          |
+| Android  | Pixel 9   | API 36.0 "Baklava" (Android 16.0) |
+
```
exploIF: This requirement is pretty specific. I understand that we don't want to lower threshold, but at the same time every time we would like to run tests we have to open specific simulators/emulators.

I would rather suggest creating a bash script that first runs proper simulator/emulator and then executes tests. It can also verify if proper maestro version is installed.

What do you think?

CONTRIBUTING.md
```diff
@@ -74,6 +74,55 @@ Remember to add tests for your change if possible. Run the unit tests by:
...
+```sh
+yarn test:e2e:android
+yarn test:e2e:ios
+```
+
+To run an individual flow, use `--platform <ios|android>` to target a specific platform:
+
+```sh
+maestro test .maestro/flows/placeholder_visual.yaml --platform <ios|android>
```
exploIF: If you decide to create separate script for running tests, I would also suggest handling specific flow execution directly inside this script, so behavior is unified

CONTRIBUTING.md
```diff
@@ -74,6 +74,55 @@ Remember to add tests for your change if possible. Run the unit tests by:
...
+```
+
+#### Visual regression tests
+
+Some flows perform visual regression checks by comparing a screenshot of the editor against a saved baseline in `.maestro/screenshots/`. By default the baseline is asserted; pass `UPDATE_SCREENSHOTS=true` to save new baselines instead.
+
+For a single flow:
+
+```sh
+maestro test .maestro/flows/inline_styles_visual.yaml --platform <ios|android> --env UPDATE_SCREENSHOTS=true
```
exploIF: Let's add `--update-screenshots` option to our test script rather than setting up env manually

apps/example/src/App.tsx
pkaramon: Great idea. I've separated the two screens into different components.

CONTRIBUTING.md
pkaramon: Great observation. I've added the required version to the docs, also script for running tests checks that the version is at least 2.2.0.

CONTRIBUTING.md
pkaramon: I've created scripts for setting up simulators/emulators and running the tests.

CONTRIBUTING.md
pkaramon: Testing is now handled by `yarn test:e2e:ios` and `yarn test:e2e:android` and developers just pass necessary flags, files to those commands.

CONTRIBUTING.md
pkaramon: Done

package.json
```diff
@@ -49,7 +49,10 @@
     "prepare": "bob build",
     "release": "release-it",
     "android-studio": "open -a 'Android Studio' apps/example/android",
-    "xcode": "open -a 'Xcode' apps/example/ios/EnrichedTextInputExample.xcworkspace"
+    "xcode": "open -a 'Xcode' apps/example/ios/EnrichedTextInputExample.xcworkspace",
+    "test:e2e:mobile": "sh .maestro/scripts/run-tests-all.sh",
```
exploIF:
```suggestion
    "test:e2e": "sh .maestro/scripts/run-tests-all.sh",
```

.maestro/scripts/run-tests.sh
```diff
@@ -0,0 +1,88 @@
...
+app_installed() {
+  if [ "$PLATFORM" = ios ]; then
+    xcrun simctl listapps "$DEVICE_ID" 2>/dev/null | grep -q "$BUNDLE_ID"
+  else
+    adb -s "$DEVICE_ID" shell pm list packages "$BUNDLE_ID" 2>/dev/null | grep -q "$BUNDLE_ID"
+  fi
+}
+
+if [ -n "$REBUILD" ] || ! app_installed; then
+  [ -n "$REBUILD" ] && echo "=== --rebuild requested, building and installing ==="
```
exploIF:
```suggestion
  [ -n "$REBUILD" ] && echo "=== rebuild requested, building and installing ==="
```

package.json
```diff
@@ -49,7 +49,10 @@
...
+    "test:e2e:android": "sh .maestro/scripts/run-tests.sh --platform android",
+    "test:e2e:ios": "sh .maestro/scripts/run-tests.sh --platform ios"
```
exploIF: Do we really have to start any of this scripts with `sh`? Unless they are executable we should be fine to skip this part

.maestro/scripts/run-tests.sh
```diff
@@ -0,0 +1,88 @@
...
+  esac
+done
+
+[ -z "$FLOWS" ] && FLOWS=".maestro/flows"
+
+case "$PLATFORM" in
+  ios)     SETUP="$SCRIPT_DIR/setup-ios-simulator.sh" ;;
+  android) SETUP="$SCRIPT_DIR/setup-android-emulator.sh" ;;
+  *)       echo "Error: --platform must be ios or android" >&2; exit 1 ;;
+esac
```
exploIF, line 59: If any of this fails we still continue executing tests. Instead we should exit immediately. I believe setting up below should solve the issue:
```bash
#!/bin/bash

set -euo pipefail
```

CONTRIBUTING.md
```diff
@@ -74,6 +74,67 @@ Remember to add tests for your change if possible. Run the unit tests by:
...
 
+### E2E tests
+
+We use [Maestro](https://maestro.mobile.dev/) for end-to-end testing. Flows live in `.maestro/flows/` and shared subflows in `.maestro/subflows/`.
+
+#### Prerequisites
+
+- **Maestro CLI** (v2.2.0+) - follow the [Getting Started guide](https://github.com/mobile-dev-inc/maestro?tab=readme-ov-file#getting-started)
+- **iOS** - Xcode
+- **Android** - Android SDK with SDK Command-line Tools, SDK Platform-Tools, Emulator.
```
exploIF: I think it's worth explicitly providing what should be added to `PATH`

.maestro/scripts/run-tests-all.sh
```diff
@@ -0,0 +1,19 @@
...
+#   ./run-tests-all.sh [options] [flow ...]
+#
+# All options (--update-screenshots, --rebuild, flow files, etc.) are
+# forwarded to run-tests.sh for each platform.
+
+set -eu
+
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+
+echo === iOS ===
```
exploIF:
```suggestion
echo "=== iOS ==="
```

.maestro/scripts/run-tests.sh
pkaramon: Good catch

package.json
pkaramon: True, `sh` is not needed

CONTRIBUTING.md
pkaramon: I've updated docs with information on what should be included in `PATH`.

PR #474, opened 2026-03-09
test: add tests for recent bugs

pkaramon:
# Summary

+ wrote tests for fixes in PRs: 405, 424, 462, 430,  428
+ added button "Max|Base" in test screen. In "Base" state it behaves just as before, with the same height, in "Max" the input is allowed to be a lot bigger(in terms of height). Useful in some tests where we have more text.
+ modals now in test screen avoid keyboard, so Maestro is able to click on inputs, buttons in them. This keyboard avoiding functionality is implemented in `ModalShell` which is then used by all modal components.
+ added tests ids to modal inputs and buttons

## Test Plan

Run `yarn test:e2e` and verify that tests pass. Also we should run `yarn test:e2e --update-screenshots` and see if any screenshots were altered, this is because as of right now Maestro has a bug, where it does not fail when baseline image and newly capture have different dimensions. 

## Screenshots / Videos

EnrichedInput is allowed to be a lot bigger in "Max" mode.

Screenshot 2026-03-09 at 15 52 04

Modals now in test screen avoid keyboard, so all the buttons and fields are accessible by Maestro.

Screenshot 2026-03-09 at 15 54 22

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -138,6 +138,13 @@ yarn test:e2e:android --update-screenshots .maestro/flows/inline_styles_visual.y
 
 Always review newly saved screenshots in `.maestro/screenshots/` before committing them.
 
+#### Troubleshooting: flaky Android tests on macOS
+
+macOS may throttle the Android emulator via **App Nap** when its window is not visible (e.g. minimized or behind other windows), which can cause test timeouts. Two workarounds:
+
+1. **Keep the emulator window visible** while tests are running.
+2. **Disable App Nap** for the emulator: `defaults write com.google.android.emulator NSAppSleepDisabled -bool YES` (requires an emulator restart). Note this may drain your battery, so you may want to re-enable it afterwards with `-bool NO`.
+
 ### Commit message convention
 
 We follow the [conventional commits specification](https://www.conventionalcommits.org/en) for our commit messages:


kacperzolkiewski:
Do some test cases occasionally fail on your side as well (Android)? On my first run, the following test cases failed:
```
[Failed] paragraph_style_toggle (22s) (Comparison error: Assert screenshot matches .maestro/screenshots/android/paragraph_style_toggle.png (threshold: 100.0%) (cropped on id: editor-input) - threshold not met, current: 99.39644%)
[Failed] initial_html_parsing (24s) (Element not found: Id matching regex: value-modal-input)
```
On the second time:
```
[Failed] image_position_stability (40s) (Element not found: Text matching regex: Photo taken.*)
[Failed] core_controls_smoke (23s) (Element not found: Id matching regex: toggle-screen-button)
```

pkaramon:
Interesting. The only failure I was able to reproduce (sadly, only once) is: `[Failed] paragraph_style_toggle (22s)`. In my case, the diff looks like this:  
paragraph_style_toggle_diff

It’s quite bizarre, as it contains `"0"`, which isn’t anywhere in the test file. @kacperzolkiewski, could you share your `paragraph_style_toggle_diff.png`?

As for the rest, I think sometimes Android takes unusually long for certain elements to appear, which might cause Maestro to exceed its default timeouts. I’ll try using `waitForAnimationToEnd` or an explicit `extendedWaitUntil` to see if that helps. I’ll also attempt to finally reproduce these failures.

apps/example/src/components/ModalShell.tsx
```diff
@@ -0,0 +1,46 @@
...
+  const inner = avoidKeyboard ? (
+    <KeyboardAvoidingView
+      style={styles.container}
+      behavior={Platform.OS === 'ios' ? 'padding' : 'height'}
+    >
+      {children}
+    </KeyboardAvoidingView>
+  ) : (
+    <View style={styles.container}>{children}</View>
+  );
```
exploIF: Do we really have to render this content conditionally. Isn't it enough to pass  [enabled](https://reactnative.dev/docs/keyboardavoidingview#enabled) prop?

apps/example/src/screens/TestScreen.tsx
```diff
@@ -45,12 +47,22 @@ export function TestScreen({ onSwitch }: TestScreenProps) {
             style={styles.button}
             testID="clear-button"
           />
+          <Button
+            title={sizeMode === 'max' ? 'Base' : 'Max'}
+            onPress={() => setSizeMode(sizeMode === 'max' ? 'base' : 'max')}
+            style={styles.button}
+            testID="size-max-button"
+          />
```
exploIF, line 55: It's more like an open question. But maybe we should think about integrating some navigation library with our example app and providing more screens which simplifies testing and development? I'm just afraid that we will have tons of conditions once we have more tests

.maestro/flows/image_position_stability.yaml
```diff
@@ -0,0 +1,51 @@
...
+
+- inputText: "Paragraph one"
+- pressKey: Enter
+- pressKey: Enter
+
+- inputText:  "Paragraph two"
+
+- tapOn:
+    id: "editor-input"
+    point: "50%,50%"
```
exploIF, line 25: I see that in many places we use coordinates as selectors. Is it possible to rely on text based selectors? Or is it not possible to query by subtext?

apps/example/src/components/ModalShell.tsx
pkaramon: Good point, you're right. I've removed the conditional and instead pass `enabled` prop.

.maestro/flows/image_position_stability.yaml
pkaramon: I also don't like those coordinates as selectors. It's not possible to rely on text based selectors, because maestro sees our input as simply: 
iOS:
```json
{
    "attributes" : {
      "accessibilityText" : "Type something here..., Vertical scroll bar, 2 pages, Horizontal scroll bar, 1 page",
      "title" : "",
      "value" : "",
      "text" : "",
      "hintText" : "",
      "resource-id" : "editor-input",
      "bounds" : "[16,152][386,332]",
      "enabled" : "true",
      "focused" : "true",
      "selected" : "false",
      "checked" : "false"
    },
    "children" : [ ],
    "enabled" : true,
    "focused" : true,
    "checked" : false,
    "selected" : false
}
```
Android: 
```json
 {
  "attributes" : {
    "text" : "​Hello\n​World\n!!!\nHello\n\n​1\n​1",
    "accessibilityText" : "",
    "hintText" : "Type something here...",
    "ignoreBoundsFiltering" : "false",
    "resource-id" : "editor-input",
    "clickable" : "true",
    "bounds" : "[42,530][1038,1404]",
    "enabled" : "true",
    "focused" : "true",
    "checked" : "false",
    "scrollable" : "false",
    "selected" : "false",
    "class" : "android.widget.EditText",
    "important-for-accessibility" : "true",
    "error" : ""
  },
  "children" : [ ],
  "clickable" : true,
  "enabled" : true,
  "focused" : true,
  "checked" : false,
  "selected" : false
}
```
So we could only `tapOn` etc. the entire `editor-input`. And also only on android because on iOS the `text` attribute is always empty. Even if the `text` attribute worked the same on both platforms it still wouldn't be useful because maestro wouldn't know where exactly each substring is positioned.

apps/example/src/screens/TestScreen.tsx
pkaramon: I see your point. I think that adding navigation library right now would probably be an overkill, but it's something that may be useful in the future. I think that in this scenerio the Base/Max button will be useful for lots of tests because frequently the default height isn't enough to show everything of interest. I would consider adding more screens if we find ourselves needing to add more and more additional buttons, controls etc.

kacperzolkiewski:
> It’s quite bizarre, as it contains "0", which isn’t anywhere in the test file. @kacperzolkiewski, could you share your paragraph_style_toggle_diff.png?

@pkaramon  Never mind, I tried to run tests today and they work :D

pkaramon:
@kacperzolkiewski I think I may know why your tests were failing. I was playing around with tests today and noticed something. If you minimize the emulator or even if it's behind something(for example VS Code window) Mac OS can turn on "App Nap":
Screenshot 2026-03-11 at 09 25 40

Then basically the app is way too slow to respond to Maestro and tests may fail. When emulator was put into "App Nap" I managed to fail all the tests on android all with the message: `Element not found: Id matching regex: toggle-screen-button` basically meaning the app didn't even manage to fully load. 

I don't really know what to do about this:
+ one option would be to add a note in the docs asking people to keep their Android emulator visible during tests
+ we could also recommend turning off App Nap for android emulators `defaults delete com.google.android.emulator NSAppSleepDisabled` but that might impact battery life

@kacperzolkiewski  @exploIF what do you think?

PR #475, opened 2026-03-10
 Set up publish workflow that can be used for both nightly and manual release

kacperzolkiewski:
# Summary

This PR refactor `publish.yml` and uses `npm-package-publish` for nightly and manual publish.

Job success: https://github.com/software-mansion/react-native-enriched/actions/runs/22892021685/job/66417143113

## Test Plan

Run Publish workflow

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -38,9 +38,9 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Setup Node
-        uses: actions/setup-node@v4
+        uses: actions/setup-node@v6
         with:
-          node-version: .nvmrc
+          node-version-file: .nvmrc
           cache: 'yarn'
           registry-url: https://registry.npmjs.org/
 


.github/workflows/publish.yml
```diff
@@ -1,70 +1,66 @@
...
-      - name: Upload NPM tarball as artifact
-        uses: actions/upload-artifact@v4
+      - name: Setup Node
+        uses: actions/setup-node@v6
         with:
-          name: ${{ steps.meta.outputs.PACKAGE_NAME }}
-          path: ${{ steps.meta.outputs.PACKAGE_NAME }}
-          overwrite: true
-          if-no-files-found: error
+          node-version: 24
```
exploIF: This is pretty unexpected for me, shouldn't we use `.nvmrc` version? https://github.com/software-mansion/react-native-enriched/blob/main/.nvmrc

exploIF:
Could you please edit PR name? It provides `publish` workflow that can be used for both nightly and regular/manual release

.github/workflows/publish.yml
kacperzolkiewski: Good point. fixed!

exploIF (approved):

Issue #365, opened 2026-01-13
[Feature request] Support for Text Alignment


Rubiojen:
**Is your feature request related to a problem? Please describe.**
Yes. The EnrichedTextInput component forces all text to be left-aligned. It ignores standard textAlign: 'center' style properties and lacks methods to adjust paragraph alignment. This makes it impossible to implement centered layouts (e.g., titles, poetry, or social media captions).

**Describe the solution you'd like**
I would like EnrichedTextInput to support the standard textAlign style prop (mapping to Gravity.CENTER on Android and textAlignment = .center on iOS). Alternatively, expose a toggleAlign('center' | 'right' | 'left') method on the ref.

**Describe alternatives you've considered**
I attempted to use setNativeProps to force alignment, but it is ignored. I also tried a "Ghost Text" wrapper to calculate width, but the internal text inside the editor remains left-aligned on multiline inputs.

**Additional context**
This is critical for apps requiring "Instagram-style" text overlays where content must be centered on the screen regardless of line length.

Wmg20:
Hi @exploIF 👋

This PR support for Text Alignment (#365)! 🚀

**What's inside:**

- **Core API**: Exposed `setParagraphAlignment(alignment)` to the JS Ref and added alignment tracking to the state events.
- **Android**: Mapped string alignments (`left`, `center`, `right`) to native `Layout.Alignment` spans (`ALIGN_NORMAL`, `ALIGN_CENTER`, `ALIGN_OPPOSITE`).
- **iOS**: Implemented native `NSTextAlignment` using `NSMutableParagraphStyle`.
- **Example App**: Updated the floating toolbar to include alignment buttons that dynamically update based on the current cursor position.

The commits are separated logically by layer (Core API -> Android -> iOS -> Example/Docs) to make reviewing easier.

I've tested this thoroughly on both platforms. Let me know if you need any changes or have any feedback! Thanks for taking a look. 🙏

MartycFly321:
@Wmg20 I think you tagged the wrong issue/PR in your comment,
You mean https://github.com/software-mansion/react-native-enriched/pull/488 right?

Thanks for working on this

kacperzolkiewski:
Hi @Rubiojen, 
Thanks for bringing this up!
I'm closing this issue since the feature has already been implemented and is ready to use:
https://github.com/software-mansion/react-native-enriched-html/pull/403
https://github.com/software-mansion/react-native-enriched-html/pull/602
[...]

PR #478, opened 2026-03-10
Support for Text Alignment #365

Wmg20:
# Summary

Support for feature request: #365

This PR introduces text alignment support (`left`, `center`, `right`) to `EnrichedTextInput`.

**Features & Implementation:**

- **Core API**: Added `setParagraphAlignment(alignment)` to the JS Ref and included alignment structures in the text change events.
- **Android**: Handled paragraph alignment using Android `Layout.Alignment` spans (`ALIGN_NORMAL`, `ALIGN_CENTER`, `ALIGN_OPPOSITE`). Emits active alignment state back to the JS context.
- **iOS**: Implemented native `NSTextAlignment` via `NSMutableParagraphStyle` and updated the active alignment state when cursor location changes.
- **Example App**: Included Left, Center, and Right buttons in the floating toolbar, correctly updating to reflect the cursor's current line alignment. Updated `API_REFERENCE.md` to document the new functionality.

## Test Plan

1. Compile and run the `example` application.
2. Type multiple paragraphs.
3. Tap the alignment buttons in the floating toolbar and verify the text correctly aligns to Left, Center, and Right.
4. Move the cursor between differently aligned paragraphs and verify the active button in the toolbar dynamically updates.
5. Verify behavior on both platforms (iOS and Android).

## Screenshots / Videos
android-text-alignment
iOS-text-alignment

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |     ✅      |
| Android |     ✅      |

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,70 +1,66 @@
 name: Publish
 on:
+  # For nightly releases
+  schedule:
+    - cron: '42 23 * * *' # at 23:42 every day
+  # For manual releases
   workflow_dispatch:
     inputs:
-      dryRun:
-        description: 'Set to true to perform a dry run'
+      release-type:
+        description: Type of release to publish.
+        type: choice
+        options: [stable, nightly, beta, rc]
+        default: stable
+      version:
+        description: Specific version to publish (usually inferred from x.y-stable branch name).
+        type: string
         required: false
-        default: 'true'
-
-permissions:
-  id-token: write
-  contents: read
+        default: ''
+      dry-run:
+        description: Whether to perform a dry run of the publish.
+        type: boolean
+        default: true
 
 jobs:
-  verify-and-publish:
+  publish:
     runs-on: ubuntu-latest
-    environment: production
+
+    permissions:
+      contents: write
+      id-token: write # for OIDC
+
+    concurrency:
+      group: publish-${{ github.ref }}
+      cancel-in-progress: false
+
     steps:
-      - name: Checkout
+      - name: Check out
         uses: actions/checkout@v4
 
-      - name: Setup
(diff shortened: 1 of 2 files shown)


kacperzolkiewski:
Hello @Wmg20,
Thank you for opening this PR, we really appreciate the contribution!

I wanted to let you know that we already have an iOS implementation of text alignment in #403, which is currently awaiting review. If it gets merged, we'd love to have the Android side implemented in a similar fashion so both platforms stay consistent.
A few key points from the iOS approach that would be great to align on:

1. List alignment is applied as a unit:
- when the cursor is inside a list, the alignment change is applied to all contiguous list items, not just the current paragraph.
2. List markers stay on the left:
- regardless of text alignment (right, center, etc.), bullets / numbers / checkboxes are always drawn at the left margin.

It would be great if you could take a look at #403 and adapt the Android implementation to follow the same conventions. That way we keep the behavior consistent across platforms.

Thanks again for your effort and feel free to reach out if you have any questions!

Wmg20:
Hi @kacperzolkiewski , 

Thanks for the clarification and the detailed context.

I’ll take a look at #403 and align the Android implementation with the iOS approach so the behavior stays consistent across platforms. In particular, I’ll make sure that:

List alignment is applied to the entire contiguous list, not just the current paragraph when the cursor is inside a list.

List markers remain left-aligned, regardless of the selected text alignment (center, right, etc.).

I’ll update the Android implementation accordingly and push the changes.
Thanks again for pointing this out!

kacperzolkiewski:
@Wmg20 I’ll close this PR since you’ve opened the Android version: https://github.com/software-mansion/react-native-enriched/pull/488

PR #481, opened 2026-03-11
feat(iOS): attributes remake

szydlovsky:
# Summary
This PR remakes iOS attributes handling to simplify style logic and make styling behavior more consistent and maintainable.
## Main points
- Introduces `AttributesManager` to centralize:
  - dirty range tracking
  - typing-attribute management
  - styling reapplication flow
- Refactors style architecture with a shared `StyleBase` contract (`getType`, `getKey`, `getValue`, `detect`, `add`, `all`, etc.).
- Generalizes default style behavior in base implementations, custom styles (links/mentions/images) keep custom handling.
- Splits responsibilities:
  - `add` = metadata/meta-attributes
  - `applyStyling` = visual rendering only
- Adds robust dirty-range processing:
  - merge/dedupe/sort ranges
  - shift ranges when text length changes
  - reset + reapply styles deterministically
- Improves typing-attribute continuation so custom inline styles propagate correctly.
- Integrates the new flow into selection updates, text editing (`didProcessEditing`), style toggles, and final post-edit styling pass.
- Includes utility cleanup, renames (`ParagraphUtils` -> `RangeUtils`) and method-name unification across styles.
## Test Plan
It is big refactor so as a test plan I would test if all of the styles work properly :D 

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

### DIFF
--- a/.maestro/flows/scrolling_after_typing.yaml
+++ b/.maestro/flows/scrolling_after_typing.yaml
@@ -43,6 +43,8 @@ appId: swmansion.enriched.example
       id: "editor-input"
     direction: DOWN
 
+- waitForAnimationToEnd
+
 - runFlow:
     file: "../subflows/capture_or_assert_screenshot.yaml"
     env:
@@ -56,6 +58,8 @@ appId: swmansion.enriched.example
     from:
       id: "editor-input"
     direction: UP
+
+- waitForAnimationToEnd
 
 - runFlow:
     file: "../subflows/capture_or_assert_screenshot.yaml"


ios/extensions/LayoutManagerExtension.mm
```diff
@@ -299,23 +299,27 @@ - (void)drawLists:(EnrichedTextInputView *)typedInput
...
                                            markerAttributes:markerAttributes
                                                      origin:origin
-                                                   usedRect:textUsedRect];
-                                     } else if (markerFormat ==
-                                                NSTextListMarkerDisc) {
+                                                   usedRect:usedRect];
+                                     } else if ([markerFormat
+                                                    isEqualToString:
+                                                        @"EnrichedUnorderedLis"
+                                                        @"t"]) {
```
szydlovsky, line 316: This got formatted in a super weird way. Can we maybe manually leave it as a single string please

ios/interfaces/ImageAttachment.mm
```diff
@@ -1,6 +1,16 @@
 #import "ImageAttachment.h"
 #import "ImageExtension.h"
 
+static NSCache<NSString *, UIImage *> *ImageAttachmentCache(void) {
```
szydlovsky, line 10: Let's add a comment explaining why we need that cache.

ios/interfaces/ImageAttachment.mm
```diff
@@ -1,6 +1,16 @@
...
+  static NSCache<NSString *, UIImage *> *cache = nil;
+  static dispatch_once_t onceToken;
+  dispatch_once(&onceToken, ^{
+    cache = [[NSCache alloc] init];
+    cache.totalCostLimit = 50 * 1024 * 1024; // 50 MB
```
szydlovsky: I'd rather make it 100MB because we risk nothing by doing that. Someone might be using multiple editors on one screen, each with high-res images -> scrolling it might then be laggy with the 50MB because we would constantly remove and re-decode images to keep it all under this constraint. 
Another upside is, if the system actually needs the memory, it can decide to remove some stuff from this cache because this API talks directly to the operating system.

ios/interfaces/StyleBase.mm
```diff
@@ -0,0 +1,254 @@
...
+#import "ZeroWidthSpaceUtils.h"
+
+@implementation StyleBase
+
+// This method gets overridden
++ (StyleType)getType {
+  return None;
+}
+
+// This method gets overridden
```
szydlovsky:
```suggestion
// This method gets overridden for inline styles
```

ios/interfaces/StyleBase.mm
```diff
@@ -0,0 +1,254 @@
...
+                   withInput:_input
+                     inRange:range
+               withCondition:^BOOL(id _Nullable value, NSRange range) {
+                 return [self styleCondition:value range:range];
+               }];
+}
+
+// This method gets overridden
+- (void)applyStyling:(NSRange)range {
+}
```
szydlovsky, line 234:
```suggestion
- (void)applyStyling:(NSRange)range {}
```

ios/interfaces/StyleBase.mm
```diff
@@ -0,0 +1,254 @@
...
+    NSMutableParagraphStyle *pStyle =
+        [newTypingAttrs[NSParagraphStyleAttributeName] mutableCopy];
+    pStyle.textLists = @[];
+    newTypingAttrs[NSParagraphStyleAttributeName] = pStyle;
+  }
+
+  _input->textView.typingAttributes = newTypingAttrs;
+}
+
+- (BOOL)styleCondition:(id)value range:(NSRange)range {
```
szydlovsky, line 183: Let's also add a comment that custom styles will most likely need it overridden.

ios/styles/StrikethroughStyle.mm
```diff
@@ -1,102 +1,24 @@
...
-  return [OccurenceUtils all:NSStrikethroughStyleAttributeName
-                   withInput:_input
-                     inRange:range
-               withCondition:^BOOL(id _Nullable value, NSRange range) {
-                 return [self styleCondition:value range:range];
-               }];
+- (void)applyStyling:(NSRange)range {
+  NSDictionary *styles =
+      @{NSStrikethroughStyleAttributeName : @(NSUnderlineStyleSingle)};
+  [self.input->textView.textStorage addAttributes:styles range:range];
```
szydlovsky: Nit: we can just make it a one liner, like `addAttributes: @{key:value}`

ios/styles/HeadingStyleBase.mm
```diff
@@ -1,310 +1,68 @@
...
-#import "OccurenceUtils.h"
-#import "ParagraphsUtils.h"
 #import "StyleHeaders.h"
 #import "TextInsertionUtils.h"
 
 @implementation HeadingStyleBase
 
-// mock values since H1/2/3/4/5/6Style classes anyway are used
-+ (StyleType)getStyleType {
+// mock values since H1/2/3/4/5/6Style classes are used
```
szydlovsky:
```suggestion
// mock values since H1/2/3/4/5/6 style classes are used
```

ios/interfaces/StyleHeaders.h
```diff
@@ -1,64 +1,58 @@
...
 - (NSRange)getFullMentionRangeAt:(NSUInteger)location;
 - (NSValue *)getActiveMentionRange;
 @end
 
-@interface HeadingStyleBase : NSObject <BaseStyleProtocol> {
-  id input;
-}
+@interface HeadingStyleBase : StyleBase
 - (CGFloat)getHeadingFontSize;
 - (NSString *)getHeadingLevelString;
```
szydlovsky: Unused method.

ios/inputParser/InputParser.mm
szydlovsky, line 716: Xcode spits warnings for unclosed tags... even in the comments. Lets just keep them as `strong` instead of `<strong>`.

ios/interfaces/StyleHeaders.h
szydlovsky, line 1: General comment for all paragraph styles: as per irl discussion we had, all paragraph styles need to have their `applyStyling` done ONLY on the provided range, never extending beyond it.

ios/EnrichedTextInputView.mm
```diff
@@ -1036,19 +1033,19 @@ - (void)tryUpdatingActiveStyles {
   // separately
   NSMutableSet *newBlockedStyles = [_blockedStyles mutableCopy];
 
-  // data for onLinkDetected event
+  //  // data for onLinkDetected event
```
szydlovsky: Why double comment here?

```suggestion
// data for onLinkDetected event
```

ios/EnrichedTextInputView.mm
```diff
@@ -1142,45 +1139,42 @@ - (void)tryUpdatingActiveStyles {
...
 
   if (detectedLinkData != nullptr) {
-    // emit onLinkeDetected event
     [self emitOnLinkDetectedEvent:detectedLinkData.text
                               url:detectedLinkData.url
                             range:detectedLinkRange];
   }
 
   if (detectedMentionParams != nullptr) {
-    // emit onMentionDetected event
```
szydlovsky, line 1175: Let's keep these comments.

ios/EnrichedTextInputView.mm
```diff
@@ -98,166 +99,158 @@ - (void)setDefaults {
...
-    @([H4Style getStyleType]) : @[],
-    @([H5Style getStyleType]) : @[],
-    @([H6Style getStyleType]) : @[],
-    @([UnorderedListStyle getStyleType]) : @[],
-    @([OrderedListStyle getStyleType]) : @[],
-    @([CheckboxListStyle getStyleType]) : @[],
-    @([BlockQuoteStyle getStyleType]) : @[],
-    @([CodeBlockStyle getStyleType]) : @[],
-    @([ImageStyle getStyleType]) : @[ @([InlineCodeStyle getStyleType]) ]
+    @([BoldStyle getType]) : @[ @([CodeBlockStyle getType]) ],
```
szydlovsky, line 226: I know it probably was like this before the change, but I found a thing we could improve. We properly update the `blockingStyles` when some heading style gets its `bold` htmlStyle prop changed. However we don't anticipate a situation where the bold is already mixed with heading and then we change the prop.
Try the following:
- add some text with heading, that has its `htmlStyle`'s `bold` set to `false`
- also add bold to the same text
- now change the `htmlStyle` to set the `bold` for the heading to `true`
- do something that emits a new html, e.g. add some character
- observe that `<b>` is still there. This means that we properly block the bold but don't remove the style.

It can actually be easily fixed, because changing `htmlStyle` always triggers input rebuild via `inputParser`. So all we have to do is also make `conflictingStyles` mutable, just like `blockingStyles` are. And now at the same time blocking styles are updated, we also need to remove or add `bold` to proper heading's conflicting styles. This ensures input rebuild will remove this bold style from there. Ideally we can create two methods `addStyleConflict` and `removeStyleConflict` based on existing `addStyleBlock` and `removeStyleBlock`.

ios/EnrichedTextInputView.mm
```diff
@@ -1710,32 +1678,26 @@ - (void)manageSelectionBasedChanges {
...
-    }
-  }
+  // attributes manager handles proper typingAttributes at all times to properly
+  // extend meta-attributes
+  BOOL onlySelectionChanged =
+      textView.selectedRange.length == 0 &&
+      [_recentInputString isEqualToString:currentString];
+  // We want to remember which attributes were removed as long as we stay at the
+  // same position. This prevents a removed attribute from being re-applied from
+  // the preceding character right after we toggled it off
```
szydlovsky: Good explanation but I'd also mention the fact that `removedTypingAttributes` aren't normally removed during the regular flow, and we do remove them only here - so when we are sure that selection changed.

ios/styles/InlineCodeStyle.mm
szydlovsky, line 1: I have another rather open question but a valid one nonetheless. In the new implementation, newlines may or may not have the meta-attributes applied to them even if it actually means nothing. This way, in the new impl we can do something like:
- add like 3 paragraphs and, command + A and toggle inline code everywhere
- see style is properly detected because the metaatribute has been applied everywhere, to the newlines too
- now select one of the newline characters and switch inline code off
- now if you select the whole input the inline code will be shown as not detected, even though it looks like it is everywhere
- if you now press toggle once, inline code will now be re-applied to the missing newline and we will have it detected, even though nothing visually changed for the user

I think it's kinda misleading and we might think of some improvement maybe for all the styles, not only inline code. So to not apply attributes to newlines and omit them when detecting a style. But let's discuss it first because it might have some parser implications...

ios/EnrichedTextInputView.mm
```diff
@@ -2025,32 +1967,32 @@ - (bool)textView:(UITextView *)textView
...
+  H2Style *h2Style = stylesDict[@([H2Style getType])];
+  H3Style *h3Style = stylesDict[@([H3Style getType])];
+  H4Style *h4Style = stylesDict[@([H4Style getType])];
+  H5Style *h5Style = stylesDict[@([H5Style getType])];
+  H6Style *h6Style = stylesDict[@([H6Style getType])];
 
   // some of the changes these checks do could interfere with later checks and
   // cause a crash so here I rely on short circuiting evaluation of the logical
   // expression either way it's not possible to have two of them come off at the
   // same time
```
szydlovsky: small update:
```suggestion
  // some of the changes these checks do could interfere with later checks and
  // cause a crash so here we rely on short circuiting evaluation of the logical
  // expression. Either way it's not possible to have two of them come off at the
  // same time
```

ios/EnrichedTextInputView.mm
```diff
@@ -1761,27 +1723,8 @@ - (void)anyTextMayHaveBeenModified {
...
-    [bqStyle manageBlockquoteColor];
-  }
-
-  // codeblock font and color management
-  CodeBlockStyle *codeBlockStyle = stylesDict[@([CodeBlockStyle getStyleType])];
-  if (codeBlockStyle != nullptr) {
-    [codeBlockStyle manageCodeBlockFontAndColor];
-  }
-
-  // mentions management: removal and editing
```
szydlovsky, line 1782: we can keep that comment too

ios/utils/ParagraphAttributesUtils.mm
```diff
@@ -53,41 +41,25 @@ + (BOOL)handleBackspaceInRange:(NSRange)range
   if (range.location == nonNewlineRange.location &&
       range.length >= nonNewlineRange.length) {
 
-    // for lists, quotes and codeblocks present we do the following:
+    // for styles that need ZWS (lists, quotes, etc.) we do the following:
     // - manually do the removing
     // - reset typing attribtues so that the previous line styles don't get
```
szydlovsky: typo 😄 
```suggestion
    // - reset typing attributes so that the previous line styles don't get
```

ios/utils/ParagraphAttributesUtils.mm
```diff
@@ -53,41 +41,25 @@ + (BOOL)handleBackspaceInRange:(NSRange)range
...
+        [TextInsertionUtils replaceText:text
+                                     at:range
+                   additionalAttributes:nullptr
+                                  input:typedInput
+                          withSelection:YES];
+        typedInput->textView.typingAttributes =
+            typedInput->defaultTypingAttributes;
+        [style add:NSMakeRange(range.location, 0)
+                withTyping:YES
+            withDirtyRange:YES];
```
szydlovsky: Will it still properly preserve current checkbox list checked status? Previous version did it intentionally.

ios/utils/ParagraphAttributesUtils.mm
```diff
@@ -186,14 +158,12 @@ + (BOOL)handleParagraphStylesMergeOnBackspace:(NSRange)range
       [conflictingStyles arrayByAddingObjectsFromArray:blockingStyles];
 
   for (NSNumber *style in allToBeRemoved) {
-    id<BaseStyleProtocol> styleClass = typedInput->stylesDict[style];
+    StyleBase *styleToRemove = typedInput->stylesDict[style];
 
-    // for ranges, we need to remove each occurence
```
szydlovsky, line 191: Let's keep this comment too

ios/utils/ZeroWidthSpaceUtils.mm
```diff
@@ -95,20 +87,29 @@ + (void)removeSpacesIfNeededinInput:(EnrichedTextInputView *)input {
   }
 }
 
+// Collects active inline (non-paragraph) meta-attributes from the style
+// dictionary so that ZWS characters carry the same meta-attributes that are
+// currently active in the typing attributes.
```
szydlovsky: Let's also add that it's only within the currently selected range!

ios/utils/ZeroWidthSpaceUtils.mm
```diff
@@ -154,15 +161,22 @@ + (void)addSpacesIfNeededinInput:(EnrichedTextInputView *)input {
...
-                     withSelection:NO];
+  if (lastParagraphRange.length == 0) {
+    BOOL anyZWSStylePresent = NO;
+    for (NSNumber *type in input->stylesDict) {
+      StyleBase *style = input->stylesDict[type];
+      if ([style needsZWS] && [style detect:lastRange]) {
+        anyZWSStylePresent = YES;
+        break;
+      }
+    }
```
szydlovsky: Nit; we're doing this search for present (or not present) ZWS style in some range a few times in this file. We could make it into a separate method to not copy the logic.

ios/utils/ParagraphAttributesUtils.mm
kacperzolkiewski: Yes it was wrong, I restore preserving checkbox checked status.

ios/utils/ParagraphAttributesUtils.mm
```diff
@@ -50,16 +52,25 @@ + (BOOL)handleBackspaceInRange:(NSRange)range
     for (NSNumber *type in typedInput->stylesDict) {
       StyleBase *style = typedInput->stylesDict[type];
       if ([style needsZWS] && [style detect:nonNewlineRange]) {
+        BOOL isCurrentlyChecked = [cbLStyle getCheckboxStateAt:range.location];
```
szydlovsky, line 55: We can move that check under `if (style == cbLStyle)` condition

kacperzolkiewski: We cannot because it will be after `replaceText` and the state will be not correct

ios/utils/ZeroWidthSpaceUtils.mm
```diff
@@ -154,15 +146,14 @@ + (void)addSpacesIfNeededinInput:(EnrichedTextInputView *)input {
...
-      ([ulStyle detectStyle:lastRange] || [olStyle detectStyle:lastRange] ||
-       [bqStyle detectStyle:lastRange] || [cbStyle detectStyle:lastRange] ||
-       [cbLStyle detectStyle:lastRange])) {
-    [TextInsertionUtils insertText:@"\u200B"
-                                at:lastRange.location
-              additionalAttributes:nullptr
-                             input:input
-                     withSelection:NO];
+  if (lastParagraphRange.length == 0) {
+    if ([self anyZWSStylePresentInRange:lastRange input:input]) {
```
szydlovsky: I think we can just combine these two conditions into a single `if` statement

ios/EnrichedTextInputView.mm
kacperzolkiewski: Fixed

ios/utils/ParagraphAttributesUtils.mm
szydlovsky: Okay

ios/EnrichedTextInputView.mm
```diff
@@ -370,9 +370,11 @@ - (void)updateProps:(Props::Shared const &)props
   if (newViewProps.htmlStyle.h1.bold != oldViewProps.htmlStyle.h1.bold) {
     [newConfig setH1Bold:newViewProps.htmlStyle.h1.bold];
 
-    // Update style blocks for bold
+    // Update style blocks and conflicts for bold
     newViewProps.htmlStyle.h1.bold ? [self addStyleBlock:H1 to:Bold]
                                    : [self removeStyleBlock:H1 from:Bold];
+    newViewProps.htmlStyle.h1.bold ? [self addStyleConflict:Bold to:H1]
+                                   : [self removeStyleConflict:Bold from:H1];
```
szydlovsky: It's a Nit but if we now want to call two methods, maybe let's just do;
```
if (newViewProps.htmlStyle.h1.bold) {
  ...
} else {
  ...
}
```
instead of this double ternary expression in all these headings?

ios/styles/CodeBlockStyle.mm
szydlovsky, line 1: Previously, backspacing the very first character of some line in the middle of a codeblock would just properly move the text to the previous line. Now, it just removes the styling in the middle. I think we want the previous solution back.

https://github.com/user-attachments/assets/c0f43c10-e053-4ec8-9f7f-a52aad986eda

ios/interfaces/StyleBase.mm
```diff
@@ -0,0 +1,256 @@
...
+               }];
+}
+
+// This method gets overridden
+- (void)applyStyling:(NSRange)range {
+}
+
+// Called during dirty range re-application to restore a style from a saved
+// StylePair
+- (void)reapplyAttributesFromStylePair:(StylePair *)pair {
```
szydlovsky: Quite a long name in comparison to all other methods in this file. Could we maybe shorten it to something like `reapplyFromStylePair`, as it's pretty obvious that if we reapply a style, we mean its metaattributes

ios/styles/CodeBlockStyle.mm
szydlovsky: Note; we can probably do just some simple backspace processing in the `CodeBlockStyle` and just put it before the general `[ParagraphUtils handleBackspace]` in the big check from `shouldReplaceText...` delegate method.

ios/styles/MentionStyle.mm
```diff
@@ -1,175 +1,138 @@
...
+  MentionParams *params = (MentionParams *)pair.styleValue;
+  if (params == nullptr) {
+    return;
+  }
+  [self applyMentionMeta:params range:range];
 }
 
-- (void)addTypingAttributes {
-  // no-op for mentions
+- (AttributeEntry *)getEntryIfPresent:(NSRange)range {
```
szydlovsky, line 77: Let's add a small comment that we don't want the mention to be extended, thus returning `nullptr` here.

ios/EnrichedTextInputView.mm
```diff
@@ -2025,32 +2000,32 @@ - (bool)textView:(UITextView *)textView
...
       [uStyle tryHandlingListShorcutInRange:range replacementText:text] ||
-      [oStyle handleBackspaceInRange:range replacementText:text] ||
       [oStyle tryHandlingListShorcutInRange:range replacementText:text] ||
-      [cbLStyle handleBackspaceInRange:range replacementText:text] ||
       [cbLStyle handleNewlinesInRange:range replacementText:text] ||
-      [bqStyle handleBackspaceInRange:range replacementText:text] ||
-      [cbStyle handleBackspaceInRange:range replacementText:text] ||
       [linkStyle handleLeadingLinkReplacement:range replacementText:text] ||
       [mentionStyle handleLeadingMentionReplacement:range
                                     replacementText:text] ||
```
szydlovsky: I just commented them both out on my end and these issues aren't there. There's a chance both these aren't needed anymore but need some more checking.

ios/styles/CodeBlockStyle.mm
szydlovsky: Turns out most today's mobile editors do it the way new impl changed it, so let's stick to that.

ios/styles/LinkStyle.mm
```diff
@@ -1,264 +1,226 @@
...
+  NSRange range = [pair.rangeValue rangeValue];
+  NSString *url = nullptr;
+  BOOL manual = YES;
+  id styleValue = pair.styleValue;
+  if ([styleValue isKindOfClass:[LinkData class]]) {
+    LinkData *linkData = (LinkData *)styleValue;
+    url = linkData.url;
+    manual = linkData.text == nullptr
+                 ? YES
+                 : ![linkData.text isEqualToString:linkData.url];
```
szydlovsky: That's quite a rough way to decide whether a link from style pair is manual or automatic. In this case, a manual link that had the exact same url as text will be converted to an automatic one. Can we make links hold both `url` and `manual` boolean flag as the attribute key's value?

ios/styles/LinkStyle.mm
```diff
@@ -1,264 +1,226 @@
...
+  }
+  if (url == nullptr) {
+    return;
+  }
+  [self applyLinkMetaForUrl:url manual:manual range:range];
 }
 
-- (void)addAttributes:(NSRange)range withTypingAttr:(BOOL)withTypingAttr {
-  // no-op for links
+- (AttributeEntry *)getEntryIfPresent:(NSRange)range {
```
szydlovsky, line 55: Would love a comment explaining we don't want to expand links too.

ios/styles/LinkStyle.mm
```diff
@@ -1,264 +1,226 @@
...
   if (range.length >= 1) {
     BOOL onlyLinks = [OccurenceUtils
         detectMultiple:@[ ManualLinkAttributeName, AutomaticLinkAttributeName ]
-             withInput:_input
+             withInput:self.input
                inRange:range
-         withCondition:^BOOL(id _Nullable value, NSRange range) {
-           return [self styleCondition:value range:range];
+         withCondition:^BOOL(id _Nullable value, NSRange r) {
+           return [self styleCondition:value range:r];
```
szydlovsky: Nit: maybe `subrange` instead of `r`

ios/styles/LinkStyle.mm
szydlovsky: And also I don't get why do we check for `LinkData`, afaik we never set it as a value for the link attribute key.

szydlovsky: The same manual check is in `InputParser` and should also be changed if we decide to keep the info about the link type

ios/styles/LinkStyle.mm
```diff
@@ -427,11 +342,11 @@ - (void)handleAutomaticLinks:(NSString *)word inRange:(NSRange)wordRange {
 
   // we don't recognize automatic links along manual ones
   __block BOOL manualLinkPresent = NO;
-  [_input->textView.textStorage
+  [self.input->textView.textStorage
       enumerateAttribute:ManualLinkAttributeName
                  inRange:wordRange
                  options:0
-              usingBlock:^(id value, NSRange range, BOOL *stop) {
+              usingBlock:^(id value, NSRange r, BOOL *stop) {
```
szydlovsky: Also let's change `r` to `subrange` for clarity

ios/styles/LinkStyle.mm
```diff
@@ -551,48 +468,33 @@ - (void)handleManualLinks:(NSString *)word inRange:(NSRange)wordRange {
...
-                                         range:newRange];
-    [_input->textView.textStorage addAttribute:NSStrikethroughColorAttributeName
-                                         value:[_input->config linkColor]
-                                         range:newRange];
-    if ([_input->config linkDecorationLine] == DecorationUnderline) {
-      [_input->textView.textStorage addAttribute:NSUnderlineStyleAttributeName
-                                           value:@(NSUnderlineStyleSingle)
-                                           range:newRange];
+    [self applyLinkMetaForUrl:manualLinkMinValue manual:YES range:newRange];
+    if (newRange.length > 0) {
```
szydlovsky: No need for the length check, `addDirtyRange` handles the zero length case.

ios/styles/LinkStyle.mm
```diff
@@ -605,46 +507,85 @@ - (BOOL)handleLeadingLinkReplacement:(NSRange)range
...
+  BOOL anyAutomatic = [OccurenceUtils any:AutomaticLinkAttributeName
+                                withInput:self.input
+                                  inRange:wordRange
+                            withCondition:^BOOL(id _Nullable value, NSRange r) {
+                              return [self styleCondition:value range:r];
+                            }];
+  BOOL anyManual = [OccurenceUtils any:ManualLinkAttributeName
+                             withInput:self.input
+                               inRange:wordRange
+                         withCondition:^BOOL(id _Nullable value, NSRange r) {
```
szydlovsky: Also here, `r` -> `subrange`

ios/EnrichedTextInputView.h
```diff
@@ -22,23 +23,30 @@ NS_ASSUME_NONNULL_BEGIN
...
 }
 - (CGSize)measureSize:(CGFloat)maxWidth;
 - (void)emitOnLinkDetectedEvent:(NSString *)text
                             url:(NSString *)url
                           range:(NSRange)range;
 - (void)emitOnMentionEvent:(NSString *)indicator text:(nullable NSString *)text;
 - (void)emitOnPasteImagesEvent:(NSArray<NSDictionary *> *)images;
+- (void)emitOnMentionDetectedEvent:(NSString *)text
+                         indicator:(NSString *)indicator
+                        attributes:(NSString *)attributes;
```
szydlovsky: I think we don't need it here, it's used only in the `EnrichedTextInputView`

ios/EnrichedTextInputView.mm
```diff
@@ -268,6 +261,10 @@ - (void)setupTextView {
   textView.delegate = self;
   textView.input = self;
   textView.layoutManager.input = self;
+  textView.textStorage.delegate = self;
+
+  textView.autoresizingMask =
+      UIViewAutoresizingFlexibleWidth | UIViewAutoresizingFlexibleHeight;
```
szydlovsky: Was this purposefully added by this PR? Maybe it's a remnant of some reverted PR?

kacperzolkiewski: Yes it was a remnant of some reverted PR, deleted

ios/styles/LinkStyle.mm
kacperzolkiewski: I added isManual to `LinkData`

ios/EnrichedTextInputView.mm
kacperzolkiewski: I deleted them

ios/styles/LinkStyle.mm
```diff
@@ -132,19 +125,47 @@ - (BOOL)any:(NSRange)range {
...
+                  isManual:NO
+                   inRange:range
+                      into:result];
+  return result;
+}
+
+- (void)collectPairsForKey:(NSString *)key
+                  isManual:(BOOL)isManual
+                   inRange:(NSRange)range
+                      into:(NSMutableArray<StylePair *> *)result {
```
szydlovsky: Can we simplify this method by not using `isManual` argument and instead just checking for:
`[key isEqualToString: ManualLinkAttributeName]`
?

ios/EnrichedTextInputView.mm
```diff
@@ -1428,13 +1429,15 @@ - (void)emitOnSubmitEdittingEvent {
 
 - (void)emitOnLinkDetectedEvent:(NSString *)text
                             url:(NSString *)url
+                       isManual:(BOOL)isManual
                           range:(NSRange)range {
```
szydlovsky: I think we could modify this method to only take `LinkData` and `range` as arguments and unpack anything that's needed from it.

ios/EnrichedTextInputView.mm
szydlovsky, line 1: Could you also add the check for `linkData` in the place that we decide whether we are on a new link or not? (the else if in line ~1119). We already compare `text`, `url` and `range` of the candidate link, let's add `linkData` there too.

Issue #477, opened 2026-03-10
[Android]: link turns into plain text after typing after it


pkaramon:
**Describe the bug**
After adding a link in the editor, typing immediately after it (e.g. pressing space) causes the link to become plain text.

**To Reproduce**
1. Launch example app
2. Add a link
3. Make sure cursor in on the right side of the link
4. Press space or any other character
5. Link turns into plain text
 

**Expected behavior**

**Screenshots**

https://github.com/user-attachments/assets/b9903087-4e00-40d3-86cd-13d00f36b3f6

**Device (please complete the following information):**
 - Device: Pixel 9
 - OS: Android 16
 - Version  API 36 "Baklava"

**Additional context**

After fixing update `link_not_extended.yaml` test to also run on android i.e. remove `ios-only` tag.

PR #485, opened 2026-03-13
test: add tests for recent bugs 3

pkaramon:
# Summary

+ added tests for PRs: #370, #421, #401, #359

## Test Plan
Execute new tests see if they pass:
```bash
yarn test:e2e ./.maestro/flows/{inline_styles_survive_block_toggle.yaml,link_not_extended.yaml,list_newline_insertion.yaml,ordered_list_renumbering.yaml}
```

## Screenshots / Videos

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

### DIFF
--- a/.maestro/flows/list_newline_insertion.yaml
+++ b/.maestro/flows/list_newline_insertion.yaml
@@ -20,10 +20,15 @@ appId: swmansion.enriched.example
         - tapOn:
             id: "editor-input"
             point: "50%,50%"
+# iOS places the cursor on word boundaries by default. Precisely positioning
+# the cursor would require a press-and-drag gesture, which Maestro does not
+# support. As a workaround we type two words, tap between them (snapping to
+# the word boundary/space), then delete the space so we end up with a
+# continuous word and the cursor in the middle.
 - runFlow:
     when:
         platform: ios
-    commands: 
+    commands:
         - inputText: "XXXXXXXXXXX XXXXXXXXXXXXX"
         - tapOn:
             id: "editor-input"


.maestro/flows/link_not_extended.yaml
```diff
@@ -0,0 +1,34 @@
+appId: swmansion.enriched.example
+tags:
+    - ios-only
+---
```
pkaramon, line 4: Reason: #477. Added a note to enable this test also on android after it's fixed.

.maestro/flows/list_newline_insertion.yaml
```diff
@@ -0,0 +1,39 @@
...
+            point: "50%,50%"
+- runFlow:
+    when:
+        platform: ios
+    commands: 
+        - inputText: "XXXXXXXXXXX XXXXXXXXXXXXX"
+        - tapOn:
+            id: "editor-input"
+            point: "50%,50%"
+        - pressKey: Backspace
```
pkaramon, line 36: iOS handles cursor placement differently. It places the cursor by default on word boundaries, if we wanted to place the cursor exactly we would need to press down and drag the cursor, but this movement is currently unsupported in Maestro. So we write two words, tap between, remove the space and then we have a continuous word with cursor in the middle of it

exploIF: It would be the best to add this comment directly inside the test, so it's much easier to understand it in the future

pkaramon: Good idea

PR #489, opened 2026-03-18
fix(android): block mention events when cursor is inside a finalized mention span

Jianlong-Nie:
## Problem
Fixes: #490 

On Android, after a mention is finalized (user selects from the suggestion list), typing new characters after the mention continues to trigger the mention suggestion list.

**Root cause:** `afterTextChangedMentions` uses a `previousWord` fallback — when the cursor moves to a new word after the mention, it looks back and finds `@username` as the previous word, which matches `mentionRegex`, causing `onMention` to fire again even though the mention span is already resolved.

## Fix

After determining the candidate range `[finalStart, finalEnd]`, scan for `EnrichedInputMentionSpan` across the full range:

- **Span text matches buffer** → mention is intact, call `endMention()` and return. This blocks the spurious event when the cursor is in a word adjacent to a finalized mention span (mirrors iOS `conflictingStyles` behaviour).
- **Span text has diverged** (user edited inside it) → remove the stale span and record `mentionStart = spanStart` so `setMentionSpan` can replace the correct range when the user picks a new mention.

## Changes

- `android/…/styles/ParametrizedStyles.kt` — reworked `afterTextChangedMentions` span-check logic.

before：

https://github.com/user-attachments/assets/00cd2d7a-6fb4-498d-bc41-5db36385af26

after：

https://github.com/user-attachments/assets/14cffa5c-5797-4530-a08b-70f83b84773c


kacperzolkiewski:
Hi @Jianlong-Nie, 
In your second video, it looks like it’s not working correctly. Deleting characters after @ should still emit the event just like it does in Slack.

Jianlong-Nie:
> In your second video, it looks like it’s not working correctly. Deleting characters after @ should still emit the event just like it does in Slack.

If you take a closer look, it does trigger — in the video, it fires when deleting the fourth or fifth character. I’ve added some throttling here, so it’s not triggered immediately on every single change. @kacperzolkiewski

kacperzolkiewski:
@Jianlong-Nie Thank you for your contribution!
We’d like community fixes to follow the issue → PR flow, so could you please create an issue describing the problem this PR addresses and link it here?
I had to run it locally and test it myself to fully understand the issue, so a clearer description would really help reviewers.

Jianlong-Nie:
Linking the issue as requested: fixes #490

kacperzolkiewski (approved):
Looks and works fine. Nice work👏

PR #491, opened 2026-03-18
feat(web): add basic setup

pkaramon:
# Summary

Explain the **motivation** for making this change: here are some points to help you:

- Basic web setup for the library and example app, intended for internal use only for now
- **Currently the code related to web is not exported to npm**
-  Added web entry + minimal web `EnrichedTextInput` stub to unblock web integration work.
-  Wired example-web app to the `EnrichedTextInput` for the web
- `index.tsx` entrypoint for android/ios, `index.web.tsx` entrypoint for the web

## Test Plan

- Run example-web app, make sure a textarea with "Hello from web!" appears.
- Check that example app still builds for android and ios

## Screenshots / Videos

image
## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅    |
| Web | ✅ |

### DIFF
--- a/src/index.web.tsx
+++ b/src/index.web.tsx
@@ -9,6 +9,7 @@ export type {
   OnChangeSelectionEvent,
   OnKeyPressEvent,
   OnPasteImagesEvent,
+  OnSubmitEditing,
   HtmlStyle,
   MentionStyleProperties,
   FocusEvent,


src/web/EnrichedTextInput.tsx
```diff
@@ -0,0 +1,50 @@
+import { useImperativeHandle } from 'react';
+import type {
+  BaseEnrichedTextInputProps,
+  EnrichedTextInputInstance,
+} from '../types';
+
+export interface EnrichedTextInputProps extends BaseEnrichedTextInputProps {}
```
pkaramon: I will try to keep the types as close as possible between native and web, but I think we will have to make at least so tiny modifications for example the `style?: ViewStyle | TextStyle` will need changing on web to `CSSProperties`.

package.json
```diff
@@ -20,6 +25,14 @@
...
     "*.podspec",
     "react-native.config.js",
+    "!src/web",
+    "!src/index.tsx",
+    "!lib/module/web",
+    "!lib/module/index.js",
+    "!lib/module/index.js.map",
+    "!lib/typescript/src/web",
+    "!lib/typescript/src/index.d.ts",
+    "!lib/typescript/src/index.d.ts.map",
```
pkaramon: I think it would be a good idea to keep the web code hidden until it's at least semi-functional. @exploIF @kacperzolkiewski curious what do you think?

kacperzolkiewski: Sounds good to me! 
But it is a little inconsistent to me that we point to excluded files here:
```
  "exports": {
    ".": {
      "react-native": {
        "types": "./lib/typescript/src/index.native.d.ts",
        "default": "./lib/module/index.native.js"
      },
      "types": "./lib/typescript/src/index.d.ts",
      "default": "./lib/module/index.js"
    }, 
```

src/web/EnrichedTextInput.tsx
exploIF: We can have only definition of types and they have to be exactly the same for both plaforms (even if some props on one platform will be no-op)

src/native/EnrichedTextInput.tsx
```diff
@@ -10,141 +9,30 @@ import { useCallback } from 'react';
...
-  /**
-   * If true, external HTML (e.g. from Google Docs, Word, web pages) will be
-   * normalized through the HTML normalizer before being applied.
-   * This converts arbitrary HTML into the canonical tag subset that the enriched
-   * parser understands.
-   * Disabled by default.
-   */
-  useHtmlNormalizer?: boolean;
-}
+export interface EnrichedTextInputProps extends BaseEnrichedTextInputProps {}
```
exploIF: I think it's redundant. We should have single types definition (but some props can be no-op)

src/index.tsx
```diff
@@ -1,4 +1,5 @@
-export * from './EnrichedTextInput';
```
exploIF, line 1: Until we don't have fully functional web, I would keep mobile implementation in `index` file and web implementation inside `index.web`. Once web is ready to be published we will rename it to `index` and `index.native`. I think it's more safe, I remember that some setups might fail if there is no regular `index` file.

src/web/EnrichedTextInput.tsx
pkaramon: I've unified the types across platforms.

package.json
pkaramon: It's now all pointing ot index.js or index.d.ts, after I've switched to index.ts -> native, index.web.ts -> web.

src/index.tsx
pkaramon: Switched to `index.ts` -> native, `index.web.ts` -> web.

PR #496, opened 2026-03-24
feat(web): added tiptap foundation

pkaramon:
# Summary

Adds a web implementation of `EnrichedTextInput` using Tiptap as the underlying editor. Also I've built out `example-web` to demonstrate existing features. Currently the only supported tag is `p` the simple paragraph.

**Props implemented on web:**
- `defaultValue`, `autoFocus`, `editable`, `placeholder`, `autoCapitalize`, `scrollEnabled`
- `onFocus`, `onBlur`, `onChangeText`, `onChangeHtml`, `onChangeSelection`, `onKeyPress`

**Imperative handle methods implemented:**
- `focus`, `blur`, `setValue`, `setSelection`, `getHTML`

**Other:**
- Added `adaptWebToNativeEvent` - wraps DOM/Tiptap events into `NativeSyntheticEvent<T>`
- Extracted `ENRICHED_TEXT_INPUT_DEFAULT_PROPS` to unify default values across native and web implementations

## Test Plan

+ Launch `yarn example-web dev`
+ Verify functionality like focus, blur, placeholder, works and console contains logs about the events.
+ Check if the output html is the same as on android/ios.

## Screenshots / Videos

image

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| Web  |    ✅     |

## Checklist

- [ ] E2E tests are passing

### DIFF
--- a/package.json
+++ b/package.json
@@ -143,6 +143,7 @@
     ],
     "testPathIgnorePatterns": [
       "/node_modules/",
+      "<rootDir>/.playwright/",
       "<rootDir>/apps/example-web/"
     ]
   },


src/web/EnrichedTextInput.css
```diff
@@ -0,0 +1,12 @@
...
+  content: attr(data-placeholder);
+  float: left;
+  height: 0;
+  pointer-events: none;
+}
+
+.eti-editor {
+  min-height: 150px;
+  max-height: 500px;
+}
```
pkaramon, line 12: This will need changing after deciding what to do with the style prop on web.

package.json
```diff
@@ -105,16 +112,47 @@
...
   },
   "peerDependencies": {
+    "@tiptap/core": ">=3.20.4",
+    "@tiptap/extension-document": ">=3.20.4",
+    "@tiptap/extension-paragraph": ">=3.20.4",
+    "@tiptap/extension-text": ">=3.20.4",
+    "@tiptap/extensions": ">=3.20.4",
+    "@tiptap/pm": ">=3.20.4",
+    "@tiptap/react": ">=3.20.4",
     "react": "*",
```
pkaramon, line 130: Should we go with `dependencies` or  optional `peerDependencies` for Tiptap packages?

I went with optional peerDependencies for all @tiptap/* packages because:
- Native-only users don’t need Tiptap at all, so this avoids pulling in stuff they’ll never use
- If someone already has Tiptap in their web project, putting it in dependencies risks ending up with duplicate installs and possibly breaking something

The downside is DX isn’t as smooth. If you’re using this on the web, you have to explicitly install the required @tiptap/* packages yourself.

I'm open to switching to `dependencies` though.

apps/example-web/package.json
```diff
@@ -10,6 +10,13 @@
     "lint": "eslint . --ext ts,tsx"
   },
   "dependencies": {
+    "@tiptap/core": "^3.20.4",
+    "@tiptap/extension-document": "^3.20.4",
+    "@tiptap/extension-paragraph": "^3.20.4",
+    "@tiptap/extension-text": "^3.20.4",
+    "@tiptap/extensions": "^3.20.4",
+    "@tiptap/pm": "^3.20.4",
+    "@tiptap/react": "^3.20.4",
```
exploIF: Apps that are using react-native-enriched shouldn't have to specify any additional dependencies. These are already added to a lib package.json, so I guess they are redundant here

src/web/EnrichedTextInput.tsx
```diff
@@ -1,21 +1,103 @@
...
+  onChangeText,
+  onChangeHtml,
 }: EnrichedTextInputProps) => {
+  const editor = useEditor({
+    extensions: [
+      Document,
+      Paragraph,
+      Text,
+      Placeholder.configure({
+        placeholder: placeholder ?? '',
```
exploIF: Maybe we can assign empty string during props destructuring like `placeholder = ''`, and then use it here like 
```suggestion
        placeholder,
```

apps/example-web/src/App.tsx
```diff
@@ -1,10 +1,106 @@
...
+      <h1 className="app-title">Enriched Text Input</h1>
+
+      <div className="editor-wrapper" onClick={() => ref.current?.focus()}>
+        <div className="editor-content">
+          <EnrichedTextInput
+            ref={ref}
+            placeholder="Type something"
+            autoFocus={true}
+            editable={true}
+            scrollEnabled={true}
```
exploIF:
```suggestion
            autoFocus
            editable
            scrollEnabled
```

src/web/EnrichedTextInput.tsx
```diff
@@ -1,21 +1,103 @@
...
+        showOnlyWhenEditable: true,
+      }),
+    ],
+    content:
+      defaultValue != null ? prepareHtmlForTiptap(defaultValue) : defaultValue,
+    editable: editable,
+    autofocus: autoFocus,
```
exploIF:
```suggestion
    editable,
    autofocus,
```

src/web/EnrichedTextInput.tsx
```diff
@@ -1,21 +1,103 @@
...
+        })
+      );
+    },
+    editorProps: {
+      handleKeyPress: (_, event) => {
+        onKeyPress?.(adaptWebToNativeEvent(event, { key: event.key }));
+        return false;
+      },
+      attributes: {
+        autoCapitalize: autoCapitalize,
```
exploIF:
```suggestion
        autoCapitalize,
```

src/web/EnrichedTextInput.tsx
```diff
@@ -1,21 +1,103 @@
...
+      Document,
+      Paragraph,
+      Text,
+      Placeholder.configure({
+        placeholder: placeholder ?? '',
+        showOnlyWhenEditable: true,
+      }),
+    ],
+    content:
+      defaultValue != null ? prepareHtmlForTiptap(defaultValue) : defaultValue,
```
exploIF: On mobile, whenever default value changes we update it. We should follow the same approach on web. 

I think it would be the best to move `prepareHtmlForTiptap` outside the `useEditor`, calculate it on every `defaultValue` change, and pass calculated value to `useEditor` deps, like

```tsx
useEditor(..., [resultOfPrepareHtmlForTiptap]) // find better naming pls ;)
```

src/web/EnrichedTextInput.tsx
```diff
@@ -1,21 +1,103 @@
...
-      getHTML: () => Promise.resolve(''),
+      focus: () => editor.commands.focus(),
+      blur: () => editor.commands.blur(),
+      setValue: (value: string) =>
+        editor.commands.setContent(prepareHtmlForTiptap(value)),
+      setSelection: (start, end) => {
+        editor
+          .chain()
+          .focus()
+          .setTextSelection({ from: start + 1, to: end + 1 })
```
exploIF: Is tiptap starting from `1`? If so, could you add a comment? This is pretty unexpected

package.json
exploIF: Personally, I would prefer direct dependencies, but let's discuss that with rest of the team

apps/example-web/package.json
pkaramon: This depends on whether tiptap packages will be peerDependencies or direct dependencies. Now they are peerDependencies so they example web app using them must list them explicitly, if we go with direct dependencies then you're right those will be redundant.

src/web/EnrichedTextInput.tsx
pkaramon: I've applied the change to `editable`, but I can't do the same for `autofocus` because the casing of the "f" is different.

package.json
pkaramon: I've went with direct dependencies.

apps/example-web/package.json
pkaramon: I've removed those since we've chosen to use direct dependencies.

src/web/EnrichedTextInput.tsx
pkaramon: So initially I thought tiptap has 2 special indexes 0: beginning of the document, and similar for end of the document and otherwise works the same as indexes on android/ios.  But it's actually a bit more complex. I've added `positionMapping.ts` which handles the mapping and explains differences between how tiptap understands positions and how positions are implemented in enriched on android/ios.

src/web/positionMapping.ts
```diff
@@ -0,0 +1,66 @@
...
+    const mid = low + Math.floor((high - low) / 2);
+    if (tiptapPosToNativePos(doc, mid) >= nativePos) {
+      result = mid; // mid is a valid candidate; try to find an earlier one
+      high = mid - 1;
+    } else {
+      low = mid + 1; // mid is too small, search right
+    }
+  }
+  return result;
+}
```
pkaramon, line 66: Base algorithm taken from https://www.geeksforgeeks.org/dsa/implement-lower-bound/.
I've tried different approaches based on walking the document tree but I've encountered difficult edge cases. This seems to be robust, and after lots of trying I couldn't find an edge case where this would break.

kacperzolkiewski: I also tested this with `ref.setSelection()` and it seems to work properly

CONTRIBUTING.md
```diff
@@ -190,6 +212,7 @@ The `package.json` file contains various scripts for common tasks:
 - `yarn test:e2e`: run E2E tests on iOS and Android sequentially.
```
exploIF: Ideally we should rename this script to: `yarn test:e2e:mobile` and have a separate `yarn:test:e2e` which runs all tests (mobile + web)

package.json
```diff
@@ -85,6 +86,15 @@
...
     "registry": "https://registry.npmjs.org/"
   },
+  "dependencies": {
+    "@tiptap/core": "^3.20.4",
+    "@tiptap/extension-document": "^3.20.4",
+    "@tiptap/extension-paragraph": "^3.20.4",
+    "@tiptap/extension-text": "^3.20.4",
+    "@tiptap/extensions": "^3.20.4",
+    "@tiptap/pm": "^3.20.4",
+    "@tiptap/react": "^3.20.4"
```
exploIF: Can we pin those versions? I would prefer to have full control over dependencies versions

.playwright/tests/setSelection.spec.tsx
exploIF, line 1: Our mobile tests `.meastro` live in a root directory, but web related tests are located inside `example-web`. Is it possible to make it consistent or it's not with current setup?

CONTRIBUTING.md
pkaramon: I've added that

package.json
pkaramon: I've pinned them

.playwright/tests/setSelection.spec.tsx
pkaramon: I've moved web tests to .playwright, similar to how we have .maestro directory.

PR #507, opened 2026-04-01
docs: add supported tags, conflicting/blocking logic

kacperzolkiewski:
# Summary

Update readme and add **Supported Tags** section with conflicting and blocking logic

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,7 +143,7 @@ Summary of what happens here:
 
 ## Supported Tags
 
-`react-native-enriched` uses the following HTML tags in its output and accepts them as input.
+`react-native-enriched` uses both standard and custom HTML tags in its output and accepts them as input.
 
 Not all styles can be combined freely. There are two kinds of restrictions:
 


README.md
```diff
@@ -25,6 +25,7 @@ We can help you build your next dream product –
 - [Installation](#installation)
 - [Usage](#usage)
 - [Non Parametrized Styles](#non-parametrized-styles)
+- [Supported Tags](#supported-tags)
```
szydlovsky, line 27: Can we move it right after the `Usage` section? It's a bit weird that its between non parametrized and custom styles

README.md
```diff
@@ -140,6 +141,50 @@ Summary of what happens here:
 - `isBlocking`: Indicates if the style is blocked by another active style (disable the button).
 - `isConflicting`: Indicates if the style is in conflict with another active style.
 
+## Supported Tags
+
+`react-native-enriched` uses the following HTML tags in its output (via [onChangeHtml](docs/API_REFERENCE.md#onchangehtml)) and accepts them as input (via [setValue](docs/API_REFERENCE.md#setValue)).
```
exploIF: This is not precise, for the output you can also use `getValue`, for the input `defaultValue`

README.md
```diff
@@ -140,6 +141,50 @@ Summary of what happens here:
...
+| ------------- | ------------------------------------------- | ---------------------------- | ---------------------- |
+| Bold          | `<b>`                                       | --                           | `<codeblock>`          |
+| Italic        | `<i>`                                       | --                           | `<codeblock>`          |
+| Underline     | `<u>`                                       | --                           | `<codeblock>`          |
+| Strikethrough | `<s>`                                       | --                           | `<codeblock>`          |
+| Inline code   | `<code>`                                    | `<a>`, `<mention>`           | `<codeblock>`, `<img>` |
+| Link          | `<a href="...">`                            | `<code>`, `<a>`, `<mention>` | `<codeblock>`, `<img>` |
+| Mention       | `<mention text="..." indicator="...">`      | `<code>`, `<a>`              | `<codeblock>`, `<img>` |
+| Image         | `<img src="..." width="..." height="..."/>` | `<a>`, `<mention>`           | `<code>`               |
```
exploIF:
```suggestion
| Link          | `<a>`                            | `<code>`, `<a>`, `<mention>` | `<codeblock>`, `<img>` |
| Mention       | `<mention>`      | `<code>`, `<a>`              | `<codeblock>`, `<img>` |
| Image         | `<img>` | `<a>`, `<mention>`           | `<code>`               |
```

README.md
```diff
@@ -140,6 +141,50 @@ Summary of what happens here:
...
+
+### Paragraph tags
+
+Only one paragraph-level style can be active per paragraph - all paragraph styles conflict with each other.
+
+| Style          | HTML tag                    | Inner content tag       | Conflicts with                                                                                                                                                        | Blocked by |
```
exploIF, line 177: If we want to use `Inner content tag` naming, this should be either described first or we should have a glossary

README.md
kacperzolkiewski: I deleted mentioned `setValue`  and `onchangehtml` from this section.

README.md
kacperzolkiewski: Yes, you're right. I described it first

README.md
```diff
@@ -140,6 +141,55 @@ Summary of what happens here:
 - `isBlocking`: Indicates if the style is blocked by another active style (disable the button).
 - `isConflicting`: Indicates if the style is in conflict with another active style.
 
+## Supported Tags
+
+`react-native-enriched` uses the following HTML tags in its output and accepts them as input.
```
szydlovsky: This would make sense if we actually listed all the tags directly after that sentence. Here it feels a bit weird with the next one. Maybe something like this?:

```suggestion
`react-native-enriched` uses both standard and custom HTML tags in its output and accepts them as input.
```

PR #509, opened 2026-04-02
Correct animated WebP frame timing on iOS

fendent:
Working on an editor where I'm inserting animated images. When using animated WebP, the frame timing of the image is off and causes the image to run way too fast.

# Summary

`delayCentisecondsForImageAtIndex` in `ImageExtension.mm` only reads frame delays from `kCGImagePropertyGIFDictionary`. Animated WebP images store their frame delays under `kCGImagePropertyWebPDictionary` instead, so every frame falls through to the hardcoded default of 1 centisecond.

The fix checks `kCGImagePropertyGIFDictionary` first, then falls back to `kCGImagePropertyWebPDictionary` with the corresponding `kCGImagePropertyWebPUnclampedDelayTime` / `kCGImagePropertyWebPDelayTime` keys.

Only impacts `ios/extensions/ImageExtension.mm`

## Test Plan

1. Insert an animated WebP image into an `EnrichedTextInput` via `setImage`
2. Observe the animation playback speed
3. **Before:** animation loops play far too fast (~100x intended speed)
4. **After:** animation loops play at the correct speed, matching how the same image renders in a standard `UIImageView` or `Image` component
5. Verify animated GIFs still play at the correct speed

## Screenshots / Videos

Video unavailable at the moment. 

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    n/a     |

Android is unaffected by this change. Cannot validate whether this bug exists under Android as well.

## Checklist

- [x] E2E tests are passing

### DIFF
--- a/ios/extensions/ImageExtension.h
+++ b/ios/extensions/ImageExtension.h
@@ -9,7 +9,7 @@
 @interface UIImage (ImageExtension)
 
 /*
-        UIImage *animation = [UIImage animatedImageWithAnimatedGIFData:theData];
+        UIImage *animation = [UIImage animatedImageWithData:theData];
 
     I interpret `theData` as an animated image (GIF or WebP).  I create an
    animated `UIImage` using the source images and frame delays from the data.
@@ -28,7 +28,6 @@
    contain frame 0 3/3 = 1 time, then frame 1 9/3 = 3 times, then frame 2
    15/3 = 5 times.  I set `animation.duration` to (3+9+15)/100 = 0.27 seconds.
 */
-+ (UIImage *_Nullable)animatedImageWithAnimatedGIFData:
-    (NSData *_Nonnull)theData;
++ (UIImage *_Nullable)animatedImageWithData:(NSData *_Nonnull)theData;
 
 @end
--- a/ios/extensions/ImageExtension.mm
+++ b/ios/extensions/ImageExtension.mm
@@ -120,8 +120,7 @@ static void releaseImages(const std::vector<CGImageRef> &images) {
   }
 }
 
-static UIImage *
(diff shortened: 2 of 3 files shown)


fendent:
Going over this a bit more, should I address the whole functions including GIF in the name even though they support animated WebP thing? Just leave it be?

Examples:
- `animatedImageWithAnimatedGIFImageSource` (private)
- `animatedImageWithAnimatedGIFReleasingImageSource` (private)
- `animatedImageWithAnimatedGIFData` (public)

Obviously changing `animatedImageWithAnimatedGIFData` would be a breaking change and changing the other two would break up the naming pattern so I figure let sleeping dogs lie?

kacperzolkiewski:
Hi @fendent,
Thank you for your work!

Regarding the naming, I actually think it’s totally worth doing the refactor now. Since the code handles `WebP` too, keeping `GIF` in the name is a bit misleading.

Don’t worry about the breaking change, it won't affect the library consumers, so we’re safe to update the public function along with the private ones. My proposal of renaming:

- `animatedImageWithAnimatedGIFImageSource` -> `animatedImageWithImageSource`
- `animatedImageWithAnimatedGIFReleasingImageSource` -> `animatedImageWithReleasingSource`
- `animatedImageWithAnimatedGIFData` -> `animatedImageWithData`

fendent:
>     * `animatedImageWithAnimatedGIFImageSource` -> `animatedImageWithImageSource`
>     * `animatedImageWithAnimatedGIFReleasingImageSource` -> `animatedImageWithReleasingSource`
>     * `animatedImageWithAnimatedGIFData` -> `animatedImageWithData`

Sounds good! I updated with the suggested method names.

PR #510, opened 2026-04-02
feat(web): enriched input style

pkaramon:
# Summary
+ This PR introduces `EnrichedInputStyle`, which is a subset of `ViewStyle | TextStyle`. 
+  Currently this new style type is only used on the web, and not on native. Since web version isn't pushed to npm this PR does not introduce breaking changes. 
+ Added a converter `EnrichedInputStyle` -> `CSSProperties` as well as tests for it.
+ Small changes related to `tsconfig`, which were needed because `index.web.tsx` now is a bit different than `index.tsx`.

## Test Plan
+ Run `yarn example-web dev`
+ Play around with `const enrichedInputStyle: EnrichedInputStyle` in `apps/example-web/src/App.tsx`

## Screenshots / Videos

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ❌     |
| Web |    ✅    |

## Checklist

- [ ] E2E tests are passing

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,7 @@ We can help you build your next dream product –
 - [Prerequisites](#prerequisites)
 - [Installation](#installation)
 - [Usage](#usage)
+- [Supported Tags](#supported-tags)
 - [Non Parametrized Styles](#non-parametrized-styles)
 - [Links](#links)
 - [Mentions](#mentions)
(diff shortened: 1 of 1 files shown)


src/web/enrichedInputStyleToCSSProperties.ts
```diff
@@ -0,0 +1,219 @@
...
+  if (value == null) return undefined;
+  if (typeof value === 'number') return `${value}px`;
+  if (typeof value === 'string') return value;
+  return undefined;
+}
+
+function toColor(value?: ColorValue): string | undefined {
+  if (typeof value === 'string') return value;
+  return undefined;
+}
```
kacperzolkiewski: Should't we handle also color values as ints here?
https://reactnative.dev/docs/colors#color-ints

kacperzolkiewski:
`EnrichedInputStyle` works correct on `EnrichedTextInput` but is actual nested DOM structure on Web intended? Currently, applying `translateX` via `enrichedInputStyle` only transforms the inner `EnrichedTextInput`, while the editor-wrapper and editor-content divs remain static?
image

apps/example-web/src/App.tsx
```diff
@@ -105,4 +107,9 @@ function App() {
   );
 }
 
+const enrichedInputStyle: EnrichedInputStyle = {
```
exploIF, line 106: Could you replace `editor-wrapper` and `editor-content` div wrappers with proper styling using `enrichedInputStyle`? This would be good check to ensure that base styles are working properly

src/web/__tests__/enrichedInputStyleToCSSProperties.test.ts
```diff
@@ -0,0 +1,963 @@
...
+    expect(convert({ marginVertical: 8 })).not.toHaveProperty('marginVertical');
+  });
+});
+
+describe('paddingHorizontal / paddingVertical shorthands', () => {
+  const cases: Array<{
+    description: string;
+    input: EnrichedInputStyle;
+    expected: CSSProperties;
+  }> = [
```
exploIF: I would move this type out of the `describe` and reuse across test cases, as it's redefined couple times

src/web/enrichedInputStyleToCSSProperties.ts
```diff
@@ -0,0 +1,219 @@
...
+        : `matrix(${item.matrix.join(', ')})`;
+    }
+    return null;
+  });
+
+  const css = parts.filter(Boolean).join(' ');
+  return css || undefined;
+}
+
+export function enrichedInputStyleToCSSProperties(
```
exploIF, line 13: I wonder if there is a way to type this function the way that whenever we add new style to `EnrichedInputStyle`, it has to be updated in order to handle all properties

src/web/EnrichedTextInput.tsx
```diff
@@ -129,9 +131,13 @@ export const EnrichedTextInput = ({
...
   );
 
-  const editorStyle: CSSProperties = {
-    overflowY: scrollEnabled ? 'auto' : 'hidden',
-  };
+  const editorStyle: CSSProperties = useMemo(() => {
+    const base = style ? enrichedInputStyleToCSSProperties(style) : {};
+    return {
+      ...base,
+      overflowY: scrollEnabled ? 'auto' : 'hidden',
```
exploIF: I think this could be handled directly inside `enrichedInputStyleToCSSProperties`, so we have only one place of creating styles

apps/example-web/src/App.tsx
pkaramon: Yup, I've removed all the wrappers around `EnrichedTextInput` and now all the styling is applied via `EnrichedInputStyle`. 

@kacperzolkiewski

src/web/__tests__/enrichedInputStyleToCSSProperties.test.ts
pkaramon: Extracted it out and reused across test cases.

src/web/EnrichedTextInput.tsx
pkaramon: I've added `extraOptions` parameter to the converter function, so now it handles the `scrollEnabled` prop as well.

src/web/enrichedInputStyleToCSSProperties.ts
pkaramon: I've added the handling of ints as well. I think that this is a very rare case, because typescript complains if one tries to use a number as a color value, but it works on android/ios so it should work on web as well.

src/web/enrichedInputStyleToCSSProperties.ts
exploIF: As discussed, let's leave it as it is for now, we'll come back to this topic later

PR #516, opened 2026-04-06
feat: add insertText command

fendent:
# Summary

Adds a new `insertText` command that inserts plain text into the input without replacing existing rich content (images, mentions, etc).

Accepts optional `start` and `end` position parameters (visible indices, skipping ZWSPs). When omitted or set to -1, inserts at the current selection/cursor position.

`setValue` replaces the entire content, and `setImage` replaces the selection with an image, but there's no way to programmatically insert plain text at a position without destroying rich content.

This is intentionally plain-text-only and complements #409 (`insertValue`), which targets rich/HTML content insertion.

```typescript
inputRef.current.insertText(' ');           // insert at cursor
inputRef.current.insertText('hello', 5);    // insert at position 5
inputRef.current.insertText('hi', 2, 5);    // replace range 2-5
```

## Test Plan

1. Insert an inline image via `setImage`
2. Call `insertText(' ')` — a space should appear after the image without destroying it
3. Call `insertText('hello', 0)` — text should be inserted at position 0
4. Call `insertText('x', 2, 5)` — should replace characters 2-5 with 'x'
5. Verify cursor moves to the end of inserted text in all cases

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist

- [ ] E2E tests are passing

### DIFF
--- a/android/src/main/java/com/swmansion/enriched/textinput/EnrichedTextInputView.kt
+++ b/android/src/main/java/com/swmansion/enriched/textinput/EnrichedTextInputView.kt
@@ -426,6 +426,28 @@ class EnrichedTextInputView :
     layoutManager.invalidateLayout()
   }
 
+  fun insertText(
+    text: String,
+    start: Int,
+    end: Int,
+  ) {
+    val editable = this.text ?: return
+    val replaceStart: Int
+    val replaceEnd: Int
+
+    if (start < 0) {
+      // Use current selection
+      replaceStart = selectionStart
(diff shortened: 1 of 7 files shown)


kacperzolkiewski:
Hi @fendent, 
The `insertValue` introduced in #409 already supports inserting plain text, so the `insertText` implementation appears to be redundant.

szydlovsky:
As explained above.

PR #519, opened 2026-04-06
feat: add autoCorrect and spellCheck props

fendent:
# Summary

Adds `autoCorrect` and `spellCheck` boolean props to control iOS `UITextAutocorrectionType` and `UITextSpellCheckingType`.

Currently there's no way to disable autocorrect or spell checking on the input. These props follow the same pattern as the existing `autoCapitalize` prop.

```tsx
<EnrichedTextInput
  autoCorrect={false}
  spellCheck={false}
/>
```

When not specified, both default to the system behavior (enabled).

## Test Plan

1. Set `autoCorrect={false}` — verify iOS does not suggest autocorrect replacements
2. Set `spellCheck={false}` — verify iOS does not underline misspelled words
3. Omit both props — verify default system behavior (both enabled)
4. Toggle props at runtime — verify behavior updates

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist

- [ ] E2E tests are passing

### DIFF
--- a/ios/EnrichedTextInputView.mm
+++ b/ios/EnrichedTextInputView.mm
@@ -897,6 +897,20 @@ Class<RCTComponentViewProtocol> EnrichedTextInputViewCls(void) {
     }
   }
 
+  // autoCorrect
+  if (newViewProps.autoCorrect != oldViewProps.autoCorrect) {
+    textView.autocorrectionType = newViewProps.autoCorrect
+                                      ? UITextAutocorrectionTypeYes
+                                      : UITextAutocorrectionTypeNo;
+  }
+
+  // spellCheck
+  if (newViewProps.spellCheck != oldViewProps.spellCheck) {
+    textView.spellCheckingType = newViewProps.spellCheck
+                                     ? UITextSpellCheckingTypeYes
+                                     : UITextSpellCheckingTypeNo;
+  }
+
   // isOnChangeHtmlSet
   _emitHtml = newViewProps.isOnChangeHtmlSet;
 
--- a/src/native/EnrichedTextInput.tsx
+++ b/src/native/EnrichedTextInput.tsx
@@ -58,6 +58,8 @@ export const EnrichedTextInput = ({
   selectionColor,
   style,
   autoCapitalize = ENRICHED_TEXT_INPUT_DEFAULT_PROPS.autoCapitalize,
+  autoCorrect,
+  spellCheck,
   htmlStyle = ENRICHED_TEXT_INPUT_DEFAULT_PROPS.htmlStyle,
   linkRegex: _linkRegex,
   onFocus,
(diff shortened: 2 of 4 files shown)


src/native/EnrichedTextInput.tsx
```diff
@@ -58,6 +58,8 @@ export const EnrichedTextInput = ({
   selectionColor,
   style,
   autoCapitalize = ENRICHED_TEXT_INPUT_DEFAULT_PROPS.autoCapitalize,
+  autoCorrect,
+  spellCheck,
```
exploIF: I would explicitly mark them as enabled by default 
```suggestion
  autoCorrect = true,
  spellCheck = true,
```

exploIF (commented):
For the `autoCorrect`, it would be the best to introduce this prop on Android too, so it's consistent with [react native TextInput](https://reactnative.dev/docs/textinput#autocorrect)

nickmccomb:
Any updates on this?

src/native/EnrichedTextInput.tsx
hejsztynx: Marked `autoCorrect`'s default explicitly as `true`. Left `spellCheck` here as it is, as its fallback value is evaluated later, using `autoCorrect`'s value - consistent with how RN `TextInput` behaves.

ios/EnrichedTextInputView.mm
```diff
@@ -829,6 +829,20 @@ - (void)updateProps:(Props::Shared const &)props
     }
   }
 
+  // autoCorrect
+  if (newViewProps.autoCorrect != oldViewProps.autoCorrect) {
+    textView.autocorrectionType = newViewProps.autoCorrect
+                                      ? UITextAutocorrectionTypeYes
+                                      : UITextAutocorrectionTypeNo;
+  }
```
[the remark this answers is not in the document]
hejsztynx: That's expected, the React Native TextInput doesn't refresh those on prop changes either

exploIF (approved):
Looks good to me

docs/docs/api-reference/enriched-text-input.md
```diff
@@ -633,6 +641,14 @@ cursor (caret) also uses this color.
...
 | [`color`](https://reactnative.dev/docs/colors) | system default | Android, iOS, Web |
 
+### `spellCheck` {#spellcheck}
+
+If false, disables spell-check style (i.e. red underlines). The default value is inherited from the `autoCorrect` prop.
+
+| Type      | Default       | Platforms |
+| --------- | ------------- | --------- |
+| `boolean` | `autoCorrect` | iOS       |
+
```
szydlovsky, line 651: Let's add a symlink to the `autoCorrect` prop api reference - both in description as well as the table. This way users can navigate straight to the source of the default value.

szydlovsky (approved):
overall looks good, one small comment

hejsztynx:
Thank you @fendent for your work and coming up with this idea!

fendent:
Of course! Glad to help. Sorry I couldn't follow up on this further past the initial PR. Life 🙃

PR #522, opened 2026-04-07
feat(web): add inline styles

pkaramon:
# Summary

+ added support for inline styles on web: bold, italic, underline, strike-through, code.
+ added setup for playwright visual regression tests. Now `yarn test:e2e:web` also accepts `--update-screenshots`.
+ added visual regression test for inline styles.
+ added handling of `htmlStyle` for inline code.
+ moved styles conversion code into `web/styleConversion`.
+ added pretty html output formatting on example-web (only for ease of development).
+ extracted `defaultHtmlStyle` from `src/utils/normalizeHtmlStyle.ts` to it's own file.

## Test Plan

+ Run unit tests `yarn test` see if they pass.
+ Run all web tests `yarn test:e2e:web` see if they pass, especially the visual regression one.
+ Launch example web, see if inlines styles work as expected.

## Screenshots / Videos

image

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ❌     |
| Web |    ✅     |

## Checklist

- [x] E2E tests are passing


apps/example-web/src/components/Toolbar.tsx
```diff
@@ -0,0 +1,113 @@
...
+          <ToolbarButton
+            key={item.key}
+            label={item.label}
+            isActive={s?.[item.key].isActive ?? false}
+            isDisabled={s?.[item.key].isBlocking ?? false}
+            variant={item.variant}
+            onPress={() => {
+              const editor = editorRef.current;
+              if (editor) item.onPress(editor);
+            }}
```
exploIF, line 109: What's the purpose of passing editor instance as a parameter? Seems pretty redundant to me.
Should be enough to call it like below, and  access editor instance directly in the handler functions


```suggestion
            onPress={item.onPress}
```

apps/example-web/src/testScreens/VisualRegressionScreen.tsx
```diff
@@ -0,0 +1,94 @@
...
+    ref.current?.setValue(htmlInput);
+  };
+
+  const handleClear = () => {
+    ref.current?.setValue('');
+    setHtmlInput('');
+  };
+
+  return (
+    <div style={{ padding: '16px' }}>
```
exploIF: Either disable inline styles eslint rule or fix those warnings. I'm fine with both

src/web/useOnChangeState.ts
```diff
@@ -0,0 +1,83 @@
+import { useEffect, useRef } from 'react';
+import { type Editor } from '@tiptap/react';
+import type { OnChangeStateEvent } from '../types';
+import type { NativeSyntheticEvent } from 'react-native';
+import { adaptWebToNativeEvent } from './adaptWebToNativeEvent';
+
+function makeFormatState(isActive: boolean) {
+  return { isActive, isConflicting: false, isBlocking: false };
```
exploIF: Could you add a TODO note to update conflicting/blocking state logic? Conflicting/blocking rules should be consistent across platforms in the final implementation state

src/web/useOnChangeState.ts
```diff
@@ -0,0 +1,83 @@
...
+}
+
+function areFormatStatesEqual(
+  a: OnChangeStateEvent['bold'],
+  b: OnChangeStateEvent['bold']
```
exploIF: Imo, it would be better to create some generic types that we can use here, `["bold"] part might be misleading`

src/web/useOnChangeState.ts
```diff
@@ -0,0 +1,83 @@
...
+  );
+}
+
+function areStatesEqual(a: OnChangeStateEvent, b: OnChangeStateEvent): boolean {
+  const keys = Object.keys(a) as Array<keyof OnChangeStateEvent>;
+
+  for (const key of keys) {
+    if (!areFormatStatesEqual(a[key], b[key])) {
+      return false;
+    }
```
exploIF: I'm thinking if there is a more performant way of checking that.

What if we calculate a hash for the object state, then we just compare hashes and we don't have to store and access previous state (enough to store previous state hash).
We could:
- for the `isActive, isConflicting and isBlocking` generate a number 0-7 like below
```ts
function getFormatHash(isActive: boolean, isConflicting: boolean, isBlocking: boolean): number {
  return (+isActive << 2) | (+isConflicting << 1) | (+isBlocking << 0);
}
``` 
- from that we could generate a 18 digit (number of supported styles) string, like `300016..`. That way we can easily compare hashes rather than objects, each state combination will always produce the same hash

exploIF (commented):
Works really well 🔥

apps/example-web/src/components/Toolbar.tsx
pkaramon: You make a valid point. If we use `editorRef` directly in the `onPress` functions, it triggers a `Cannot access refs during render error`. It's a false positive, but I think I'd rather keep this approach instead of using an eslint-disable comment.

apps/example-web/src/testScreens/VisualRegressionScreen.tsx
pkaramon: I've fixed them

src/web/useOnChangeState.ts
pkaramon: Added a `TODO` note.

src/web/useOnChangeState.ts
pkaramon: After switching to the hashes for comparing states, this `OnChagneStateEvent['bold']` is no longer needed.

src/web/useOnChangeState.ts
pkaramon: Makes sense. I've updated the PR with this implementation

kacperzolkiewski:
It mostly works as expected but I noticed an issue with inline code.
When I toggle inline code and start typing, the cursor is positioned correctly next to the text. However, if I move the cursor away and then return to the end of the inline code, there’s a visible gap between the cursor and the text when I continue typing.


https://github.com/user-attachments/assets/2814ef02-2f3e-4140-be44-61462c9d9cfa

kacperzolkiewski:
I found another issues.
1. When the cursor is placed at the end of a styled segment the style is correctly shown as active. However, if I delete characters in front of the styled text and end up at the same position the style is no longer shown as active:

https://github.com/user-attachments/assets/6dd690ee-5ca0-45fb-9676-a513894e644a


2. When the cursor is placed at the beginning of a styled segment, the style should not be active. However, at the very start of the input, it is incorrectly shown as active:

https://github.com/user-attachments/assets/99fd71ee-83f6-4bf0-979e-1414732b9cd6


3. When I select all text and delete it at once, no styles are active. However, if I delete characters one by one the style remains active even after the input becomes empty:

https://github.com/user-attachments/assets/d26d4851-7159-4118-be86-91f9648cb369

kacperzolkiewski:
Is it possible to format `<br>` tags so that they are rendered as new lines in the output, similar to how they appear on mobile?
Right now, the generated HTML looks like this:

https://github.com/user-attachments/assets/db5e6794-59b7-4963-8c1f-76a5034ebdeb

pkaramon:
@kacperzolkiewski I've fixed the inline code cursor problem on safari.

I think I've managed to fix your 3 issues from the second comment by writing a ProseMirror plugin to match it's behavior to the way things work on iOS/Android. I also wrote tests for those 3 issues.

As for the `<br>` tags, I would like for them to always have a newline i.e.
```html
<br>
<br>
<br>
``` 
instead of `<br><br><br>` but the formatting library didn't want to do that, I've tried and decided it's fine, as the formatted html is for development purpose only.

kacperzolkiewski:
The issues mentioned above no longer occur, but I’ve found some new ones:
1. (it only occurs on safari) There’s a strange cursor behavior when inline code is at the beginning of a line and I try to type before it. For example, I can’t move the cursor below, and it appears in the wrong position:

https://github.com/user-attachments/assets/f8b97ed0-5e31-44b8-afc4-ac190a6b99bd

2. The style is deactivated when pressing Enter inside a styled segment:

https://github.com/user-attachments/assets/1c80d948-602f-4c16-8f67-2281e1e93eb1

3. When text is after some styled segment and we press enter the style does not continue in newline:

https://github.com/user-attachments/assets/0f53c7b3-f33e-4c7f-90ff-dfc98132cb26

pkaramon:
@kacperzolkiewski I've fixed the newer issues as well.

kacperzolkiewski:
The issues mentioned above no longer occur, but I’ve found new one:

1. When applying an inline style across multiple lines, placing the cursor at the beginning of a line does not mark the style as active. I think that it should be active in this case?

We discussed this with @pkaramon, and it seems difficult to achieve this behavior. @pkaramon suggested that a better approach might be to activate the style when the cursor is at the beginning of a styled line (which is the default TipTap behavior).
@exploIF, what are your thoughts on this?


https://github.com/user-attachments/assets/c460216c-6e67-410b-a89c-dc48b17f4b65

pkaramon:
@exploIF A little more context to @kacperzolkiewski question: The style applied at the very start of a line on native depends on whether or not a style was applied to the newline before the current line. This is hard to replicate using Tiptap/ProseMirror as newlines don't really exist, when you press enter you add a new paragraph node.

https://github.com/user-attachments/assets/38a9c6f9-8716-46ef-a12d-1549f07be1f2

src/web/formats/EnrichedBold.ts
```diff
@@ -0,0 +1,16 @@
+import Bold from '@tiptap/extension-bold';
+
+export const EnrichedBold = Bold.extend({
+  addInputRules() {
+    return [];
+  },
+  addKeyboardShortcuts() {
```
exploIF, line 7: Let's keep it as it is for now, but once base web version is implemented, I would like to add keyboard shortcuts for most of the styles, so we handle platform specific behavior correctly

exploIF (approved):

exploIF:
@pkaramon @kacperzolkiewski I'm fine with keeping it as it is for now. It's expected that there will be some inconsistencies between platforms. If we find out that current solution is not sufficient, we can always come back to this topic later

PR #527, opened 2026-04-09
feat: add maxLength support to EnrichedTextInput

xiamu14:
## Summary
- add `maxLength` support to `EnrichedTextInput`
- emit `onMaxLengthExceeded` when native input rejects text that would exceed the limit
- wire the new prop through the JS/native bridge on Android and iOS

## Why
Some apps need to block rich-text input before content is inserted, instead of accepting text and reverting it afterward. This is especially important for candidate insertion and other multi-character commits.

## Notes
- Android rejects overflow using an input filter and emits `onMaxLengthExceeded`
- iOS rejects overflow inside `shouldChangeTextInRange` and emits `onMaxLengthExceeded`
- JS exposes both `maxLength` and `onMaxLengthExceeded`

## Validation
- verified app-side integration against a local package copy in a consumer app
- linted the consumer integration file and the native bridge entry updates locally

### DIFF
--- a/android/src/main/java/com/swmansion/enriched/textinput/EnrichedTextInputView.kt
+++ b/android/src/main/java/com/swmansion/enriched/textinput/EnrichedTextInputView.kt
@@ -10,6 +10,7 @@ import android.graphics.Rect
 import android.graphics.text.LineBreaker
 import android.os.Build
 import android.text.Editable
+import android.text.InputFilter
 import android.text.InputType
 import android.text.Spannable
 import android.text.SpannableString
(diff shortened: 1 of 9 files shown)


android/src/main/java/com/swmansion/enriched/textinput/EnrichedTextInputView.kt
```diff
@@ -338,6 +341,38 @@ class EnrichedTextInputView :
...
+  fun setMaxLength(value: Int?) {
+    maxLength = value
+    filters =
+      if (value == null) {
+        emptyArray()
+      } else {
+        arrayOf(
+          InputFilter { source, start, end, dest, dstart, dend ->
+            val remaining = value - ((dest?.length ?: 0) - (dend - dstart))
+            val incomingLength = end - start
```
exploIF, line 353: We use a lot of `ZWS` characters internally, I think they shouldn't be taken into consideration when calculating length of the text

src/spec/EnrichedTextInputNativeComponent.ts
```diff
@@ -382,6 +387,7 @@ export interface NativeProps extends ViewProps {
   onRequestHtmlResult?: DirectEventHandler<OnRequestHtmlResultEvent>;
   onInputKeyPress?: DirectEventHandler<OnKeyPressEvent>;
   onPasteImages?: DirectEventHandler<OnPasteImagesEvent>;
+  onMaxLengthExceeded?: DirectEventHandler<OnMaxLengthExceededEvent>;
```
exploIF, line 390: I think this event is redundant. There is no such event in the RN core, and I would prefer to avoid adding it into the library. You should be able to handle such case on the JS side with new `maxLength` prop and `onKeyPress` event. T

exploIF (changes requested):

hejsztynx:
This PR has become outdated and appears to be inactive, so we're closing it. We'll continue the development of this feature on our side with a fresh implementation.

PR #532, opened 2026-04-09
feat(iOS): EnrichedText

### LINKED ISSUE #374: [Feature request] EnrichedText

kacperzolkiewski:
# Summary

Fixes: #374 
Implements `EnrichedText` component which can be used for displaying text created by `EnrichedTextInput`.

## Test Plan

Verify that `EnrichedText` component is rendered properly, styles can be customized and any valid html or raw text is displayed properly

## Screenshots / Videos

https://github.com/user-attachments/assets/f995a526-b2a1-4ca4-89ea-46dd45bf72ba

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist

- [X] E2E tests are passing


pkaramon:
An issue which may be not related to this PR: basically after clearing the `EnrichedTextInput` there is still a "phantom image" left:

https://github.com/user-attachments/assets/7183e994-cb74-47cd-9e0b-0ec900c212c3


EnrichedText has problems when it comes to images:



https://github.com/user-attachments/assets/696c972b-c2f2-4eb3-857e-451d1568e167

https://github.com/user-attachments/assets/0335ee4a-86e2-4bb3-8253-29f11eeab5fd


https://github.com/user-attachments/assets/9662506b-acef-442d-941f-5b9254973af0



https://github.com/user-attachments/assets/471dcd4d-dc12-46ce-a091-0ed73b66b746

pkaramon:
This bug was not introduced by this pr, nevertheless it's a bit funny/confusing:

https://github.com/user-attachments/assets/16446b05-a453-4e81-a7b0-9c8b5c578354

ios/config/EnrichedConfig.h
```diff
@@ -101,4 +101,8 @@
 - (void)setCheckboxListBoxColor:(UIColor *)newValue;
 - (UIImage *)checkboxCheckedImage;
 - (UIImage *)checkboxUncheckedImage;
+
+// MARK: - Text only props
+- (UIColor *)linkPressColor;
+- (void)setLinkPressColor:(UIColor *)newValue;
```
szydlovsky, line 111: Not essential but we could create 3 sections: `Shared props`, `Input only props` and `Text only props`. Only highlighting text only props kinda defeats the purpose for me

ios/config/EnrichedConfig.mm
szydlovsky, line 1: Same as in the header file, if we decide to mark and differentiate props and methods of `text` and `input`, we should have 3 separate sections `shared`, `text`, `input`

ios/htmlParser/HtmlParser.h
szydlovsky, line 1: A general comment about the parsers. I am a big fan of the decision to have shared parsing in `HtmlParser` and then any component-specific parsing done in `InputParser`, but I think `EnrichedTextView` should also have its own specialised parser that applies the html in a way it wants. Currently applying is done inside the component and I'd rather we did it the same way input does it.

And a second thing, if we decide to do that I'd change naming a little bit:
- `EnrichedTextInputView`'s parser: `InputParser` -> `InputHtmlParser` (don't forget to change catalog name too)
- `EnrichedTextView`'s parser: `TextHtmlParser` would be cool.

What do you think? IMO it will make the structure ordered better and easier to get into.

ios/interfaces/EnrichedViewHost.h
```diff
@@ -0,0 +1,25 @@
...
+#import "AttributesManager.h"
+#import "EnrichedConfig.h"
+#import "StyleTypeEnum.h"
+#import <UIKit/UIKit.h>
+
+@protocol EnrichedViewHost <NSObject>
+@required
+@property(nonatomic, readonly) UITextView *textView;
+@property(nonatomic, readonly) EnrichedConfig *config;
+@property(nonatomic, readonly) NSDictionary<NSNumber *, id> *stylesDict;
```
szydlovsky: These three are missing nullability specifier. I think all of them should be `nonnull`

ios/interfaces/EnrichedViewHost.h
```diff
@@ -0,0 +1,25 @@
...
+@property(nonatomic, readonly, nullable) AttributesManager *attributesManager;
+@property(nonatomic, readonly, nullable)
+    NSMutableDictionary<NSNumber *, NSArray<NSNumber *> *> *conflictingStyles;
+@property(nonatomic, readonly, nullable)
+    NSMutableDictionary<NSNumber *, NSArray<NSNumber *> *> *blockingStyles;
+@property(nonatomic, readonly, nullable)
+    NSMutableDictionary<NSAttributedStringKey, id> *defaultTypingAttributes;
+@property(nonatomic) BOOL blockEmitting;
+@optional
+- (BOOL)handleStyleBlocksAndConflicts:(StyleType)type range:(NSRange)range;
```
szydlovsky: We don't need it whatsoever, all the call sites can be easily replaced with `StyleUtils` version of this method.

ios/interfaces/EnrichedViewHost.h
```diff
@@ -0,0 +1,25 @@
...
+#import "EnrichedConfig.h"
+#import "StyleTypeEnum.h"
+#import <UIKit/UIKit.h>
+
+@protocol EnrichedViewHost <NSObject>
+@required
+@property(nonatomic, readonly) UITextView *textView;
+@property(nonatomic, readonly) EnrichedConfig *config;
+@property(nonatomic, readonly) NSDictionary<NSNumber *, id> *stylesDict;
+@property(nonatomic, readonly, nullable) AttributesManager *attributesManager;
```
szydlovsky: Maybe a nit, but if we now use `AttributesManager` only for the input, maybe let's rename it to `InputAttributesManager`? I'd keep the property as `attributesManager` though. And if possible, we can move it under the `@optional` specifier below.

ios/EnrichedTextInputView.h
```diff
@@ -15,11 +16,11 @@
 NS_ASSUME_NONNULL_BEGIN
 
 @interface EnrichedTextInputView
-    : RCTViewComponentView <MediaAttachmentDelegate> {
+    : RCTViewComponentView <MediaAttachmentDelegate, EnrichedViewHost> {
```
szydlovsky: Nit: maybe the other way around? I think it's a bit more important that it's a host than the info it's a media attachment delegate 😄 
```suggestion
    : RCTViewComponentView <EnrichedViewHost, MediaAttachmentDelegate> {
```

ios/EnrichedTextInputView.mm
```diff
@@ -1363,7 +1225,10 @@ - (void)focus {
 - (void)setValue:(NSString *)value {
   NSString *initiallyProcessedHtml = [parser initiallyProcessHtml:value];
   if (initiallyProcessedHtml == nullptr) {
-    // just plain text
+    // reset the text first and reset typing attributes
+    textView.text = @"";
+    textView.typingAttributes = defaultTypingAttributes;
+    // set new text
```
szydlovsky, line 1232: Out of curiosity, does this part fix any issue you found along the way?

ios/EnrichedTextInputView.mm
```diff
@@ -1648,55 +1513,10 @@ - (void)startMentionWithIndicator:(NSString *)indicator {
...
-    } else {
-      if ([style detect:range]) {
-        [resultArray addObject:type];
-      }
-    }
-  }
-  return resultArray;
+  return [StyleUtils handleStyleBlocksAndConflicts:type
+                                             range:range
+                                           forHost:self];
```
szydlovsky: As stated way before, we can just always use `StyleUtils` implementation for this method so it's not needed to implement it here (basically just forwarding that implementation anyway).

ios/EnrichedTextInputView.mm
```diff
@@ -2205,129 +2025,19 @@ - (void)textStorage:(NSTextStorage *)textStorage
...
-      changeInLength:0];
-
-  dispatch_async(dispatch_get_main_queue(), ^{
-    [self layoutAttachments];
-  });
+  [AttachmentLayoutUtils handleAttachmentUpdate:attachment
+                                       textView:textView
+                                  onLayoutBlock:^{
+                                    [self layoutAttachments];
+                                  }];
```
szydlovsky: The type for `mediaAttachmentDidUpadte` argument is wrong now, it needs to be `MediaAttachment *` after the changes.

ios/utils/AttachmentLayoutUtils.mm
```diff
@@ -0,0 +1,143 @@
...
++ (NSMutableDictionary<NSValue *, UIImageView *> *)
+    layoutAttachmentsInTextView:(UITextView *)textView
+                         config:(EnrichedConfig *)config
+                  existingViews:
+                      (NSMutableDictionary<NSValue *, UIImageView *> *)
+                          attachmentViews {
+  NSTextStorage *storage = textView.textStorage;
+  if (storage.length == 0)
+    return attachmentViews;
+
```
szydlovsky: This early return skips dangling attachment views deletion that is done at the very end of the method, which may result in memory leaks when removing whole content from an `EnrichedTextInputView` (removing will cause storage to be empty but the old attachments will still be mounted!)

ios/utils/OccurenceUtils.mm
```diff
@@ -26,24 +26,31 @@ + (BOOL)detect:(NSAttributedStringKey _Nonnull)key
...
-    attrValue = input->textView.typingAttributes[key];
-  } else if (index == input->textView.textStorage.string.length) {
+  // Only trust typingAttributes when the textView is actually editable.
+  // Non-editable hosts (e.g. EnrichedTextView) keep selectedRange pinned at
+  // (0, 0), so without this gate every detection at index 0 would match the
+  // selection and read stale/default typingAttributes instead of the real
+  // attribute at that position in textStorage.
+  if (host.textView.isEditable &&
+      NSEqualRanges(host.textView.selectedRange, detectionRange)) {
+    NSLog(@"[OccurenceUtils] isEditable srodek: %d", host.textView.isEditable);
```
szydlovsky: debug log 🚒

ios/utils/ZeroWidthSpaceUtils.h
```diff
@@ -1,9 +1,12 @@
+#import "EnrichedViewHost.h"
 #import <UIKit/UIKit.h>
 #pragma once
 
 @interface ZeroWidthSpaceUtils : NSObject
-+ (void)handleZeroWidthSpacesInInput:(id)input;
++ (void)handleZeroWidthSpacesInInput:(id<EnrichedViewHost>)host;
++ (void)addSpacesIfNeededinInput:(id<EnrichedViewHost>)host
+                         inRange:(NSRange)range;
```
szydlovsky: I think the names are not relevant anymore, maybe `handleZeroWidthSpaceUtilsInHost` and `addSpacesIfNeededInHost`?

ios/utils/ZeroWidthSpaceUtils.mm
```diff
@@ -245,20 +256,19 @@ + (BOOL)handleBackspaceInRange:(NSRange)range
 }
 
 + (BOOL)anyZWSStylePresentInRange:(NSRange)range
-                            input:(EnrichedTextInputView *)input {
-  for (NSNumber *type in input->stylesDict) {
-    StyleBase *style = input->stylesDict[type];
+                            input:(id<EnrichedViewHost>)host {
```
szydlovsky: Should be `host:(id<EnrichedViewHost>)host` now I think.

ios/utils/ZeroWidthSpaceUtils.mm
```diff
@@ -83,13 +85,13 @@ + (void)removeSpacesIfNeededinInput:(EnrichedTextInputView *)input {
 // dictionary so that ZWS characters carry the same meta-attributes that are
 // currently active in the typing attributes. Only within the currently selected
 // range!
-+ (NSDictionary *)inlineMetaAttributesForInput:(EnrichedTextInputView *)input {
++ (NSDictionary *)inlineMetaAttributesForInput:(id<EnrichedViewHost>)host {
```
szydlovsky: Also `input` not relevant here anymore.
```suggestion
+ (NSDictionary *)inlineMetaAttributesForHost:(id<EnrichedViewHost>)host {
```

ios/utils/ZeroWidthSpaceUtils.mm
```diff
@@ -245,20 +256,19 @@ + (BOOL)handleBackspaceInRange:(NSRange)range
...
     }
   }
   return NO;
 }
 
-+ (BOOL)removeZWSStyleInRange:(NSRange)range
-                        input:(EnrichedTextInputView *)input {
-  for (NSNumber *type in input->stylesDict) {
-    StyleBase *style = input->stylesDict[type];
++ (BOOL)removeZWSStyleInRange:(NSRange)range input:(id<EnrichedViewHost>)host {
```
szydlovsky: Needs renaming too.
```suggestion
+ (BOOL)removeZWSStyleInRange:(NSRange)range host:(id<EnrichedViewHost>)host {
```

ios/utils/TextInsertionUtils.h
szydlovsky, line 1: All the methods from this util should be now using `host` instead of `input` as the argument labels

ios/utils/ZeroWidthSpaceUtils.h
```diff
@@ -1,9 +1,12 @@
...
 + (BOOL)handleBackspaceInRange:(NSRange)range
                replacementText:(NSString *)text
-                         input:(id)input;
+                         input:(id<EnrichedViewHost>)host;
```
szydlovsky: This one too should use `host` instead of `input` now.

ios/EnrichedTextViewManager.mm
```diff
@@ -0,0 +1,13 @@
...
+#import <React/RCTViewManager.h>
+
+@interface EnrichedTextViewManager : RCTViewManager
+@end
+
+@implementation EnrichedTextViewManager
+
+RCT_EXPORT_MODULE(EnrichedTextView)
+
+RCT_EXPORT_VIEW_PROPERTY(text, NSString)
```
szydlovsky, line 11: unsure if it's even needed but there's no harm in leaving it here as input's manager also has some random property there 😄

ios/EnrichedTextView.mm
```diff
@@ -0,0 +1,880 @@
...
+
+  NSMutableAttributedString *body = [[NSMutableAttributedString alloc]
+      initWithString:plainText
+          attributes:defaultTypingAttributes];
+  [textView.textStorage setAttributedString:body];
+  [self applyProcessedStyles:processedStyles];
+  [self layoutAttachments];
+}
+
+- (void)applyProcessedStyles:(NSArray *)processedStyles {
```
szydlovsky: If we decide for a separate `TextHtmlParser`, the `applyProcessedStyles` method should probably go there.

ios/EnrichedTextInputView.mm
kacperzolkiewski: Yes, this fix a lot :D 
There were problems when we had something styled in the input and then call `setValue` with some text. 
If we had some for example list styling in the input the bullet would be drawn for new text without this.

ios/EnrichedTextViewManager.mm
kacperzolkiewski: let's leave it here right now :P

ios/htmlParser/HtmlParser.h
kacperzolkiewski: Nice idea, done ;)

ios/interfaces/EnrichedViewHost.h
kacperzolkiewski: done

ios/utils/AttachmentLayoutUtils.mm
kacperzolkiewski: Nice catch, fixed

ios/textHtmlParser/TextHtmlParser.mm
```diff
@@ -0,0 +1,163 @@
...
+#import "HtmlParser.h"
+#import "LinkData.h"
+#import "MentionParams.h"
+#import "StyleHeaders.h"
+#import "StyleUtils.h"
+#import "ZeroWidthSpaceUtils.h"
+#import <React/RCTLog.h>
+
+@interface TextHtmlParser ()
+@end
```
szydlovsky: Is that needed?

src/spec/EnrichedTextNativeComponent.ts
szydlovsky, line 1: Not sure, but is the flag also supported in Android implementation that we are merging into? Just making sure.

ios/utils/EnrichedTextTouchHandler.mm
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    NSMutableDictionary *newAttrs = [[NSMutableDictionary alloc] init];
+    if (color) {
+      newAttrs[NSForegroundColorAttributeName] = color;
+      newAttrs[NSUnderlineColorAttributeName] = color;
+      newAttrs[NSStrikethroughColorAttributeName] = color;
+    }
+    if (bgColor) {
+      newAttrs[NSBackgroundColorAttributeName] = bgColor;
```
szydlovsky: We need to set the bg the same way regular mention's bg is set, so with a proper alpha using `colorWithAlphaIfNotTransparent:0.4` on the UIColor.
Now the color change is way too drastic 😄

szydlovsky (commented):
It's not a regression but rather an inconsistency  in example app:
- `EnrichedText` uses `htmlStyle` from `src/constants/htmlStyle`
- `EnrichedTextInput` uses `htmlStyle` from `src/constants/editorConfig`
and they are, well, different:

Image

Can we please fix it? I think the best thing to do is to make `enrichedTextHtmlStyle` extend the `constants/editorConfig` style, not the one it does now

src/spec/EnrichedTextNativeComponent.ts
kacperzolkiewski: I will add it in a moment :D

ios/textHtmlParser/TextHtmlParser.mm
kacperzolkiewski: removed

szydlovsky (approved):
Image

PR #549, opened 2026-04-15
feat(web): add headings, blockquotes and codeblocks

pkaramon:
# Summary

+ added `<h1>` to `<h6>`, `<blockquote>` and `<codeblock>` to web
+ added handling of htmlStyle for new elements
+ added visual regression tests for new elements
+ added tests for custom keyboard navigation, merging adjacent blocks etc. behavior that was added to match tiptap to the native implemenation
+ moved StrictMarksPlugin to `pmPlugins` (ProseMirror plugins), a directory that contains a lot of the custom logic to match tiptap to native
+ VisualRegression screen now displays the current html and also allows customizing htmlStyle during a test
+ removed padding from inline code, as the native does not have it, and it also caused issues with headings, where the padding intruded on elements below.

## Test Plan

+ Run `yarn test:e2e:web` to see if all the new web tests pass
+ Run `yarn example-web dev` and play around with new elements

## Screenshots / Videos

image

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ❌     |
| Web  |    ✅     |

## Checklist

- [x] E2E tests are passing


kacperzolkiewski:
There’s some unexpected behavior when selecting lines with and without paragraph styles. The toolbar still indicates that paragraph styles can be toggled, but they don’t actually apply.

https://github.com/user-attachments/assets/b28a1f13-c9a0-4892-a9c6-bf55430f8b43

src/web/formats/EnrichedCodeBlock.ts
```diff
@@ -0,0 +1,55 @@
...
+  interface Commands<ReturnType> {
+    enrichedCodeBlock: {
+      toggleEnrichedCodeBlock: () => ReturnType;
+    };
+  }
+}
+
+// We extend Blockquote, not CodeBlock: this node wraps (paragraph)+; TipTap's CodeBlock is a leaf block, not a paragraph container.
+export const EnrichedCodeBlock = Blockquote.extend({
+  name: 'enrichedCodeBlock',
```
kacperzolkiewski: Since we already use `blockquote` and `heading`, would it make sense to rename this to `codeblock` instead of `enrichedCodeBlock` for consistency?

src/web/EnrichedTextInput.tsx
```diff
@@ -49,6 +64,19 @@ export const EnrichedTextInput = ({
   const tiptapContent =
     defaultValue != null ? prepareHtmlForTiptap(defaultValue) : defaultValue;
 
+  const resolvedHtmlStyle = useMemo(
+    () => mergeWithDefaultHtmlStyle(htmlStyle),
+    [htmlStyle]
+  );
+
+  const htmlStyleRef = useRef(resolvedHtmlStyle);
+  htmlStyleRef.current = resolvedHtmlStyle;
```
kacperzolkiewski: I am not sure if we are allowed to assign refs like this, look here:
https://react.dev/reference/react/useRef#avoiding-recreating-the-ref-contents

apps/example-web/src/testScreens/VisualRegression.tsx
```diff
@@ -7,13 +7,47 @@ import {
   type OnChangeStateEvent,
 } from 'react-native-enriched';
 import { Toolbar } from '../components/Toolbar';
+import { defaultHtmlStyle } from '../defaultHtmlStyle';
+
+function mergeHtmlStyle(
```
kacperzolkiewski, line 12: Why we don't use here already created `mergeWithDefaultHtmlStyle`?

apps/example-web/src/defaultHtmlStyle.ts
```diff
@@ -0,0 +1,43 @@
+import type { HtmlStyle } from 'react-native-enriched';
+
+export const defaultHtmlStyle: HtmlStyle = {
```
kacperzolkiewski: Shouldn’t this be capitalized like in the native example for consistency?

src/web/pmPlugins/strictMarksPlugin.ts
kacperzolkiewski, line 1: We should rename this file to `strictMarksPlugin.ts` to match the camelCase convention used by other files in this directory.

src/web/formats/EnrichedCodeBlock.ts
pkaramon: TLDR: I've switched to just `codeBlock` without `enriched` prefix.

The original thought behind this `enriched` prefix was that we are not basing our `EnrichedCodeBlock` on tiptap's `CodeBlock`. We instead write our own, based on `Blockquote`. So this `enriched` makes it explicit that this is something a lot different than tiptap's `Codeblock` which uses the `codeBlock` name.

Now looking at how much we customize the extensions, I don't think this `enriched` prefix is very useful after all.

src/web/EnrichedTextInput.tsx
pkaramon: You're right. I've moved setting the ref to a `useEffect` hook.

apps/example-web/src/defaultHtmlStyle.ts
pkaramon: That constant is only used to share the same styles on web: in the example app screen and in the visual regression screen. I've renamed it to `WEB_DEFAULT_HTML_STYLE`.

apps/example-web/src/testScreens/VisualRegression.tsx
pkaramon: Good question. Two reasons:
- `mergeWithDefaultHtmlStyle` doesn’t accept a parameter for which defaults to use. The web example’s defaults differ from the library’s. That could be refactored, but see below.
- Even then, that helper (or a generalized version) **isn’t exported** from our package and we don’t intend to export it. So the `example-web` won't be able to import that function.

pkaramon:
@kacperzolkiewski thank you for catching that bug. It has been fixed.

src/web/useOnChangeState.ts
```diff
@@ -31,37 +37,53 @@ export const useOnChangeState = (
...
-  // blocking. Make sure conflicting and blocking states are in sync between web
-  // and native
-  return { isActive, isConflicting: false, isBlocking: false };
-}
+function buildState(
+  editor: Editor,
+  htmlStyle: Required<HtmlStyle>
+): OnChangeStateEvent {
+  const isBlockquoteActive = editor.isActive('blockquote');
+  const isCodeBlockActive = editor.isActive('codeBlock');
```
exploIF: Nit, why it's declared on top of the function for block quote and code block, but inline for headings? It should be consistent

src/web/EnrichedTextInput.tsx
```diff
@@ -125,14 +170,14 @@ export const EnrichedTextInput = ({
...
       toggleInlineCode: () => editor.chain().focus().toggleCode().run(),
-      toggleH1: () => {},
-      toggleH2: () => {},
-      toggleH3: () => {},
-      toggleH4: () => {},
-      toggleH5: () => {},
-      toggleH6: () => {},
-      toggleCodeBlock: () => {},
-      toggleBlockQuote: () => {},
+      toggleH1: () => editor.chain().focus().toggleHeading({ level: 1 }).run(),
```
exploIF: `editor.chain.focus()` is repeated over and over, maybe it could moved to separate function?

src/web/EnrichedTextInput.tsx
```diff
@@ -97,9 +136,15 @@ export const EnrichedTextInput = ({
     [tiptapContent]
   );
 
+  useEffect(() => {
+    if (!editor) return;
+    editor.commands.normalizeBoldInStyledHeadings();
```
exploIF:
```suggestion
    editor?.commands.normalizeBoldInStyledHeadings();
```

exploIF (approved):

kacperzolkiewski (approved):
LGTM 🚀

apps/example-web/src/testScreens/VisualRegression.tsx
```diff
@@ -7,13 +7,47 @@ import {
...
+  const isPlainObject = (val: unknown): val is Record<string, unknown> =>
+    typeof val === 'object' && val !== null && !Array.isArray(val);
+
+  const result = { ...base } as Record<string, unknown>;
+
+  for (const [key, val] of Object.entries(override)) {
+    const prev = result[key];
+    result[key] =
+      isPlainObject(prev) && isPlainObject(val) ? { ...prev, ...val } : val;
+  }
```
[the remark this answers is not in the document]
pkaramon: This function is only a helper for a test screen. We do not export this as part of our library therefore it's fine.

src/web/useOnChangeState.ts
pkaramon: I guess I didn't have a particular reason. I've inlined those two variables.

src/web/EnrichedTextInput.tsx
pkaramon: I've created a `runFocused` helper.

PR #551, opened 2026-04-17
fix(iOS): formatting stripped when autocorrect applied

### LINKED ISSUE #548: [iOS] Bold and other formatting stripped when autocorrect replaces a word
Describe the bug
  When typing with Bold enabled on iOS, autocorrect/autocomplete interferes with the bold formatting. As you type a word, the autocorrect suggestion replaces the word and simultaneously removes the bold style. The replacement text loses its bold
   formatting.

  This is a regression introduced in 0.6.x — the issue did not exist in 0.5.2.

  To Reproduce
  1. Open the app and create a new journal entry
  2. Tap the Bold button in the toolbar to enable bold formatting
  3. Start typing a word, e.g., "hello worls"
  4. Notice "worls" gets auto-selected by iOS autocorrect (suggesting "world")
  5. Press Space to accept the autocorrect suggestion
  6. Observe: the bold formatting is removed from the corrected word ("world" appears in normal weight), and the bold state is deactivated in the toolbar

  Expected behavior
  When autocorrect replaces a misspelled word while Bold is active:
  - The replacement text ("world") should retain bold formatting
  - The Bold state in the toolbar should remain active
  - Subsequent typing should continue in bold

  This worked correctly in version 0.5.2.

  Vidoe
https://github.com/user-attachments/assets/2ac95dc4-06e9-4db6-b2e5-52e5ae56000a

  Device (please complete the following information):
  - Device: iPhone (tested on simulator and real device)
  - OS: iOS 26+
  - Library version: react-native-enriched 0.6.1 

  Additional context
  - This only happens on iOS. Android does not exhibit this behavior.

kacperzolkiewski:
# Summary

Fixes: #548 

This PR prevents iOS autocorrect from stripping formatting by capturing active meta-attributes in `shouldChangeTextInRange` and reapplying them in `didProcessEditing`

## Test Plan

Run reproduction steps from #548, the issue should be resolved.

## Screenshots / Videos

https://github.com/user-attachments/assets/dce5a5b5-0626-49bd-8fd1-1c8b34b19efc

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist

- [ ] E2E tests are passing

### DIFF
--- a/ios/EnrichedTextInputView.mm
+++ b/ios/EnrichedTextInputView.mm
@@ -1995,8 +1995,9 @@ Class<RCTComponentViewProtocol> EnrichedTextInputViewCls(void) {
 - (bool)textView:(UITextView *)textView
     shouldChangeTextInRange:(NSRange)range
             replacementText:(NSString *)text {
-  // Capture the attributes of the word being replaced (autocorrect /
-  // predictive) so didProcessEditing: can re-stamp them onto the replacement.
+  // Capture the attributes at range.location that are being replaced
+  // (autocorrect / predictive) so didProcessEditing: can re-stamp them onto the
+  // replacement.
   if (range.length > 0) {
     _capturedAttributesBeforeChange =
         [textView.textStorage attributesAtIndex:range.location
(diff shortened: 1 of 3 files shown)


ios/EnrichedTextInputView.mm
```diff
@@ -1994,6 +1995,14 @@ - (void)handleKeyPressInRange:(NSString *)text range:(NSRange)range {
...
     shouldChangeTextInRange:(NSRange)range
             replacementText:(NSString *)text {
+  // Capture the attributes of the word being replaced (autocorrect /
+  // predictive) so didProcessEditing: can re-stamp them onto the replacement.
+  if (range.length > 0) {
+    _capturedAttributesBeforeChange =
+        [textView.textStorage attributesAtIndex:range.location
+                                 effectiveRange:NULL];
+  }
+
```
pkaramon: This comment is slightly misleading since we only capture attributes at `range.location`, not the entire range.

In a case like `<b>hen<i>lo</i></b>`, we’d only capture the bold from the first character and lose the italic styling. That’s probably fine. If the replacement text differs, there’s no clear way to reapply multiple styles anyway.

pkaramon (approved):
LGTM

ios/EnrichedTextInputView.mm
szydlovsky: +1

ios/EnrichedTextInputView.mm
```diff
@@ -2190,6 +2199,26 @@ - (void)textStorage:(NSTextStorage *)textStorage
...
+                              value:_capturedAttributesBeforeChange[key]
+                              range:editedRange];
+        }
+      }
+    }
+
+    // Clear after consuming
+    _capturedAttributesBeforeChange = nil;
+  }
+
```
szydlovsky: Can't we put this whole logic inside the following identical
`if ((editedMask & NSTextStorageEditedCharacters) != 0)` statement? Just before we do anything with dirty ranges though

szydlovsky (approved):
Works and looks good, some more stylistic comments.

ios/EnrichedTextInputView.mm

ios/EnrichedTextInputView.mm

PR #552, opened 2026-04-20
feat(android): Add gumbo parser to enriched text

kacperzolkiewski:
# Summary

Add gumbo parser to enriched text

## Test Plan
Insert html to `EnrichedText` that is not fully aligned with `EnrichedTextInput` html output:
`<p>hello</p>`
`<codeblock>code</codeblock>`

The html should be displayed properly.
## Screenshots / Videos

https://github.com/user-attachments/assets/04300737-2a40-4848-b7df-adf311c99462

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

## Checklist

- [x] E2E tests are passing


pkaramon:
It would be good to add e2e tests to cover this.

.maestro/enrichedText/flows/external_html_parsing.yaml
```diff
@@ -0,0 +1,19 @@
...
+    id: 'toggle-screen-button'
+
+- tapOn:
+    id: 'toggle-enriched-text-screen-button'
+
+- runFlow:
+    file: '../subflows/set_enriched_text_value.yaml'
+    env:
+      VALUE: 'Hello<pre>code</pre><span>span paragraph</span><blockquote>blockquote</blockquote>'
+- runFlow:
```
pkaramon, line 17: I would add a newline here

```suggestion
 
- runFlow:
```

pkaramon (approved):
LGTM

exploIF (commented):
In the input when we are not able to normalize html we do use raw html value as a content, looks like this PR uses different approach. Could you elaborate on that?

android/src/main/java/com/swmansion/enriched/text/EnrichedTextView.kt
```diff
@@ -73,19 +76,47 @@ class EnrichedTextView : AppCompatTextView {
...
     valueDirty = false
-    val isHtml = text.startsWith("<html>") && text.endsWith("</html>")
-    if (!isHtml) {
+
+    val parsed = parseText(text, style)
+    if (parsed != null) {
+      setText(parsed, BufferType.SPANNABLE)
+      observeAsyncImages()
+    } else {
       this.text = text
```
kacperzolkiewski, line 85: > In the input when we are not able to normalize html we do use raw html value as a content, looks like this PR uses different approach. Could you elaborate on that?

@exploIF here we also use raw html if we weren't able to normalize HTML?
`parseText` returns null so we use raw html

exploIF (approved):

PR #555, opened 2026-04-20
fix(iOS): `onLinkDetected` fires multiple times with partial ranges for multi-word links

### LINKED ISSUE #550: [iOS]: onLinkDetected called multiple times when cursor is in the same link
**Describe the bug**
`onLinkDetected`, when moving the cursor around in multiword link, is called too frequently. It's called when crossing the word boundaries. It should be called once with the entire link text.

**To Reproduce**
1. In `useEditorState.ts` set
```js
  const handleLinkDetected = (state: CurrentLinkState) => {
    console.log('link detected', state);
    setCurrentLink(state);
  };
```
2. Launch example app
3. Set content to: 
```html
<html>
<p><a href="swmansion.com">The quick brown fox jumps over the lazy dog</a></p>
</html>
```
4. Click on the link in the input and move the cursor around
5. onLinkDetected is called even when we are still in the same link but we've only crossed a word boundary. Also the length implied by `end-start` does not match with the length of the `text` passed

**Expected behavior**

onLinkDetected should be called only once, additional cursor movement within the link shouldn't cause onLinkDetected to be called multiple times

**Screenshots**

https://github.com/user-attachments/assets/8c2532d2-e9de-4940-913a-b57184aacea0

 - Device: iPhone17
 - OS: iOS26.2

**Additional context**

kacperzolkiewski:
# Summary
Fixes: #550 
Override `isEqual:` so runs with identical link data get merged and the full range is returned correctly.

`LinkData` was missing `isEqual:`, so `NSAttributedString` used pointer equality to compare attribute values. This kept adjacent runs of the same logical link separate, causing `getFullLinkRangeAt:` to return a partial (per-word) range and `onLinkDetected` to re-fire on every word boundary.

## Test Plan

Run reproduction steps from #550,
The issue should be resolved.

## Screenshots / Videos

https://github.com/user-attachments/assets/e78ceb5a-8165-4a6c-924f-78fb62ec69bd

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist

- [ ] E2E tests are passing

### DIFF
--- a/ios/interfaces/LinkData.mm
+++ b/ios/interfaces/LinkData.mm
@@ -26,4 +26,18 @@
   return equalText && equalUrl && equalIsManual;
 }
 
+- (BOOL)isEqual:(id)other {
+  if (other == self) {
+    return YES;
+  }
+  if (![other isKindOfClass:[LinkData class]]) {
+    return NO;
+  }
+  return [self isEqualToLinkData:(LinkData *)other];
+}
+
+- (NSUInteger)hash {
+  return [self.text hash] ^ [self.url hash] ^ (self.isManual ? 1u : 0u);
+}
+
 @end


pkaramon (approved):
Works! 🚀

szydlovsky (approved):

PR #557, opened 2026-04-20
fix(android): remove link styling inside codeblock

### LINKED ISSUE #554: [Android]: it's possible to create a link in a codeblock
**Describe the bug**
If we input a link with a link, and on that line we toggle the codeblock style the link style isn't stripped, it's visible and it's also outputted in the html.

**To Reproduce**
1. Launch example app
2. Create an example link
3. Toggle codeblock
4. The link styles aren't stripped.

**Expected behavior**
Link style should be stripped.

**Screenshots**

https://github.com/user-attachments/assets/580e7be0-c947-4c37-8194-197117a788a1

 - Device: Pixel 9
 - OS: Android 16
 - Version 36 "Baklava"

**Additional context**

kacperzolkiewski:
# Summary

Fixes: #554 

This PR adds link and mention as conflicting styles to codeblock

## Test Plan

Run reproduction steps from: #554 
The issue should be resolved.

## Screenshots / Videos

https://github.com/user-attachments/assets/c485ffdb-a669-486c-ba20-19339d8a5e7b

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

## Checklist

- [ ] E2E tests are passing


android/src/main/java/com/swmansion/enriched/textinput/styles/ParametrizedStyles.kt
```diff
@@ -220,8 +220,17 @@ class ParametrizedStyles(
...
+
+    // ZWS-only insertions are internal anchors for paragraph/list spans,
+    // not user input. Skip auto-detection for them so toggle operations
+    // don't accidentally re-apply links they just removed.
+    if (editEnd - editStart == 1 &&
+      editStart < spannable.length &&
+      spannable[editStart] == EnrichedConstants.ZWS
+    ) {
+      return
+    }
```
exploIF: I think it's not a reliable way of checking whether link can be applied or not. There is a `canLinkBeApplied` function which should be used. For the `toggle operations don't accidentally re-apply links they just removed` logic, this is why we have `runAsTransaction` to prevent from re-applying styles when they have been programmatically modified

exploIF (changes requested):

android/src/main/java/com/swmansion/enriched/textinput/styles/ParametrizedStyles.kt
kacperzolkiewski: I deleted this part of code. Right now I wrap toggleStyle in `runAsATransaction` and it works

exploIF (approved):

PR #559, opened 2026-04-21
fix(android): ellipsizeMode in EnrichedTextView

kacperzolkiewski:
# Summary

This PR fixes setting `ellipsizeMode` in `EnrichedText` 

- refactors how async images are handled by creating separate  `observeAsyncDrawableLoaded` for `EnrichedTextImageSpan`
- remove `movementMethod` and override `onTouchEvent` because `ellipsizeMode` does not work with `movementMethod`:  https://stackoverflow.com/questions/46056046/textview-maxlines-movement-method-and-ellipsize

## Test Plan

Run example app and 
1. play around with displaying images/async images inside EnrichedText.
2. test link and mention touch event handlers

## Screenshots / Videos

https://github.com/user-attachments/assets/a591cdf7-1e9a-4e68-b15e-297bd4039adf

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

## Checklist

- [ ] E2E tests are passing


kacperzolkiewski:
I’m not sure why, but `link_press.yaml` and `mention_press.yaml` tests are failing. It looks like they’re unable to trigger presses on links and mentions. However, everything works as expected when tested manually.
@pkaramon, have you encountered a similar issue in other tests?

pkaramon:
@kacperzolkiewski really interesting issue because usually when maestro isn't able to click on something you get an error saying that. But here it fails silently and only the screenshot part fails due to different dimensions. By doing `maestro hierarchy` when the link in `EnrichedText` is visible I've checked and 
```json
{
  "attributes" : {
    "text" : "Link",
    "accessibilityText" : "",
    "hintText" : "",
    "ignoreBoundsFiltering" : "false",
    "resource-id" : "",
    "clickable" : "false",
    "bounds" : "[85,1184][995,1252]",
    ...
  },
  "children" : [ ],
  "clickable" : false,
  ...
                    
}
```

it's able to see it. Maybe it doesn't like the fact that `clickable: false` I'm not sure though. Updating the maestro version to `2.4.0` also didn't help so I don't have a clean solution for this. As a workaround you could do:
```yaml
- tapOn:
    id: "enriched-text"
    point: "5%,50%"
```
instead of
```yaml
- tapOn:
    text: 'Link'
```


Maybe using something like [AccessibilityClickableSpanCompat](https://developer.android.com/reference/kotlin/androidx/core/view/accessibility/AccessibilityClickableSpanCompat?hl=en) would set `clickable:true` and allow maestro to click that normally. But I'm not sure.

kacperzolkiewski:
@pkaramon thanks for the suggestion. I’ve updated the tests accordingly and fixed the issue using:
```yaml
- tapOn:
    id: "enriched-text"
    point: "5%,50%"
```

pkaramon:
I don't think all of the combinations of `numberOfLines` and `ellipsizeMode` are working correctly. It seems liks only `ellipsizeMode = tail` works correctly for 2 lines.

The html in the example is just `<html><p> ... a long string ... </p></html>`

https://github.com/user-attachments/assets/309a17f8-d1fd-49a5-bd1d-1117b009e176

.maestro/enrichedText/flows/ellipsize_head.yaml
pkaramon, line 1: Flows `ellipsize_{head, middle, tail}.yaml` are pretty much identical. Maybe we could extract a subflow, something like: 
```yaml
appId: swmansion.enriched.example
---
- tapOn:
    id: "number-of-lines-${NOL_BUTTON}-button"

- assertVisible: "numberOfLines: ${NOL_LABEL}"

- tapOn:
    id: "ellipsize-${ELLIPSIZE_MODE}-button"

- assertVisible: "ellipsizeMode: ${ELLIPSIZE_MODE}"

- runFlow:
    file: "./capture_or_assert_screenshot.yaml"
    env:
      SCREENSHOT_NAME: "${SCREENSHOT_NAME}"
```

and create a single `ellipsizeMode.yaml` that would go into the test screen, set the value with example text and then tested out every combination of `ellipsizeMode` and `numberOfLines`.

exploIF:
@pkaramon yes, truncation mode works good just for the single line text. That's a [known issue in react-native too](https://reactnative.dev/docs/text#ellipsizemode). Let's keep it as it is and make sure it's documented clearly in our API Reference

apps/example/src/screens/EnrichedTextScreen.tsx
```diff
@@ -9,13 +9,17 @@ import { Button } from '../components/Button';
 import { ValueModal } from '../components/ValueModal';
 import { enrichedTextHtmlStyle } from '../constants/htmlStyle';
 
+type EllipsizeMode = 'head' | 'middle' | 'tail';
```
exploIF: Maybe instead:

```suggestion
type EllipsizeMode = EnrichedTextProps['ellipsizeMode']
```

exploIF (approved):
🚀

.maestro/enrichedText/flows/ellipsize_head.yaml
kacperzolkiewski: @pkaramon nice idea, fixed ;)

pkaramon (approved):
🚀

PR #560, opened 2026-04-22
feat(web): unordered and ordered lists

pkaramon:
# Summary
+ added support for unordered and ordered lists in the web version
+ updated example-web to be able to use `<ol>` and `<ul>`
+ added necessary e2e tests for list
+ customized default tiptap's keybindings to match the native version: backspace at the start of a list item removes the style,
enter inside a list always extends it.
+ added style handling for `<ol>` and `<ul>`
+ fixed a small bug related to mismatched transaction in `toggleParagraphFormat` we need to pass the chain in order not to create two transactions

## Test Plan

- Run `yarn test:e2e:web` and verify all the tests pass.
- Play around with lists in example-web, especially toggling, selection etc.

## Screenshots / Videos

image

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ❌     |
| Web |    ✅     |

## Checklist

- [x] E2E tests are passing


kacperzolkiewski:
Is the toolbar scrollable? Because I cannot scroll it horizontally and lists buttons are not visible? 

image

exploIF:
Looks like shortcuts are not working at all, ideally we should have consistency here between web and native (which uses `- ` and `1. `)

.playwright/tests/applyWrappingListSelection.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+  '<p>four</p>',
+  '<p>five</p>',
+  '</html>',
+].join('');
+
+const ROUND_TRIP_SNAPSHOT = 'list-wrap-selection-round-trip.png';
+
+const LIST_VARIANTS = [
+  { label: 'bullet', toolbarTestId: 'unorderedList' as const },
+  { label: 'ordered', toolbarTestId: 'orderedList' as const },
```
exploIF: Is it required?
```suggestion
  { label: 'bullet', toolbarTestId: 'unorderedList' },
  { label: 'ordered', toolbarTestId: 'orderedList' },
```

.playwright/tests/listKeyboard.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+    listTagName: 'ul' as const,
+  },
+  {
+    label: 'ordered',
+    wrap: (inner: string) => `<html><ol>${inner}</ol></html>`,
+    wrapperSelector: '.eti-editor ol',
+    toolbarTestId: 'orderedList' as const,
+    listTagName: 'ol' as const,
+  },
+] as const;
```
exploIF, line 34: Same here, do we need `as const` for specific items if whole array is marked `as const`?

.playwright/tests/listKeyboard.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+      await setEditorHtml(page, wrap('<li><p>Line</p></li>'));
+
+      await editor.click();
+      await items.first().click();
+      await editor.press('End');
+
+      const enters = 3;
+      for (let i = 0; i < enters; i++) {
+        await editor.press('Enter', { delay: 60 });
+      }
```
exploIF:
```suggestion
      for (let i = 0; i < 3; i++) {
        await editor.press('Enter', { delay: 60 });
      }
```

.playwright/tests/listKeyboard.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+      const secondLine = lines.nth(1);
+      await secondLine.click();
+      await editor.press('End');
+      for (let i = 0; i < 'second'.length; i++) {
+        await editor.press('ArrowLeft', { delay: 60 });
+      }
+
+      await editor.press('Backspace', { delay: 60 });
+
+      await page.waitForTimeout(200);
```
exploIF: This statement with hardcoded `200` is repeated over and over. Let's define 200 as a const at the top of the file

.playwright/tests/listKeyboard.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+      await expect(editor).toHaveScreenshot(
+        `list-keyboard-backspace-after-lift-${label}.png`
+      );
+
+      await editor.press('Backspace', { delay: 60 });
```
exploIF: Same here for the delay value, let's move it to separate const

.playwright/tests/listNestedHtml.spec.ts
```diff
@@ -0,0 +1,53 @@
+import { test, expect } from '@playwright/test';
+
+import {
+  editorLocator,
+  gotoVisualRegression,
+  setEditorHtml,
+} from '../helpers/visual-regression';
+
+const CASES = [
```
exploIF, line 9: Same, regarding `as const`

apps/example-web/src/components/Toolbar.tsx
```diff
@@ -148,6 +148,20 @@ export function Toolbar({ editorRef, state }: ToolbarProps) {
         editor?.toggleCodeBlock();
       },
     },
+    {
```
exploIF, line 153: Those items are not visible in my example app. Missing scrolling?

src/web/formats/listKeyboard.ts
```diff
@@ -0,0 +1,54 @@
...
+export function listEnter(editor: Editor, itemName: string): boolean {
+  if (!editor.isActive(itemName)) {
+    return false;
+  }
+  if (editor.commands.splitListItem(itemName)) {
+    return true;
+  }
+
+  const { selection } = editor.state;
+  if (!isTextSelection(selection) || !selection.empty) {
```
exploIF: Minor one, I would flip this condition as truthy check is more performant than invoking function
 ```suggestion
   if (!selection.empty || !isTextSelection(selection))
```

File: src/web/formats/listKeyboard.ts [hunk 9]
```diff
@@ -0,0 +1,54 @@
...
+    return false;
+  }
+
+  const $from = selection.$from;
+  if ($from.parent.content.size > 0) {
+    return false;
+  }
+
+  // Flat lists only: list item is always the parent block of the paragraph (depth − 1).
+  const itemDepth = $from.depth - 1;
```
exploIF, line 32: Is it possible to land on `-1` here?

pkaramon:
Sorry, I've forgotten to mention that the list icons are visible you just have to use horizontal scrolling, for me it swiping left and right on a trackpad. Web doesn't have as nice support for gestures as native so creating this hold down mouse  to scroll behavior isn't as straightforward but since it was so confusing I've decided to implement it anyway. 

So now you can either use trackpad left right motion or press down on the trackpad and drag it.

.playwright/tests/applyWrappingListSelection.spec.ts
pkaramon: No, I've removed them

src/web/formats/listKeyboard.ts
pkaramon: Good idea

.playwright/tests/listKeyboard.spec.ts
pkaramon: Nope, I've removed them

.playwright/tests/listKeyboard.spec.ts
pkaramon: I've extracted out a `KEY_ACTION_DELAY` constant

src/web/formats/listKeyboard.ts
pkaramon: `depth` is only 0 if "the position points directly into the root node". Normal text selections always have a `depth >= 1` because you're inside a paragraph, list item, etc. A depth of 0 happens during an entire document selection, or a `NodeSelection` on a top-level node (like selecting an image or paragraph from the as a single object).

apps/example-web/src/components/Toolbar.tsx
pkaramon: https://github.com/software-mansion/react-native-enriched/pull/560#issuecomment-4311465993

pkaramon:
@exploIF - and 1. shortcuts are in place

src/web/formats/EnrichedUnorderedList.ts
```diff
@@ -0,0 +1,40 @@
...
+  addKeyboardShortcuts() {
+    return {};
+  },
+
+  addCommands() {
+    return {
+      toggleBulletList:
+        () =>
+        ({ editor, commands, chain }) => {
+          if (editor.isActive('bulletList')) {
```
kacperzolkiewski: We don't want to give it name 'unorderedList' for consistency with iOS/Android?

kacperzolkiewski:
Works nice 👏🚀

src/web/formats/EnrichedUnorderedList.ts
pkaramon: Good idea. I've changed everywhere we've used 'bullet' to 'unordered'

kacperzolkiewski (approved):

exploIF (approved):

PR #561, opened 2026-04-22
feat(web): links

pkaramon:
# Summary

+ adds support for links on the web
+ **features missing: autodetecting links, `removeLink` and `onLinkDetected`** those will be in subsequent PRs
+ implemented `setLink`
+ wrote tests for the new element and for `setLink`
+ new test screen `TestLinks`
+ handled htmlStyle for links

## Test Plan

+ Run `yarn test:e2e:web` to see if all the new e2e tests pass.
+ Run `yarn example-web dev` and play around with links.

## Screenshots / Videos

image

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ❌     |
| Web |    ✅     |

## Checklist

- [x] E2E tests are passing
- [x]  Required E2E tests have been added (if applicable)


apps/example-web/src/components/Toolbar.tsx
```diff
@@ -148,6 +149,13 @@ export function Toolbar({ editorRef, state }: ToolbarProps) {
         editor?.toggleCodeBlock();
       },
     },
+    {
```
exploIF, line 154: It's not visible in my example app

apps/example-web/src/components/Toolbar.tsx
```diff
@@ -148,6 +149,13 @@ export function Toolbar({ editorRef, state }: ToolbarProps) {
...
+      key: 'link',
+      label: '🔗',
+      onPress: () => {
+        onOpenLinkModal();
+      },
```
exploIF:
```suggestion
      onPress: openLinkModal,
```

src/web/useOnChangeState.ts
```diff
@@ -46,40 +46,51 @@ function buildState(
...
-      isConflicting: false,
+      isConflicting,
       isBlocking: isFormatBlocked(tiptapName, editor, htmlStyle),
     };
   }
 
-  function paragraphFormat(isActive: boolean) {
+  function paragraphFormat(
+    isActive: boolean,
+    additionalIsConflicting: boolean = false
```
exploIF: I think type definition is redundant when passing default value

```suggestion
    additionalIsConflicting = false
```

apps/example-web/src/components/Toolbar.tsx
pkaramon: https://github.com/software-mansion/react-native-enriched/pull/560#issuecomment-4311465993

That fix was also cherry-picked here.

src/web/useOnChangeState.ts
```diff
@@ -46,40 +46,48 @@ function buildState(
...
     orderedList: paragraphFormat(false),
     unorderedList: paragraphFormat(false),
     checkboxList: paragraphFormat(false),
-    link: { isActive: false, isConflicting: false, isBlocking: false },
+    link: inlineFormat(
+      'link',
+      editor.isActive('code') ||
+        editor.isActive('link') ||
+        editor.isActive('enrichedCodeBlock')
+    ),
```
kacperzolkiewski, line 90: link is conflicting with link? 
`editor.isActive('link')`

kacperzolkiewski:
@pkaramon  It will be probably fixed when you will implement `onLinkDetected` but right now we can create link inside another link:

https://github.com/user-attachments/assets/e780b3a7-b1f8-42d2-927c-54a8e888efa0

src/web/useOnChangeState.ts
pkaramon: It is at least on iOS:
```objc++
+ (NSDictionary *)conflictMap {
  return @{
    @([BoldStyle getType]) : @[],
    @([ItalicStyle getType]) : @[],
    @([UnderlineStyle getType]) : @[],
    @([StrikethroughStyle getType]) : @[],
    @([InlineCodeStyle getType]) :
        @[ @([LinkStyle getType]), @([MentionStyle getType]) ],
    @([LinkStyle getType]) : @[
      @([InlineCodeStyle getType]), @([LinkStyle getType]),
      @([MentionStyle getType])
    ],
```

on Android, I've checked now, and there it's not. So I'm not sure which one to pick. I think in general when it comes to `isConflicting` and `isBlocking` we have quite a lot of differences between platforms.

pkaramon:
@kacperzolkiewski you're right that this is fixed by `onLinkDetected`, because with it we can make the link modal "smarter". But I feel like this is still a bug, we never should have a situation where we have a link inside of another. So I've updated `excludes` property of `EnrichedLink` to `link code`. That way tiptap will never produce output in which two links are nested, regardless of like the implementation of `setLink`, our modal etc.

src/web/pmPlugins/strictMarksPlugin.ts
```diff
@@ -18,7 +18,18 @@ function resolveStrictMarks(
 
   // Character to the left: strictly inherit from it
   if ($from.nodeBefore) {
-    return $from.nodeBefore.marks;
+    let marks = $from.nodeBefore.marks;
+    // Link is non-inclusive: do not carry the link onto text typed after the last
+    // linked character (e.g. Space must "exit" the link, matching native).
+    const linkType = newState.schema.marks.link;
+    if (linkType && linkType.isInSet(marks)) {
```
exploIF:
```suggestion
    if (linkType?.isInSet(marks)) {
```

exploIF (approved):

PR #565, opened 2026-04-23
feat(web): onLinkDetected

pkaramon:
# Summary

+ implemented `onLinkDetected` callback
+ wrote tests for it
+ changed the behavior of link modal, if we are currently on a link we fetch it's text and url into the modal

## Test Plan

+ Run `yarn test:e2e:web`
+ Play around with `onLinkDetected` in the example-web. Probably the easiest way is to open console and search for `onLinkDetected`

## Screenshots / Videos

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ❌     |
| Web |    ✅     |

## Checklist

- [x] E2E tests are passing


kacperzolkiewski:
I am not sure if it is connected with this PR but during first render I cannot add link to empty input.
I have to type something inside input to add link:

https://github.com/user-attachments/assets/46b6ef8c-25f8-4162-bcab-79d90bd959ef

pkaramon:
@kacperzolkiewski this is unrelated. We are not creating a link if we don't have a selection. This also appears on the native version, where if you try to insert a link without first focusing on the text input the link won't be inserted.

The difference is that on web `onSelectionUpdate` doesn't fire when we just focus on an empty input. Which causes `onChangeSelection` to not fire. I think this should be fixed but probably better in a separate PR, because it's related to selection and not links.

apps/example-web/src/App.tsx
```diff
@@ -30,8 +38,18 @@ function App() {
...
+    useState<OnLinkDetected>(DEFAULT_LINK_STATE);
   const [isLinkModalOpen, setIsLinkModalOpen] = useState(false);
 
+  const insideCurrentLink =
+    !!editorState?.link.isActive &&
+    currentLink.url.length > 0 &&
+    !!(currentLink.start || currentLink.end) &&
+    selection !== null &&
+    selection.start >= currentLink.start &&
+    selection.end <= currentLink.end;
```
exploIF: Could it be somehow simplified? Looks pretty complicated

exploIF (approved):

kacperzolkiewski (approved):

apps/example-web/src/App.tsx
pkaramon: I've actually copied this from the native version. I've now refactored this using helper variables.

PR #567, opened 2026-04-24
feat(android): add copy text as rich content to EnrichedText

kacperzolkiewski:
# Summary

 Add copying text as rich content to `EnrichedText`

## Test Plan

1. Add `selectable={true}` on `EnrichedText` component. 
2. Copy content from `EnrichedText`
3. Paste it inside `EnrichedTextInput`

The content pasted into `EnrichedTextInput` should be styled.

## Screenshots / Videos

https://github.com/user-attachments/assets/6f3d565e-abb9-4aae-b13f-cabe1de9901d

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

## Checklist

- [ ] E2E tests are passing

### DIFF
--- a/android/src/main/java/com/swmansion/enriched/text/EnrichedTextView.kt
+++ b/android/src/main/java/com/swmansion/enriched/text/EnrichedTextView.kt
@@ -1,9 +1,12 @@
 package com.swmansion.enriched.text
 
+import android.content.ClipData
+import android.content.ClipboardManager
 import android.content.Context
 import android.graphics.Color
 import android.graphics.text.LineBreaker
 import android.os.Build
+import android.text.Spannable
 import android.text.SpannableString
 import android.text.Spanned
 import android.text.TextUtils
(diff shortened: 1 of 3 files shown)


pkaramon:
The following bug wasn't introduced as part of this PR(I think at least):  when you have an image in a list it breaks the numbered list ordering, for unnumbered an unnecessary space is added. 

https://github.com/user-attachments/assets/1d2487d1-e0f0-44e3-af1b-4f61ff127400


But it also breaks copy and paste behavior:


https://github.com/user-attachments/assets/723cc480-fcf1-4cb1-bbd2-bfa86245f961

kacperzolkiewski:
@pkaramon I think it is general problem with image parsing in lists on Android. 
The bug occur even when we insert value to input by `Set Value` button:

https://github.com/user-attachments/assets/49b2b4b2-6800-49ea-8c0e-38df78157381


I am looking into it 👀

android/src/main/java/com/swmansion/enriched/text/EnrichedTextView.kt
```diff
@@ -131,6 +134,31 @@ class EnrichedTextView : AppCompatTextView {
...
+    if (start < end) {
+      val selectedText = spannable.subSequence(start, end) as Spannable
+      val selectedHtml = EnrichedParser.toHtml(selectedText)
+
+      val clipboard = context.getSystemService(Context.CLIPBOARD_SERVICE) as ClipboardManager
+      val clip = ClipData.newHtmlText(CLIPBOARD_TAG, selectedText, selectedHtml)
+      clipboard.setPrimaryClip(clip)
+    }
+  }
+
```
pkaramon, line 161: This is like a 1:1 copy from EnrichedTextInputView, maybe we could extract this common functionality? If we do so we should also extract `CLIPBOARD_TAG` as now it's duplicated in both files

kacperzolkiewski: That’s a good idea. I’m just not sure if we want to extract it. `EnrichedTextInput` and `EnrichedText` might need different behavior in the future, even if they currently share the same logic. @exploIF what do you think?

exploIF: I think it's fine to keep it separately, but I totally agree with moving `CLIPBOARD_TAG ` to shared object `EnrichedConstants`. Those two values have to be perfectly aligned between components

exploIF (approved):
Looks good!

Let's just move `CLIPBOARD_TAG` to `EnrichedConstants`

PR #569, opened 2026-04-24
fix(android): image parsing inside lists

kacperzolkiewski:
# Summary

Fixes: https://github.com/software-mansion/react-native-enriched/pull/567#issuecomment-4312086097

## Test Plan

Insert into the input:
```html
<html>
<ul>
<li><img src="abc" width="80" height="80"/></li>
</ul>
</html>
```
Or some other html `blockquote/codeblock/ol` but with empty lines containing only ``. The html should be parsed properly.

## Screenshots / Videos

https://github.com/user-attachments/assets/de0e5897-400e-4814-a6bc-b3eded72d145

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

## Checklist

- [X] E2E tests are passing


pkaramon:
I think it would be good to add an e2e test, as it's really simple in this case

pkaramon (approved):
LGTM, only I would add a tiny test for this just in case

exploIF (approved):

PR #575, opened 2026-04-27
feat(web): checkbox list

pkaramon:
# Summary

+ added checkboxList to the web version
+ handled htmlStyle for this new style
+ wrote e2e tests for checkboxList

I've used the `TaskList` and `TaskItem` extensions from `@tiptap/extension-list`. I
I handled compatibility with the native version by implementing mappings between Tiptap’s checkbox list HTML and the native format.

## Test Plan

+ Run `yarn test:e2e:web` to make sure all the e2e tests pass.
+ Run `yarn example-web dev` and play around with checkbox lists.

## Screenshots / Videos

image

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ❌     |
| Web |    ✅     |

## Checklist

- [x] E2E tests are passing


apps/example-web/src/components/Toolbar.tsx
```diff
@@ -170,6 +170,13 @@ export function Toolbar({ editorRef, state, onOpenLinkModal }: ToolbarProps) {
         editor?.toggleOrderedList();
       },
     },
+    {
+      key: 'checkboxList',
+      label: '☐',
```
exploIF: Super minor, could you update label to some more meaningful character/icon?

exploIF (approved):

apps/example-web/src/components/Toolbar.tsx
pkaramon: Changed to ☑

kacperzolkiewski:
On native, tapping a checkbox moves the cursor to the corresponding line. On web, this doesn’t happen. Should we make the web behavior consistent with native?

https://github.com/user-attachments/assets/3d0615eb-e5b6-420f-9e72-37f5f877e125



https://github.com/user-attachments/assets/54e7033e-d3cd-4249-ad53-ffe566cd59c4

pkaramon:
@kacperzolkiewski I personally prefer not moving the cursor when tapping a checkbox, but I don’t feel strongly about it. 

@exploIF what do you think?

exploIF:
@pkaramon @kacperzolkiewski I think we can leave it as it is for now. Changing cursor position on web is much easier, so I wouldn't say it's something crucial and important for the first iteration.

What I would like to have is an internal documentation with all platform inconsistencies, so we can revisit this topic later on and either document it or align between platforms

kacperzolkiewski (approved):

PR #576, opened 2026-04-28
feat(web): images

pkaramon:
# Summary

+ adds images to the web version 
+ added e2e tests for them
+ **what's missing**: pasting images and thus handling for `onPasteImages` callback.

**Differences compare to native**:
+ On Android, applying a link to a selection that includes an image behaves differently than it does on iOS. The web uses the iOS behavior in this case.

https://github.com/user-attachments/assets/494556be-4998-4a15-a14d-ae7a8f24875a

+ The cursor behaves a bit differently compared to native. On web when using the arrow keys we can enter into selection state that you can achieve on iOS by double clicking on the image. The native version "skips" over the image. Changing this would require writing low level plugins for ProseMirror which can often introduce subtle bugs and would be difficult to get right. That said, I'm open to changing this if it’s considered crucial.

https://github.com/user-attachments/assets/91e6ec3a-a6a6-4f53-86fe-8d827cefa734

## Test Plan

+ Run `yarn test:e2e:web` make sure all the new e2e tests pass.
+ Run `yarn example-web dev` and play around with images.

## Screenshots / Videos

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ❌     |
| Web |    ✅     |

## Checklist

- [x] E2E tests are passing


src/web/formats/EnrichedImageNodeView.tsx
```diff
@@ -0,0 +1,97 @@
...
+  let imgStyle: CSSProperties | undefined;
+  if (rawW != null && rawH != null) {
+    imgStyle = undefined;
+  } else if (rawW != null) {
+    imgStyle = { height: 'auto' };
+  } else if (rawH != null) {
+    imgStyle = { width: 'auto' };
+  } else {
+    imgStyle = { height: 'auto' };
+  }
```
kacperzolkiewski, line 73: We can simplify those two ifs into one:
```ts
let imgDims: { width?: number; height?: number };
let imgStyle: CSSProperties | undefined;

if (rawW != null && rawH != null) {
  imgDims = { width: rawW, height: rawH };
  imgStyle = undefined;
} else if (rawH != null) {
  imgDims = { height: rawH };
  imgStyle = { width: 'auto' };
} else {
  imgDims = { width: rawW ?? IMAGE_FALLBACK_SIZE };
  imgStyle = { height: 'auto' };
}
```

kacperzolkiewski:
Works well, but as we discussed on Slack, I think list markers should align with the text baseline.
image

Native:
image

pkaramon:
@kacperzolkiewski the image placement in lists ought to be resolved. Can you check it again?

kacperzolkiewski (approved):

exploIF (approved):

exploIF:
Please update web API docs

PR #580, opened 2026-04-29
fix(android): inline styles line overlapping

### LINKED ISSUE #511: [Android]: bold formatting line overlapping
**Describe the bug**

Applying bold styling after inserting new lines causes the existing text to overlap. Specifically, the first paragraph (test) becomes visually overlapped or corrupted after formatting changes.

**To Reproduce**
1. Open the app with initial content like 
```
<html><p>test</p><p>Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum has been the industry standard dummy text ever since the 1500s, when an unknown printer took a galley of type and scrambled it to make a type specimen book. It has survived not only five centuries, but also the leap into electronic typesetting, remaining essentially unchanged. It was popularised in the 1960s with the release of Letraset sheets containing Lorem Ipsum passages, and more recently with desktop publishing software like Aldus PageMaker, including versions of Lorem Ipsum.</p></html>
```
2. Tap on the first paragraph
3. Insert a new line
4. Type any bold text
5. Insert new lines
6. Observe that `test` text got overlapped

**Expected behavior**

Bold style should not overlap the text

**Screenshots**

https://github.com/user-attachments/assets/50ce26ae-b477-4d6d-9675-2748bd249bae

**Additional context**

kacperzolkiewski:
# Summary

Fixes: #511 

Android has some issues with layout calculation when applying inline styles (e.g. bold, inline code or italic), which are `MetricAffectingSpans`, on new lines. To prevent line overlapping, we manually trigger a re-layout for those spans. 
This is the same approach we use for headings.

## Test Plan

Run reproduction steps from #511 
Issue should be resolved

## Screenshots / Videos

https://github.com/user-attachments/assets/c8278ed2-6309-4820-990e-7b43df730230

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

## Checklist

- [X] E2E tests are passing


pkaramon (approved):

android/src/main/java/com/swmansion/enriched/textinput/spans/interfaces/EnrichedInlineMetricSpan.kt
```diff
@@ -0,0 +1,8 @@
+package com.swmansion.enriched.textinput.spans.interfaces
+
+/**
+ * Interface for inline spans that extend MetricAffectingSpan
+ * (e.g. bold, italic, inline code). This interface is used by EnrichedSpanWatcher to
+ * detect spans and trigger the necessary layout invalidation.
+ */
+interface EnrichedInlineMetricSpan
```
exploIF: I don't like it. If we create a new inline style that extends `MetricAffectingSpan` we will have exactly the same issue  (until we extend this interface).

Instead, I would prefer to do a check that looks like:
`if (span is EnrichedInlineSpan && span is MetricAffectingSpan)`. That way we don't have to remember about adding this interface every time we add/modify inline span

kacperzolkiewski: Fixed

exploIF (approved):
Thank you!

PR #582, opened 2026-04-29
chore: expose web and switch .tsx -> .native.tsx, .web.tsx -> .tsx

pkaramon:
# Summary

+ switched `index.tsx` -> `index.native.tsx`, `index.web.tsx` -> `index.tsx`
+ exposed web, meaning we no longer filter out web files when publishing to the npm

## Test Plan

Run `example` on iOS and Android run `example-web` see if everything compiles and runs.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅❌     |
| Android |    ✅❌     |

## Checklist

- [ ] E2E tests are passing


exploIF:
I've noticed one issue in the example app. When I run `yarn clean` to clean repo, then `yarn install` & `yarn start` & `yarn android`, metro tries to resolve `index` instead of `index.native`. However, when I run `yarn prepare` the issue is gone. Ideally I would like to have it working without a need to build JS files

pkaramon:
@exploIF I've added `"source": "./src/index.native.tsx"` to `"react-native"` section in `"exports"`, so the bundler can resolve to native code even if no `lib` directory exists. Does that approach look okay to you?

kacperzolkiewski (approved):

exploIF (approved):

PR #584, opened 2026-04-30
feat(web): image paste

pkaramon:
# Summary

+ implemented image pasting on web
+ handled `onPasteImages` callback

The reason for blob urls is that it allows us to keep the same type for `OnPasteImagesEvent`, now `uri` is just a blob url.
The other alternative was base64 encoded urls but they do not work on Android. The caveat is that we generally should clean such files by doing `URL.revokeObjectURL()`(even if someone forgets to call that, those temporary files will still be deleted it's just that they may be in the memory for a bit longer than absolutely needed).  I generally think that we cannot really know the best moment to call this ourselves, only our users do. I initially thought that when the component unmounts is a good time to remove those objects, but for example: an upload to a server might have failed and a retry is happening, if our component unmounted and deleted those temporary files the application code would no longer have an image to upload.

Curious to see what you think

## Test Plan

+ Try out image pasting

## Screenshots / Videos

https://github.com/user-attachments/assets/1659b08c-0b3e-4ed6-8c83-11f074ac021c

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ❌     |
| Web |    ✅     |

## Checklist

- [x] E2E tests are passing


kacperzolkiewski (approved):
Looks and works fine 🚀

src/web/EnrichedTextInput.tsx
```diff
@@ -124,6 +125,11 @@ export const EnrichedTextInput = ({
     htmlStyleRef.current = resolvedHtmlStyle;
   }, [resolvedHtmlStyle]);
 
+  const onPasteImagesRef = useRef(onPasteImages);
+  useEffect(() => {
+    onPasteImagesRef.current = onPasteImages;
+  }, [onPasteImages]);
```
exploIF, line 131: Not related to your PR as we use this approach in other places as well, so we can keep it as it is for now. 

However, I'm not the biggest fan of this approach, maybe there is a better way to prevent unnecessary re-renders. Or we can create a reusable helper hook, instead of repeating this approach over and over

exploIF (approved):

src/web/EnrichedTextInput.tsx
kacperzolkiewski: I've just added this to our Linear

PR #594, opened 2026-05-06
feat(web): mentions

pkaramon:
# Summary
+  **what is missing**: tests, mention styling. Those will be separate PRs as this one is already big enough.
+ adds mentions to web version of the library
+ adds controls in the example-web app for mentions
+ implemented all ref functions related to mentions: `setMention` and `startMention`
+ implemented all props related to mentions: `onMentionDetected`, `onStartMention`, `onChangeMention`, `onEndMention`
+ mentions are created as a custom mark, with fully custom state tracking and logic for: removing marks when a single character changes, striping partial mentions during paste etc. All of this custom logic lives in `pmPlugins/mentionPlugin`

## Test Plan

+ Run example app and test out mentions you can also verify logs related to mentions in the console

## Screenshots / Videos

https://github.com/user-attachments/assets/b9028481-4f46-4307-b2a3-1eeeab7c220f

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ❌     |
|  Web  |    ✅     |

## Checklist

- [x] E2E tests are passing


exploIF:
What does the abbreviation pmPlugins stand for?

src/web/EnrichedTextInput.tsx
```diff
@@ -104,37 +117,73 @@ export const EnrichedTextInput = ({
     []
   );
 
+  const mentionIndicatorsRef = useRef(mentionIndicators);
```
exploIF, line 121: Why does it have to be stored as a ref? I'm afraid that storing too much inside ref just to prevent re-renders is not a reliable solution

pkaramon: It's stored as a ref, because updating extensions(passing new/updated data to them) in a react way is not really possible because, if we added `mentionIndicators` to dependency array of `extensions` then the `useEditor` would recreate the entire editor when `mentionIndicators` would change resulting in losing content, selection, focus, history etc. So the ref is there so that the custom plugin can read the most recent value without needing to recreate the editor which is slow and also makes it pretty much unusable

pkaramon (commented):
`pmPlugins` stands for ProseMirror(the underlying document engine that TipTap is built on) plugins. This is where we keep low-level logic, for example stripping all marks(bold, italic etc.) from codeblocks.

kacperzolkiewski (approved):
Works properly 🚀

exploIF (approved):

PR #595, opened 2026-05-08
feat(web): add styles for mentions

pkaramon:
# Summary

+ added handling of `htmlStyle` for mentions
+ Styling for mentions is a bit different than for the rest of the styles, because we cannot write out all the css rules in `EnrichedTextInput.css`(as we do not know all of the indicators up front). So like for the rest of the styles we still create css vars, in this case e.g. `--eti-mention-u0040-color` in this case `u0040` is escaped version of `@` as we cannot use `@` in a css variable. We also generate css rules dynamically, we always generate a rule `.eti-editor mention` for the default case, which works for any `<mention>` tag, and rules of form `.edit mention[indicator='@']` for specific indicators. Those specific css rules overwrite the default/general one. Finally we render those new css rules in a `<style>` tag.

## Test Plan

+ Launch example-web app
+ Play around with `apps/example-web/src/defaultHtmlStyle.ts` to see how the mention styling changes

## Screenshots / Videos

image
## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ❌     |
| Web |    ✅     |

## Checklist

- [x] E2E tests are passing

### DIFF
--- a/apps/example-web/src/defaultHtmlStyle.ts
+++ b/apps/example-web/src/defaultHtmlStyle.ts
@@ -62,4 +62,16 @@ export const WEB_DEFAULT_HTML_STYLE: HtmlStyle = {
     marginLeft: 24,
     boxColor: 'rgb(0, 26, 114)',
   },
+  mention: {
+    '@': {
+      backgroundColor: 'lightgreen',
+      color: 'green',
+      textDecorationLine: 'none',
+    },
+    '#': {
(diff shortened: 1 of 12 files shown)


exploIF (approved):

kacperzolkiewski (approved):
LGTM

PR #598, opened 2026-05-11
feat(web): automatic links

pkaramon:
# Summary

+ implemented automatic links on web
+ handled `linkRegex` prop
+ wrote e2e tests for automatic links
+ refactored `useOnLinkDetected` so it uses a shared `LinkEmitter` with autolink plugin, because we want to emit an  event when an automatic link is created, but we also do no want to send the same event twice.
+ this PR doesn't use automatic link functionality from `Link` extensions from Tiptap, as it doesn't allow passing any regex, and it also struggles with extending automatic links on the right. I've implemented a custom plugin that looks at changed rages and research using a regex for texts where we should applied the automatic link mark.

## Test Plan

+ Run `yarn test:e2e:web`  to see if all the tests pass
+ Run `yarn example-web dev` and play around with automatic links, change `LINK_REGEX` in `apps/example-web/App.tsx`

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ❌     |
| Web |    ✅     |

## Checklist

- [x] E2E tests are passing


kacperzolkiewski (commented):
When we type swmansion.com, in HTML I can see that https:// is automatically added to the href. Is this the intended behavior?

https://github.com/user-attachments/assets/377a1843-33db-41e9-9d93-a62af3eab696

kacperzolkiewski:
When we paste text containing a link, it is treated as a manually added link, which allows inserting new lines inside it:

https://github.com/user-attachments/assets/bac4cb79-4352-415a-90f1-e49b4d8d1a06

hejsztynx:
> When we type swmansion.com, in HTML I can see that https:// is automatically added to the href. Is this the intended behavior?
> 
>  Screen.Recording.2026-05-26.at.09.02.49.mov

removed automatic `https` prefix addition

hejsztynx:
> When we paste text containing a link, it is treated as a manually added link, which allows inserting new lines inside it:
> 
>  Screen.Recording.2026-05-26.at.09.42.49.mov

now it checks if `href` equals to the `text` value of the html node - if so, we treat it as an autolink, just like in the native

src/web/pmPlugins/AutolinkPlugin/index.ts
```diff
@@ -0,0 +1,222 @@
...
+
+    const ranges = findAutolinkRangesInWord(word, linkRegex);
+    const fullMatch = ranges.some(
+      (r) => r.start === 0 && r.endExclusive === word.length
+    );
+
+    if (!fullMatch) continue;
+    if (rangeHasManualLink(doc, linkType, wordStart, wordEnd)) continue;
+
+    // const href = prepareUrl(word);
```
kacperzolkiewski: To be removed

kacperzolkiewski (approved):
Works good!

exploIF:
It looks to me like `linkRegex` prop is not working properly. Whenever I paste some valid URL, it is recognized as a link, even if I pass `linkRegex={null}`

apps/example-web/src/testScreens/TestLinks.tsx
```diff
@@ -12,10 +12,20 @@ function toInteger(value: string): number {
...
   const ref = useRef<EnrichedTextInputInstance>(null);
   const [htmlInput, setHtmlInput] = useState('<html><p></p></html>');
   const [editorHtml, setEditorHtml] = useState('');
+  const [linkRegexMode, setLinkRegexMode] = useState<LinkRegexMode>('default');
+  const [linkRegexPattern, setLinkRegexPattern] = useState(
+    String.raw`issue-\d+`
+  );
+  const [appliedLinkRegex, setAppliedLinkRegex] = useState<
+    RegExp | null | undefined
+  >(undefined);
```
exploIF:
```suggestion
  const [appliedLinkRegex, setAppliedLinkRegex] = useState<
    RegExp | null
  >();
```

src/web/pmPlugins/AutolinkPlugin/autolinkRegex.ts
```diff
@@ -0,0 +1,44 @@
...
+
+// URL-like substrings inside a single whitespace-delimited token (`\S+`),
+export function findAutolinkRangesInWord(
+  word: string,
+  linkRegex: RegExp | undefined
+): readonly AutolinkRangeInWord[] {
+  const re =
+    linkRegex === undefined
+      ? new RegExp(DEFAULT_AUTOLINK_SUBSTRING_PATTERN, 'gi')
+      : asGlobalRegex(linkRegex);
```
exploIF: Isn't it the same as?
```suggestion
  const re =
    linkRegex
      ? asGlobalRegex(linkRegex)
      : new RegExp(DEFAULT_AUTOLINK_SUBSTRING_PATTERN, 'gi');
```

src/web/pmPlugins/AutolinkPlugin/index.ts
```diff
@@ -0,0 +1,221 @@
...
+  to: number
+): boolean {
+  let found = false;
+
+  doc.nodesBetween(from, to, (node) => {
+    if (found) return false;
+    if (!node.isText) return true;
+
+    const link = linkType.isInSet(node.marks);
+    if (link && link.attrs.auto !== true) found = true;
```
exploIF, line 67: Will this work?
```suggestion
    if (!link?.attrs.auto) found = true;
```

src/web/pmPlugins/AutolinkPlugin/index.ts
```diff
@@ -0,0 +1,221 @@
...
+  from: number,
+  to: number
+): void {
+  if (from >= to) return;
+
+  doc.nodesBetween(from, to, (node, pos) => {
+    if (!node.isText) return true;
+
+    const link = linkType.isInSet(node.marks);
+    if (link?.attrs.auto === true) {
```
exploIF, line 42: Will this work?
```suggestion
    if (!link?.attrs.auto) found = true;
```

exploIF (changes requested):

src/web/pmPlugins/AutolinkPlugin/index.ts
hejsztynx: I assume that suggestion was supposed to refer to the previous comment

src/web/pmPlugins/AutolinkPlugin/index.ts
hejsztynx: not really, now the condition would pass when `link` is `undefined`, which happens when we don't find a link mark at given range

src/web/pmPlugins/AutolinkPlugin/index.ts
```diff
@@ -0,0 +1,221 @@
...
+    };
+  },
+  addProseMirrorPlugins() {
+    return [
+      new Plugin({
+        key: new PluginKey('autolinkDetector'),
+        appendTransaction: (transactions, _oldState, newState) => {
+          const state = this.options.getLinkEmitter();
+          if (!state || state.linkRegex === null) return null;
+
```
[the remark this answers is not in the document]
hejsztynx: I checked, the transaction don't re-loop by itself, it's managed internally by ProseMirror

src/web/formats/EnrichedLink.ts
```diff
@@ -8,6 +8,18 @@ import { isLinkBlocked } from './formatRules';
...
 
+  addAttributes() {
+    return {
+      ...this.parent?.(),
+      auto: {
+        default: false,
+        parseHTML: (el) => el.getAttribute('href') === el.textContent,
+        // Internal only
+        renderHTML: () => ({}),
+      },
```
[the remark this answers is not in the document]
hejsztynx: Now there is no such check, we persist the information if the link is manual or auto with a custom attribute `data-auto`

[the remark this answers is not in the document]
hejsztynx: Changed the approach to the previous one, we compare `href == textContent` but also we check if it would be recognized be the `linkRegex`. If so, we treat it as an autolink, otherwise as manual one

exploIF:
Functionally works well, however there are failing e2e tests:
```
[chromium] › .playwright/tests/codeblockInlineStyles.spec.ts:60:7 › codeblock inline styles › paste into code block strips copied inline styles 
[chromium] › .playwright/tests/headingBoldFromStyle.spec.ts:45:7 › h1 bold from htmlStyle (h1.bold true) › paste into h1 strips copied bold mark 
[chromium] › .playwright/tests/headingBoldFromStyle.spec.ts:100:7 › h1 bold from htmlStyle (h1.bold false) › paste into h1 keeps copied bold mark when htmlStyle h1.bold is false
```

exploIF (approved):

PR #602, opened 2026-05-13
feat(android): text alignment

kacperzolkiewski:
# Summary

Adds text alignment for Android
*(Note: `justify` alignment is not natively supported for single spans on Android and will not work).*

## Test Plan
Run example app and play with `textAlignment` when different styles are toggled.
Check if paragraphs/lists with different `textAlignment` than auto are properly parsed and contain `style="text-align:center|right"` inside HTML

## Screenshots / Videos

https://github.com/user-attachments/assets/089543ad-93e4-4842-8b87-95316a3cb747

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

## Checklist

- [x] E2E tests are passing


android/src/main/java/com/swmansion/enriched/textinput/styles/AlignmentStyles.kt
```diff
@@ -0,0 +1,348 @@
...
+import com.swmansion.enriched.textinput.EnrichedTextInputView
+import com.swmansion.enriched.textinput.spans.EnrichedInputAlignmentSpan
+import com.swmansion.enriched.textinput.spans.EnrichedInputCheckboxListSpan
+import com.swmansion.enriched.textinput.spans.EnrichedInputOrderedListSpan
+import com.swmansion.enriched.textinput.spans.EnrichedInputUnorderedListSpan
+import com.swmansion.enriched.textinput.utils.getParagraphBounds
+import com.swmansion.enriched.textinput.utils.getSafeSpanBoundaries
+import com.swmansion.enriched.textinput.utils.safelyInsertZWS
+
+class AlignmentStyles(
```
exploIF, line 16: Why do we need to have a separate abstraction for alignement? Cannot we reuse what's implemented for the paragraph spans?

android/src/main/java/com/swmansion/enriched/textinput/styles/ListStyles.kt
exploIF, line 1: Why do we need to modify logic inside list styles?

android/src/main/java/com/swmansion/enriched/textinput/utils/EnrichedSpannableStringBuilder.kt
```diff
@@ -22,3 +27,35 @@ fun SpannableStringBuilder.removeZWS(
...
+/**
+ * Removes ZWS in [start, end) only if no alignment still requires them as a layout anchor.
+ */
+fun SpannableStringBuilder.safelyRemoveZWS(
+  start: Int,
+  end: Int,
+) {
+  val safeStart = start.coerceAtLeast(0)
+  val safeEnd = end.coerceAtMost(length)
+  if (safeStart >= safeEnd) return
```
exploIF: In such scenario, shouldn't we use `start` as `end` and the other way around?

android/src/main/java/com/swmansion/enriched/textinput/styles/ListStyles.kt
kacperzolkiewski: The only behavior we modify in list styles is the handling of ZWS insertion and removal.
We need this because there may already be a ZWS present in a paragraph due to an applied alignment style. In that case, when applying a list style, we don't want to insert another ZWS, we should reuse the existing one.

Similarly, when removing a list style, we don't want to remove the ZWS if it is still needed by alignment style applied to that paragraph.

android/src/main/java/com/swmansion/enriched/textinput/styles/AlignmentStyles.kt
kacperzolkiewski: I separated the alignment logic into `AlignmentStyles` because it differs from other paragraph styles in several ways:

1. When pressing Backspace on an empty aligned paragraph that also has a paragraph/list style applied, we need to preserve the alignment before the ZWS is deleted (`anchorAlignmentToRestore`) so it can be restored after the paragraph/list style is removed.
3. When deleting a ZWS that only contains alignment, we don't simply remove the span for that line. Instead, we move the cursor back to the previous line.
3. Alignment is not a toggleable style, it is a multi-value property (left, center, right, auto).
4. When applied to a selection containing list items, alignment expands to the entire contiguous list rather than affecting only the selected items.

android/src/main/java/com/swmansion/enriched/textinput/utils/EnrichedSpannableStringBuilder.kt
kacperzolkiewski: Fixed, I use `coerceIn` and return only when the range is empty'

exploIF:
What's the plan for implementing `justify` on Android? There is no ready to use span for it, but we should find a different solution and bring justify support as it's pretty base alignement

android/src/main/java/com/swmansion/enriched/common/parser/EnrichedParser.java
```diff
@@ -448,9 +482,23 @@ public Spanned convert() {
       int end = mSpannableStringBuilder.getSpanEnd(zeroWidthSpaceSpan);
 
       if (mSpannableStringBuilder.charAt(start) != EnrichedConstants.ZWS) {
-        // Insert zero-width space character at the start if it's not already present.
+        // Collect spans before inserting ZWS. SPAN_EXCLUSIVE_EXCLUSIVE spans will
+        // shift to start+1. We must re-anchor them back to `start` to prevent
+        // the loop from processing them again and inserting duplicate ZWS.
+        Object[] colocated = mSpannableStringBuilder.getSpans(start, start + 1, Object.class);
```
exploIF: Should we iterate over every span or just `EnrichedSpan`?

android/src/main/java/com/swmansion/enriched/textinput/utils/EnrichedSpanState.kt
```diff
@@ -52,6 +52,13 @@ class EnrichedSpanState(
     private set
   var mentionStart: Int? = null
     private set
+  var currentAlignment: String = "auto"
+    private set
+
+  fun setAlignment(value: String) {
```
exploIF: Minor, could you move this function closer to the paragraph related styles?

android/src/main/java/com/swmansion/enriched/textinput/EnrichedTextInputView.kt
```diff
@@ -141,6 +143,16 @@ class EnrichedTextInputView :
...
   }
 
+  override fun scrollTo(
+    x: Int,
+    y: Int,
+  ) {
+    // Android's internal cursor tracker gets confused by ALIGN_CENTER + LeadingMarginSpan
+    // and attempts to scroll the text horizontally.
+    // We lock the horizontal scroll to 0 to prevent the view from shifting.
+    super.scrollTo(0, y)
```
exploIF, line 168: This means that horizontal scrolling is disabled, is it currently supported?

android/src/main/java/com/swmansion/enriched/textinput/EnrichedTextInputView.kt
```diff
@@ -962,6 +974,14 @@ class EnrichedTextInputView :
     parametrizedStyles?.setMentionSpan(text, indicator, attributes)
   }
 
+  fun setTextAlignment(alignment: String) {
+    runAsATransaction {
+      alignmentStyles?.setAlignment(alignment)
+    }
+    selection?.validateStyles()
+    layoutManager.invalidateLayout()
```
exploIF: Why it's needed?

android/src/main/java/com/swmansion/enriched/textinput/EnrichedTextInputView.kt
kacperzolkiewski: No it wasn't supported in our input, because during preparing component we set:
```
isSingleLine = false
isHorizontalScrollBarEnabled = false
isVerticalScrollBarEnabled = true
```
So our input is always multiline as I understand it correctly.

I needed to lock the horizontal scroll because in lists/blockquotes even that the scroll was disabled I had something like this:


https://github.com/user-attachments/assets/3ff8fc6a-7d1b-4525-9ace-1a13aeda3652

android/src/main/java/com/swmansion/enriched/common/parser/EnrichedParser.java
kacperzolkiewski: Yes, you're right, we should only iterate over `EnrichedSpan`, fixed

android/src/main/java/com/swmansion/enriched/textinput/utils/EnrichedSpanState.kt
kacperzolkiewski: Fixed

android/src/main/java/com/swmansion/enriched/textinput/EnrichedTextInputView.kt
kacperzolkiewski: Yes, I tested this once again without `layoutManager.invalidateLayout` and it works properly. So deleted

android/src/main/java/com/swmansion/enriched/common/spans/EnrichedAlignmentSpan.kt
```diff
@@ -0,0 +1,19 @@
+package com.swmansion.enriched.common.spans
+
+import android.text.Layout
+import android.text.style.AlignmentSpan
+import com.swmansion.enriched.common.spans.interfaces.EnrichedZeroWidthSpaceSpan
+
+open class EnrichedAlignmentSpan(
+  val cssValue: String,
+) : AlignmentSpan.Standard(cssValueToLayoutAlignment(cssValue)),
+  EnrichedZeroWidthSpaceSpan {
```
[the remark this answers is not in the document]
kacperzolkiewski: But inside `EnrichedParser.java` we use `EnrichedTextSpanFactory` or `EnrichedTextInputSpannableFactory` that when creating alignment span will return `EnrichedInputAlignmentSpan` or `EnrichedTextAlignmentSpan` and those spans inherit from `EnrichedSpan` so parser will catch them properly.

exploIF (approved):
Looks good to me

hejsztynx:
Currently there's a bug - in a flow:
1. apply some alignment to a base text
2. apply a paragraph style

Alignment is not visible unless it is pressed again on the toolbar or we write something in the input


https://github.com/user-attachments/assets/ea92e787-a9ab-49b6-8933-1defe262cca8

kacperzolkiewski:
> Currently there's a bug - in a flow:
> 
> 1. apply some alignment to a base text
> 2. apply a paragraph style
> 
> Alignment is not visible unless it is pressed again on the toolbar or we write something in the input
> 
>  alignment_bug.mov

fixed

hejsztynx:
In this flow the cursor behaves weird


https://github.com/user-attachments/assets/838bfdde-cf45-4d4a-b371-2b5a9325e92d

hejsztynx:
When aligned, the ordered list isn't *merged* immediately after deleting a row. Modifying a row fixes that

https://github.com/user-attachments/assets/8e45f72f-7bd6-4f94-ae14-79d30b9eec9c

hejsztynx:
When switching from `link` to `codeblock`, the lose the alignment similarly as in the first bug i mentioned


https://github.com/user-attachments/assets/11a9b119-7eaf-4a3c-bd0b-c09e5647d609

hejsztynx:
Works great 🚀

hejsztynx (approved):

PR #603, opened 2026-05-14
feat(web): added shortcuts

pkaramon:
# Summary

- added shortcuts for styles
- shortcuts are inspired by Slack for bold, italic etc. and by google docs when it comes to headings, lists
- I've verified on my Windows laptop that on it shortcuts also work
- `WEB.md` contains a detailed list of all shortcuts

## Test Plan

+ Run `yarn test:e2e:web` to verify new e2e tests pass
+ Run `yarn example-web dev` and play around with shortcuts
 
## Screenshots / Videos

N/A

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌   |
| Android |    ❌    |
| Web     |   ✅    |

## Checklist

- [x] E2E tests are passing

### DIFF
--- a/src/web/EnrichedTextInput.tsx
+++ b/src/web/EnrichedTextInput.tsx
@@ -70,6 +70,7 @@ import {
 } from './pmPlugins/mentionPlugin';
 
 import { StripMarksOnImagePlugin } from './pmPlugins/stripMarksOnImagePlugin';
+import { ShortcutManager } from './pmPlugins/shortcutManager';
 function runFocused(
   editor: Editor,
   apply: (chain: ChainedCommands) => ChainedCommands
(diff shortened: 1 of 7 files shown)


src/web/EnrichedTextInput.tsx
```diff
@@ -178,12 +187,18 @@ export const EnrichedTextInput = ({
       MergeAdjacentSameKindBlocksPlugin,
       StrictMarksPlugin,
       mentionPlugin,
+      shortcutManager,
```
exploIF: Maybe we could use the same naming convention for every extension/plugin? Currently `tiptap` provided plugins use PascalCase, but custom extensions use camelCase, how about unifying them?

Also, maybe instead of `shortcutManager`  we could have `ShortcutPlugin`?

exploIF (approved):
Works super good 🚀

src/web/EnrichedTextInput.tsx
pkaramon: I've renamed shortcutManager to shortcutPlugin. I agree about the naming convention, but I'd prefer to do it in a separate PR as autolinks introduce also them and they haven't been merged.

kacperzolkiewski (approved):

PR #600, opened 2026-05-11
feat(web): submit props

pkaramon:
# Summary

+ implemented handling of `submitBehavior`, `onSubmitEditing` and partially `returnKeyType`(only a portion of options work, https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/enterkeyhint)
+ `returnKeyLabel` is not implemented as browser APIs do not allow such modifications
+ wrote tests for new props, and new callback
 
## Test Plan

+ Launch example-web app.
+ Play around with submit props in `apps/example-web/App.tsx`.
+ Run `yarn test:e2e:web` to see if new e2e tests pass.
+ To test out `returnKeyLabel` you can run the example-web inside simulator's/emulator's browsers

## Screenshots / Videos

`returnKeyType` in action: 

Screenshot 2026-05-11 at 12 27 23

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ❌     |
| Web |    ✅     |

## Checklist

- [x] E2E tests are passing
- [x] Required E2E tests have been added (if applicable)

PR #605, opened 2026-05-14
docs: updated INPUT_API_REFERENCE.md to acknowledge web

pkaramon:
# Summary
- only review once the PRs below are fully merged
- Replaces all `Both` platform labels with explicit lists, so it's future-proof for macOS support
- Adds an experimental web support notice at the top with a note that `EnrichedText` is coming to web
- Documents three open PRs as if they were merged:
  - **#584** `feat(web): image paste`
  - **#600** `feat(web): submit props`
  - **#603** `feat(web): shortcuts`
## Test Plan

N/A

## Screenshots / Videos

N/A

## Compatibility

N/A 

## Checklist

N/A


docs/INPUT_API_REFERENCE.md
```diff
@@ -429,54 +449,96 @@ export interface OnPasteImagesEvent {
...
+| ---------------------------------------------- | ------------- | ----------------- |
+| `RefObject<EnrichedTextInputInstance \| null>` | -             | iOS, Android, Web |
+
+### `returnKeyLabel`
+
+Overrides the return key label with a custom string. Use `returnKeyType` for cross-platform return key control.
+
+| Type     | Default Value | Platform     |
+| -------- | ------------- | ------------ |
+| `string` | -             | iOS, Android |
```
kacperzolkiewski: Let's add some (Android only) note to this description, since iOS keyboards don't support custom return key labels.
https://reactnative.dev/docs/textinput#returnkeylabel-android

docs/INPUT_API_REFERENCE.md
```diff
@@ -429,54 +449,96 @@ export interface OnPasteImagesEvent {
...
+
+### `returnKeyType`
+
+Specifies the label or icon shown on the keyboard's return key.
+
+Accepts the standard React Native `ReturnKeyTypeOptions` values: `'default' | 'go' | 'google' | 'join' | 'next' | 'route' | 'search' | 'send' | 'yahoo' | 'done' | 'emergency-call'`.
```
kacperzolkiewski: Maybe it would be good to add info like in RN docs:
https://reactnative.dev/docs/textinput#returnkeytype

That some of values are cross platform, iOS only and Android only?

docs/INPUT_API_REFERENCE.md
```diff
@@ -443,54 +463,115 @@ export interface OnPasteImagesEvent {
...
+
+### `returnKeyType`
+
+Specifies the label or icon shown on the keyboard's return key.
+
+Is not supported on Android, as `returnKeyType` doesn't work with multiline inputs.
+
+Accepts the standard React Native `ReturnKeyTypeOptions` values: `'go' | 'next' | 'search' | 'send' | 'done' | 'default' | 'google' | 'join' | 'route' | 'yahoo' | 'emergency-call' | 'previous' | 'none'`.
+
+| Type                   | Platform          |
```
kacperzolkiewski: I'm not sure we need this table at the moment. We've already stated that this feature isn't supported on Android, but the table suggests that some types are supported there, which could be confusing and send mixed signals to readers.

docs/INPUT_API_REFERENCE.md
```diff
@@ -503,9 +584,9 @@ If true, Android will use experimental synchronous events. This will prevent fro
 
 If true, external HTML pasted/inserted into the input (e.g. from Google Docs, Word, or web pages) will be normalized into the canonical tag subset that the enriched parser understands. However, this is an experimental feature, which has not been thoroughly tested. We may decide to enable it by default in a future release.
 
-| Type   | Default Value | Platform |
-| ------ | ------------- |----------|
-| `bool` | `false`       | Both     |
+| Type   | Default Value | Platform          |
+| ------ | ------------- | ------------      |
+| `bool` | `false`       | iOS, Android, Web |
```
kacperzolkiewski: I would remove the Web row from this table. These docs are intended for the 0.8.0 release, whereas `useHtmlNormalizer` will only be available in a later release. Including it here could be confusing for users who expect the feature to be available in 0.8.0.
You can update the documentation to include Web support in the `useHtmlNormalizer` PR instead.

docs/INPUT_API_REFERENCE.md
hejsztynx: We allow all native `ReturnKeyTypeOptions` values, so I see this table as a reference which values are actually supported internally by the library. Without it a user will have no idea which `returnKeyType` values are android-only and will not work at all.

docs/INPUT_API_REFERENCE.md
```diff
@@ -99,27 +101,27 @@ If `false`, text is not editable.
...
 A custom regex pattern for detecting links in the input. If not provided, a default regex will be used.
 With this approach you can customize what patterns should be recognized as links, for example you can make it so that only links starting with `https://` are detected, or you can support custom schemes.
 Keep in mind that not all JS regex features are supported, for example variable-width lookbehinds won't work.
 
-| Type     | Default Value                 | Platform |
-|----------|-------------------------------|----------|
-| `RegExp` | default native platform regex | Both     |
+| Type     | Default Value                 | Platform          |
+| -------- | ----------------------------- | ----------------- |
+| `RegExp` | default native platform regex | iOS, Android, Web |
```
exploIF: It's not true actually, it will be true once https://github.com/software-mansion/react-native-enriched/pull/598 is merged

docs/TEXT_API_REFERENCE.md
```diff
@@ -1,5 +1,7 @@
 # EnrichedText API Reference
 
+> **Web is not supported.** `EnrichedText` is only available on iOS and Android.
```
exploIF:
```suggestion
> **Web is not yet supported.** `EnrichedText` is only available on iOS and Android.
```

docs/INPUT_API_REFERENCE.md
szydlovsky: Both doesn't work with 3 platforms 😄

docs/INPUT_API_REFERENCE.md
```diff
@@ -503,9 +584,9 @@ If true, Android will use experimental synchronous events. This will prevent fro
...
 If true, external HTML pasted/inserted into the input (e.g. from Google Docs, Word, or web pages) will be normalized into the canonical tag subset that the enriched parser understands. However, this is an experimental feature, which has not been thoroughly tested. We may decide to enable it by default in a future release.
 
-| Type   | Default Value | Platform |
-| ------ | ------------- |----------|
-| `bool` | `false`       | Both     |
+| Type   | Default Value | Platform     |
+| ------ | ------------- | ------------ |
+| `bool` | `false`       | iOS, Android |
 
 ## Ref Methods
```
szydlovsky, line 650: I know it might be a big ask but I think we should have compatibility tables for methods too if some are not supported on Web.

docs/INPUT_API_REFERENCE.md
kacperzolkiewski: removed

docs/INPUT_API_REFERENCE.md
kacperzolkiewski: fixed

docs/INPUT_API_REFERENCE.md
kacperzolkiewski: I added a note under `setTextAlignment` indicating that it isn't supported on Web. Since the other methods are supported across all platforms, I don't think it's necessary to add platform support tables throughout the documentation

exploIF (approved):

PR #606, opened 2026-05-15
fix: setting mention without indicator typed

### LINKED ISSUE #601: setMention method on ref does not trigger mention insertion
**To Reproduce**
```js
ref.current?.setMention('@', `@${item.name}`, {
      id: item.id,
      type: 'user',
    });.
```

**Expected behavior**
The input field should replace the trigger character @ with the mention label (e.g., @JohnDoe) and treat it as an atomic enriched entity. The cursor should move to the position immediately following the mention.

kacperzolkiewski:
# Summary

This PR fixes: #601

## Test Plan

Run reproduction steps from #601, the issue should be resolved

## Screenshots / Videos

https://github.com/user-attachments/assets/9bc1f5f4-028a-4d39-9a6f-18fdf6bd7aca

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist

- [x] E2E tests are passing

### DIFF
--- a/android/src/main/java/com/swmansion/enriched/textinput/styles/ParametrizedStyles.kt
+++ b/android/src/main/java/com/swmansion/enriched/textinput/styles/ParametrizedStyles.kt
@@ -365,7 +365,7 @@ class ParametrizedStyles(
       spannable.removeSpan(span)
     }
 
-    val start = mentionStart ?: return
+    val start = mentionStart ?: selectionStart
 
     view.runAsATransaction {
       spannable.replace(start, selectionEnd, text)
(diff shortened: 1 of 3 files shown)


android/src/main/java/com/swmansion/enriched/textinput/styles/ParametrizedStyles.kt
```diff
@@ -383,6 +383,7 @@ class ParametrizedStyles(
 
     view.mentionHandler?.reset()
     view.selection.validateStyles()
+    mentionStart = null
```
exploIF, line 386: Could you clarify why it's needed?

kacperzolkiewski: Without setting `mentionStart` to `null`, the value is never reset. As a result, when a mention is first added by typing @ and then another one is inserted manually via `setMention`, `mentionStart` still points to the position of the first typed mention, which is incorrect.

https://github.com/user-attachments/assets/d1d13fe4-81b3-4b63-8a7b-6d77a71a2ebf

exploIF (approved):

PR #612, opened 2026-05-20
refactor(web): extensions array clean up, plugin usage and naming consistency

hejsztynx:
# Summary

- there was some inconsistency with plugin namings, as some used `camelCase` and some `PascalCase`, now every plugin is in `PascalCase`
- introduced inline `*Plugin` configurations, just like it was done in the latter `Placeholder` plugin for clarity and to get rid of `useMemo`s and to shorten the dependency array
- unified custom plugin creations, now each one uses `.configure()` method


hejsztynx:
The last commit solely introduces files and directory renamings, to make the pr easier to look at

src/web/pmPlugins/MentionPlugin/makeMentionPluginState.ts
```diff
@@ -1,9 +1,9 @@
 import type { StateField } from '@tiptap/pm/state';
 import { isCaretInBlockedContext } from './isCaretInBlockedContext';
-import type { MentionPluginOptions, TriggerState } from './types';
+import type { TriggerState } from './types';
 
 export function makeMentionPluginState(
-  options: MentionPluginOptions
+  getIndicators: () => string[]
```
kacperzolkiewski:
```suggestion
  getIndicators: MentionPluginOptions['getIndicators']
```

src/web/pmPlugins/MentionPlugin/index.ts
```diff
@@ -0,0 +1,45 @@
...
+export { startMention } from './startMention';
+export { subscribeMentionEvents } from './subscribeMentionEvents';
+
+export const MentionPlugin = Extension.create<MentionPluginOptions>({
+  name: 'mentionTrigger',
+  addOptions() {
+    return {
+      getIndicators: () => {
+        throw new Error(
+          'MentionPlugin.configure({ getIndicators, getCallbacks }) is required'
```
kacperzolkiewski:
```suggestion
          'MentionPlugin.configure({ getIndicators }) is required'
```

src/web/pmPlugins/MentionPlugin/subscribeMentionEvents.ts
```diff
@@ -2,17 +2,17 @@ import { getMarkRange } from '@tiptap/core';
...
+import type { MentionCallbacks, TriggerState } from './types';
 
 export function subscribeMentionEvents(
   editor: Editor,
-  callbacksRef: MentionPluginOptions['callbacksRef']
+  getCallbacks: () => MentionCallbacks
 ): () => void {
   let prevTriggerState: TriggerState = { active: false };
   let prevMentionKey: string | null = null;
+  const cb = getCallbacks();
```
kacperzolkiewski: `subscribeMentionEvents` accepts a `getCallbacks` getter but it only invokes it once when the subscription is created. As a result, when a mention callback changes (e.g. `handleEndMention`) and the app saves, the old callback is still used.
The fix for that would be to call `getCallbacks` inside `handleTransaction` and `handleBlur` right before invoking the callbacks, so the latest versions are always used.

hejsztynx: thanks, good catch!

kacperzolkiewski (approved):
LGTM!

exploIF (approved):

PR #613, opened 2026-05-25
fix: unify detection events between platforms

### LINKED ISSUE #571: Mention detection has some bugs
## Describe the bug

There appear to be issues with both mention-related callbacks:

### 1. `onMentionEnd`

`onMentionEnd` is not triggered immediately after the mention ends. Instead, it seems to wait until multiple spaces are entered, then fires only after typing a subsequent non-space character.

### 2. `onMentionDetected`

`onMentionDetected` does not detect the mention indicator (`@`) correctly.

Instead, it appears to fire when selecting any word, regardless of whether it is a mention. This causes false positives and makes the callback unreliable for mention handling.

---

## To Reproduce

1. Open the example app.
2. Navigate to `TestScreen`.
3. Add `console.log()` statements to `onMentionEnd` and `onMentionDetected`.
4. Type text containing mentions.
5. Observe the callback behavior.

---

### `onMentionEnd`

Expected to fire immediately after the mention is terminated (for example, after the first space following `@john`).

Potentially, mention termination could also happen on special characters such as `.`, `,`, `(`, `)`, `$`, `^`, etc., depending on intended behavior.

### `onMentionDetected`

Expected to fire when the mention trigger character (`@`) is typed and mention parsing begins.

It should not fire for unrelated word selections or non-mention text.

---

### `onMentionEnd` issue

https://github.com/user-attachments/assets/5cd49b12-85e3-4303-bd6b-1392743f59d9

### `onMentionDetected` issue

https://github.com/user-attachments/assets/f2927a94-21ff-45e1-8f36-ed12ebc2226c

---

## Environment

- Device: Emulator
- OS: Android 16
- Library version: 0.6.1

---

## Additional context

Tested on the example app included in the repository.

kacperzolkiewski:
# Summary

Fixes: #571 

Unify `onMentionDeteted` and `onLinkDetected` events emitting between platforms.

## Test Plan

Verify that `onMentionDetected` and `onLinkDetected` are triggered correctly in the following scenarios:
- moving the cursor into or selecting a mention/link emits the detection event
- moving the cursor outside a mention/link emits the clear event

## Screenshots / Videos

https://github.com/user-attachments/assets/53b8ce07-803c-4f77-b0e4-77fc3cd734d0

Web:
 

https://github.com/user-attachments/assets/a25f217a-1245-4068-bfe4-f9b4b23926b5

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| Web |    ✅     |

## Checklist

- [ ] E2E tests are passing


ios/EnrichedTextInputView.mm
```diff
@@ -1011,60 +1013,69 @@ - (void)tryUpdatingActiveStyles {
...
+        } else if (![_recentlyActiveLinkData
+                       isEqualToLinkData:candidateLinkData] ||
+                   !NSEqualRanges(_recentlyActiveLinkRange,
+                                  candidateLinkRange)) {
+          // we changed selection from one link to the other or modified
+          // current link's text
+          detectedLinkData = candidateLinkData;
+          detectedLinkRange = candidateLinkRange;
+        }
+      } else if (wasActive && [self wasLinkPreviouslyDetected]) {
```
szydlovsky: Just thinking out loud, but isn't `wasActive` check enough here? It's a case where link is not active now, but was active in the previous payload. If we add that to the fact that we ***always*** want to emit link detection at some point with a given link, I think `wasMentionPreviouslyDetected` is actually excessive here - we are just sure that it must have been detected previously.

ios/EnrichedTextInputView.mm
```diff
@@ -1011,60 +1013,69 @@ - (void)tryUpdatingActiveStyles {
...
+                       isEqualToString:candidateMentionParams.text] ||
+                   ![_recentlyActiveMentionParams.attributes
+                       isEqualToString:candidateMentionParams.attributes] ||
+                   !NSEqualRanges(_recentlyActiveMentionRange,
+                                  candidateMentionRange)) {
+          // selection changed from one mention to another
+          detectedMentionParams = candidateMentionParams;
+          detectedMentionRange = candidateMentionRange;
+        }
+      } else if (wasActive && [self wasMentionPreviouslyDetected]) {
```
szydlovsky: Same as with links, I think that `wasMentionPreviouslyDetected` is not needed here.

ios/EnrichedTextInputView.mm
kacperzolkiewski: It looks like it is working without `wasMentionPreviouslyDetected`, nice catch!

szydlovsky (approved):
iOS looks and works good

exploIF (approved):

PR #614, opened 2026-05-25
fix(ios): removed selection change event emition on input focus

hejsztynx:
# Summary

- on iOS there was an inconsistency with `selectionChange` events in comparison with other platforms
- `selectionChange` event was emitted every time we focused on the input, which was inconsistent and emitted a confusing event about selection change, even though its state never changed internally
- this also introduced a path when we blurred the input and then focused on a different part of the text that we blurred on, two `selectionChange` events were emitted unnecessarily

## Videos

Comparison of changes

Before:

https://github.com/user-attachments/assets/34cf4003-ab1b-4e74-8b20-f163c4a8623d

After:

https://github.com/user-attachments/assets/472b92e9-0db8-4a7d-bee6-62c7d217e8ac

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist

- [ ] E2E tests are passing

### DIFF
--- a/ios/EnrichedTextInputView.mm
+++ b/ios/EnrichedTextInputView.mm
@@ -1748,15 +1748,6 @@ Class<RCTComponentViewProtocol> EnrichedTextInputViewCls(void) {
     if (_emitFocusBlur) {
       emitter->onInputFocus({});
     }
-
-    NSString *textAtSelection =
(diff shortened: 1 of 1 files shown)


kacperzolkiewski (approved):
LGTM!

szydlovsky (approved):

PR #616, opened 2026-05-26
fix(android): inline styles reapply

### LINKED ISSUE #512: [Android]: inline styles reapply automatically after deselecting them
**Describe the bug**

Inline styles are reapplying in case we deselect them from the editor if we type text in styled format text

**To Reproduce**
1. Open the editor
2. Type any text with bold/italic/underline/strike-through style
3. Put the cursor on the middle of the text
4. Deselect all styles
5. Type text
6. Text becomes styled

**Expected behavior**

Text should be clear since we deselected styles

**Screenshots**

https://github.com/user-attachments/assets/c72df2ce-12b8-4328-926f-eafeb1c1684c

**Additional context**

kacperzolkiewski:
# Summary

Fixes: #512 

## Test Plan

Run reproduction steps from #512 
Issue should be resolved.

## Screenshots / Videos

https://github.com/user-attachments/assets/b76d8c5d-4ad9-49ad-9539-cea176ad60b3

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

## Checklist

- [X] E2E tests are passing

### DIFF
--- a/android/src/main/java/com/swmansion/enriched/textinput/styles/InlineStyles.kt
+++ b/android/src/main/java/com/swmansion/enriched/textinput/styles/InlineStyles.kt
@@ -103,8 +103,16 @@ class InlineStyles(
 
   fun afterTextChanged(
     s: Editable,
+    startCursorPosition: Int,
     endCursorPosition: Int,
   ) {
+    if (endCursorPosition > startCursorPosition) {
+      for ((style, config) in EnrichedSpans.inlineSpans) {
+        if (view.spanState?.getStart(style) != null) continue
+        splitSpanOnInsertion(s, config.clazz, startCursorPosition, endCursorPosition)
+      }
+    }
+
     for ((style, config) in EnrichedSpans.inlineSpans) {
       val start = view.spanState?.getStart(style) ?: continue
       var end = endCursorPosition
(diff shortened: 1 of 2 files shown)


android/src/main/java/com/swmansion/enriched/textinput/styles/InlineStyles.kt
```diff
@@ -103,8 +103,16 @@ class InlineStyles(
 
   fun afterTextChanged(
     s: Editable,
+    startCursorPosition: Int,
     endCursorPosition: Int,
   ) {
+    if (endCursorPosition > startCursorPosition) {
```
exploIF, line 109: Could you elaborate on why this check is needed?

android/src/main/java/com/swmansion/enriched/textinput/styles/InlineStyles.kt
```diff
@@ -117,6 +125,47 @@ class InlineStyles(
 
       setSpan(s, config.clazz, start, end)
     }
+
+    // Collapse same-type inline spans that ended up adjacent
+    // Without this the HTML output would emit separate tags like <b>...</b><b>...</b>.
+    for ((_, config) in EnrichedSpans.inlineSpans) {
+      mergeAdjacentSpansOfType(s, config.clazz)
```
exploIF: This logic should be ready implemented in the `setSpan` method, why it's duplicated there?

android/src/main/java/com/swmansion/enriched/textinput/styles/InlineStyles.kt
```diff
@@ -117,6 +125,47 @@ class InlineStyles(
...
+    }
+  }
+
+  private fun <T> mergeAdjacentSpansOfType(
```
exploIF: There is already a logic with the same responsibility inside `setSpan`

exploIF (changes requested):

android/src/main/java/com/swmansion/enriched/textinput/styles/InlineStyles.kt
kacperzolkiewski: This check is needed because we only want to run `splitSpanOnInsertion` when user actually typed something in the input `(endCursorPosition > startCursorPosition)`

android/src/main/java/com/swmansion/enriched/textinput/styles/InlineStyles.kt
kacperzolkiewski: `setSpan` only merges neighbours when it is called. When the user deletes the unstyled gap between two bold spans, `setSpan` is never invoked, so we need to call this `mergeAdjacentSpansOfType` to collapse them into one bold span properly.

kacperzolkiewski: Hmm, but maybe we can combine this logic somehow

exploIF: I see, but in such scenario, shouldn't we somehow reuse merging logic that is already implemented? I would prefer to avoid code duplication

android/src/main/java/com/swmansion/enriched/textinput/styles/InlineStyles.kt
kacperzolkiewski: I reuse setSpan logic right now, fixed

android/src/main/java/com/swmansion/enriched/textinput/styles/InlineStyles.kt
kacperzolkiewski: I reuse setSpan logic right now, fixed

android/src/main/java/com/swmansion/enriched/textinput/styles/InlineStyles.kt
```diff
@@ -143,6 +167,33 @@ class InlineStyles(
...
+    for (span in spans) {
+      val spanStart = spannable.getSpanStart(span)
+      val spanEnd = spannable.getSpanEnd(span)
+      if (spanStart < 0 || spanEnd < 0) continue
+
+      spannable.removeSpan(span)
+
+      if (spanStart < insertStart) {
+        val (safeStart, safeEnd) = spannable.getSafeSpanBoundaries(spanStart, insertStart)
+        val left = type.getDeclaredConstructor(HtmlStyle::class.java).newInstance(view.htmlStyle)
```
exploIF, line 186: Could you elaborate on why do we need to use such approach? I don't think we use it in other places.

Isn't it enough to use `span` that is already retrieved?

kacperzolkiewski: We use this approach in other places inside the project, take a look for example here: 
https://github.com/software-mansion/react-native-enriched/blob/5c28968eb1690835869764139fa342b7006a7031/android/src/main/java/com/swmansion/enriched/textinput/styles/ParagraphStyles.kt#L106

Actually we even have that in this file:
https://github.com/software-mansion/react-native-enriched/blob/5c28968eb1690835869764139fa342b7006a7031/android/src/main/java/com/swmansion/enriched/textinput/styles/InlineStyles.kt#L42

exploIF (approved):

PR #617, opened 2026-05-26
fix(iOS): inline styles parsing

kacperzolkiewski:
# Summary

We introduced a parser regression after merging #403, causing the `inline_styles_display.yaml` test to fail.
When multiple inline styles are applied to the same word, the styles incorrectly extend onto the surrounding text
image

Also fixes applying alignment inside input from parsed html
## Test Plan

`inline_styles_display.yaml` test passes.
Parsing works properly when alignment set

## Screenshots / Videos

n/a

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist

- [X] E2E tests are passing


ios/htmlParser/HtmlParser.mm
szydlovsky, line 1: Do I see it right that this whole change in that file is essentially just a cosmetic touch-up and doesn't change our parsing logic at all?

ios/inputHtmlParser/InputHtmlParser.mm
```diff
@@ -213,6 +212,7 @@ - (void)applyProcessedAlignments:(NSArray<AlignmentEntry *> *)alignments
                       withTyping:NO
                   withDirtyRange:NO];
   }
+  [_input anyTextMayHaveBeenModified];
```
szydlovsky: I'd rather we moved the `anyTextMayHaveBeenModified` call to after the call sites of `applyProcessedAlignments`. In the future we might need to add another `apply...` method and might also forget to move it into the last `apply...` call.

szydlovsky: So in here, let's just make `anyTextMayHaveBeenModified` as the last called method of `replaceFromHtml`, `replaceWholeFromHtml` and `insertFromHtml`

ios/htmlParser/HtmlParser.mm
kacperzolkiewski: With previous logic the post-tag cleanup wasn't reached:
```
      closingTag = NO;
      currentTagName = [[NSMutableString alloc] initWithString:@""];
      currentTagParams = [[NSMutableString alloc] initWithString:@""];
```

ios/inputHtmlParser/InputHtmlParser.mm
kacperzolkiewski: fixed

ios/htmlParser/HtmlParser.mm
szydlovsky: ohh okay! makes sense

szydlovsky (approved):
:shipit:

Issue #492, opened 2026-03-19
[Feature request] allowFontScaling support


alizahid:
**Is your feature request related to a problem? Please describe.**

It would be nice to have support for the `allowFontScaling` prop. In my composer where enriched lives, there's a few items I've set `allowFontScaling` to `false` so enriched looks a bit awkward.

**Describe the solution you'd like**

Support for `allowFontScaling` prop

**Describe alternatives you've considered**

None

**Additional context**

None

largiter:
I also encountered this problem, my temporary solution is to do the calculations in "style", you can use react native "PixelRatio" or if you are using react-native-unistyles (as I do) - "rt".

```
const style = StyleSheet.create((theme, rt) => ({
  input: {
    fontSize: theme.typography.p2.fontSize / rt.fontScale,
    lineHeight: theme.typography.p2.lineHeight / rt.fontScale,
  },
}))

```

kacperzolkiewski:
Hi @alizahid, 

Thanks for bringing this up!
I'm closing this issue since the feature has already been implemented and is ready to use: https://github.com/software-mansion/react-native-enriched-html/pull/620

PR #620, opened 2026-05-27
feat: allowFontScaling prop

hejsztynx:
# Summary

Thanks for the feature suggestion in this [issue](https://github.com/software-mansion/react-native-enriched/issues/492)!

Implemented a `allowFontScaling` prop

- when set to `true`, the font size will be scaled to satisfy the system's accessibility font size settings (default behaviour)
- when set to `false`, it works independently of the current settings

## Videos

`EnrichedTextInput`

https://github.com/user-attachments/assets/a9fbac99-bc35-41b8-9965-f04539bfce3b

`EnrichedText`

https://github.com/user-attachments/assets/11469ae4-852e-48da-a43c-9800064a0e1b

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅    |
| Android |    ✅     |

## Checklist

- [ ] E2E tests are passing


.maestro/enrichedInput/flows/font_scaling.yaml
```diff
@@ -0,0 +1,57 @@
...
+---
+- launchApp
+
+- tapOn:
+    id: "toggle-screen-button"
+
+- runFlow:
+    file: "../subflows/capture_or_assert_screenshot.yaml"
+    env:
+        SCREENSHOT_NAME: "font_scaling__placeholder"
```
kacperzolkiewski:
```suggestion
        SCREENSHOT_NAME: "font_scaling_placeholder"
```

.maestro/enrichedText/flows/font_scaling.yaml
```diff
@@ -0,0 +1,34 @@
+appId: swmansion.enriched.example
+tags:
```
kacperzolkiewski, line 2: You should either mark this as iOS-only or change the HTML value, because the Android screenshot is not correct for alignment - the Android  text alignment support hasn’t been merged yet.

apps/example/src/screens/DevScreen.tsx
```diff
@@ -75,6 +75,7 @@ export function DevScreen({ onSwitch }: DevScreenProps) {
             onPasteImages={(e) => editor.handlePasteImagesEvent(e.nativeEvent)}
             useHtmlNormalizer
             testID="editor-input"
+            allowFontScaling
```
exploIF: It's redundant, as it's a default value, correct?

.maestro/enrichedInput/flows/font_scaling.yaml
hejsztynx: Currently there are two screenshots being taken in the `font_scaling` flow. If we had the screenshot named `font_scaling_placeholder`, the name would be confusing - it looks like there is a separate flow named `font_scaling_placeholder`. I thought indicating dependent screenshots with `__` would clear things up, what do you think?

.maestro/enrichedText/flows/font_scaling.yaml
hejsztynx: Good point 👍 I've changed the html

apps/example/src/screens/DevScreen.tsx
hejsztynx: Correct, removed

.maestro/enrichedInput/flows/font_scaling.yaml
kacperzolkiewski: We already have E2E tests that generate multiple screenshots. For example, `scrolling_with_paragraph_styles.yaml` produces two screenshots: `scrolling_paragraph_styles_top` and `scrolling_paragraph_styles_bottom`.
As you can see, those screenshot names do not use a double underscore (`_ _`), so I would prefer to follow the existing convention and rename this to `font_scaling_placeholder`.

hejsztynx: Ok, sorry for missing that, applying your suggestion then

exploIF (approved):
Works good

apps/example/src/screens/DevScreen.tsx
exploIF: It's still there, let's use default value on the `DevScreen`

ios/config/EnrichedConfig.mm
```diff
@@ -67,6 +68,7 @@ - (instancetype)init {
   _primaryFontNeedsRecreation = YES;
   _monospacedFontNeedsRecreation = YES;
   _olMarkerFontNeedsRecreation = YES;
+  _allowFontScaling = YES;
```
szydlovsky: I wouldn't put it here. Only internal flags used in config are instantiated here (see flags that track font recreation above). Since `_allowFontScaling` is essentially mirroring a prop and only set via `updateProps`, I think we can entirely drop this statement.

ios/EnrichedTextInputView.mm
```diff
@@ -596,6 +596,15 @@ - (void)updateProps:(Props::Shared const &)props
     [textView setScrollEnabled:newViewProps.scrollEnabled];
   }
 
+  if (newViewProps.allowFontScaling != oldViewProps.allowFontScaling ||
+      isFirstMount) {
+    textView.adjustsFontForContentSizeCategory = newViewProps.allowFontScaling;
+    _placeholderLabel.adjustsFontForContentSizeCategory =
+        newViewProps.allowFontScaling;
```
szydlovsky: I did some research and we actually shouldn't need the `adjustsFontForContentSizeCategory` prop. It is used for automatic resizing ***only if*** the font was instantiated either via
`UIFont.preferredFont(forTextStyle: ..)`
or
`UIFontMetrics(forTextStyle: ..).scaledFont(for: ..)`

We don't use either - instead we are subscribed to `traitCollectionDidChange`, which in turn calls `[config invalidateFonts]`.

Instead of these calls however, we definitely want to run `[config invalidateFonts]` here if the `allowFontScaling` prop was changed - maybe we didn't touch the accessibility scaling but just the prop change alone should cause fonts recalculation.

ios/EnrichedTextInputView.mm
```diff
@@ -1985,6 +1994,12 @@ - (void)textViewDidChange:(UITextView *)textView {
 - (void)traitCollectionDidChange:(UITraitCollection *)previousTraitCollection {
   [super traitCollectionDidChange:previousTraitCollection];
 
+  const auto &viewProps =
+      *std::static_pointer_cast<EnrichedTextInputViewProps const>(_props);
+  if (!viewProps.allowFontScaling) {
+    return;
+  }
+
```
szydlovsky: Instead of reading props we can just access the `allowFontScaling` from the `config`.

kacperzolkiewski (approved):
Looks and works correct! 🚀

ios/EnrichedTextView.mm
```diff
@@ -582,6 +582,14 @@ - (void)updateProps:(Props::Shared const &)props
     useHtmlNormalizer = newViewProps.useHtmlNormalizer;
   }
 
+  // allowFontScaling
+  if (newViewProps.allowFontScaling != oldViewProps.allowFontScaling ||
+      isFirstMount) {
+    textView.adjustsFontForContentSizeCategory = newViewProps.allowFontScaling;
```
szydlovsky: Same as in input, we probably want to exchange this call with a `[config invalidateFonts]`.

szydlovsky: Also, I believe we don't need `isFirstMount` check here - `oldViewProps` on first mount will usually be not defined so any first value will always be different than the compared "old" one. Please also check it for the input.

ios/EnrichedTextView.mm
```diff
@@ -698,6 +706,39 @@ - (void)tryUpdatingHeight {
...
+
+  if (_props == nullptr) {
+    return;
+  }
+
+  const auto &viewProps =
+      *std::static_pointer_cast<EnrichedTextViewProps const>(_props);
+  if (!viewProps.allowFontScaling) {
+    return;
+  }
```
szydlovsky: Same as in input, we can just access the `allowFontScaling` from the `config`.

ios/EnrichedTextView.mm
```diff
@@ -698,6 +706,39 @@ - (void)tryUpdatingHeight {
...
+
+  if (previousTraitCollection.preferredContentSizeCategory ==
+      self.traitCollection.preferredContentSizeCategory) {
+    return;
+  }
```
szydlovsky, line 728: With all the changes I'm requesting I believe this is not needed either. The `traitCollectionDidChange` event usually means something in terms of sizing changed and I believe it's safer to always do the recalculation. The trait collection changes are so infrequent in normal app use that it shouldn't slow the library down and we get a safety net.

szydlovsky (changes requested):

ios/EnrichedTextView.mm
hejsztynx: We've discussed about it and we'll keep it as it is

ios/EnrichedTextInputView.mm
```diff
@@ -1985,6 +1990,10 @@ - (void)textViewDidChange:(UITextView *)textView {
 - (void)traitCollectionDidChange:(UITraitCollection *)previousTraitCollection {
   [super traitCollectionDidChange:previousTraitCollection];
 
+  if (!config.allowFontScaling) {
+    return;
+  }
+
   if (previousTraitCollection.preferredContentSizeCategory !=
       self.traitCollection.preferredContentSizeCategory) {
```
szydlovsky, line 2009: One more improvement, we can make the `previousTraitCollection` an early return too:
(might be formatted wrongly)

```suggestion
  if (!config.allowFontScaling) {
    return;
  }

if (previousTraitCollection.preferredContentSizeCategory == self.traitCollection.preferredContentSizeCategory) {
return;
}
```

docs/INPUT_API_REFERENCE.md
```diff
@@ -2,6 +2,14 @@
 
 ## Props
 
+### `allowFontScaling`
+
+If `true`, it respects the system's accessibility font scaling settings.
```
szydlovsky:
```suggestion
If `true`, the input respects the system's accessibility font scaling settings.
```

docs/TEXT_API_REFERENCE.md
```diff
@@ -2,6 +2,14 @@
 
 ## Props
 
+### `allowFontScaling`
+
+If `true`, it respects the system's accessibility font scaling settings.
```
szydlovsky:
```suggestion
If `true`, the text respects the system's accessibility font scaling settings.
```

szydlovsky (approved):

PR #624, opened 2026-06-01
fix: split manual links when newline inserted in link

### LINKED ISSUE #556: onLinkDetected invalid behavior when splitting a link over multiple lines
**Describe the bug**
If we create a link, and then split it across many lines by pressing enter the html output shows we have a couple of seperate link elements. But:
+ on Android `onLinkDetected` treats those multiple links as a single one
+ on iOS we treat those links as seperate but the `text` attribute is invalid, it returns the link text before splitting it.

**To Reproduce**
1. In `useEditorState.ts` set
```js
  const handleLinkDetected = (state: CurrentLinkState) => {
    console.log('link detected', state);
    setCurrentLink(state);
  };
```
2. Launch example app
3. Add an example link
4. Repeatedly press enter in the middle of the link to split it across multiple lines.
6. On Android `onLinkDetected` treats those multiple links as a single one. On iOS the `text` attribute is wrong, it tells us the link text before splitting.

**Expected behavior**
On Android `onLinkDetected` shouldn't treat those multiple links as one. On iOS `text` should return substrings  in the video example `H`, 'e', `l`, `l`, 'o`
**Screenshots**

https://github.com/user-attachments/assets/27125c21-6aee-4529-bd93-840c0ba35990

 - Device: iPhone 17, Pixel 9
 - OS: iOS 26.2, Android 16
 - Version -, API 36 "Baklava"

**Additional context**

kacperzolkiewski:
# Summary

Fixes: #556 

## Test Plan

Run reproduction steps from issue: #556 
Issue should be resolved

## Screenshots / Videos

https://github.com/user-attachments/assets/647fca9a-f67d-4b3b-ad36-a4b8ee69ea6b

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist

- [x] E2E tests are passing


ios/EnrichedTextInputView.mm
```diff
@@ -1591,8 +1591,11 @@ - (void)manageSelectionBasedChanges {
   [attributesManager
       manageTypingAttributesWithOnlySelection:onlySelectionChanged];
 
-  // always update active styles
-  [self tryUpdatingActiveStyles];
+  // When text changed, anyTextMayHaveBeenModified runs tryUpdatingActiveStyles
+  if ([_recentInputString isEqualToString:currentString]) {
+    // always update active styles
```
kacperzolkiewski:
```suggestion
    // update active styles
```

szydlovsky (approved):
iOS works fine

ios/EnrichedTextInputView.mm
```diff
@@ -1692,12 +1692,11 @@ - (void)anyTextMayHaveBeenModified {
...
 
   if (![textView.textStorage.string isEqualToString:_recentInputString]) {
+    _recentInputString = [textView.textStorage.string copy];
+
     // emit onChangeText event
     auto emitter = [self getEventEmitter];
     if (emitter != nullptr && _emitTextChange) {
-      // set the recent input string only if the emitter is defined
-      _recentInputString = [textView.textStorage.string copy];
-
```
szydlovsky: We can't have that here. It will bring back an issue from https://github.com/software-mansion/react-native-enriched-html/pull/270.

There is however a thing we could do to probably satisfy both needs:
set `_recentInputString` if emitter is defined, but not checking the `_emitTextChange` here - it should only be used for the actual emitting a bit later in the code.

szydlovsky (changes requested):

ios/EnrichedTextInputView.mm
kacperzolkiewski: Fixed

szydlovsky (approved):
lgtm

PR #628, opened 2026-06-08
chore: remove js-beautify

exploIF:
# Summary

Remove `js-beautify` as we don't really need it. That way we can get rid of vulnerable `js-cookie` as well


kacperzolkiewski (approved):
Works well!
Remember to delete this PR https://github.com/software-mansion/react-native-enriched/pull/625 after merging :)

PR #629, opened 2026-06-09
fix(android): set piority of applied spans

kacperzolkiewski:
# Summary

Apply priority to android spans that inline spans will be applied after paragraph spans.
See: https://stackoverflow.com/questions/46002947/how-to-define-order-setpriority-of-spannable-in-android

## Test Plan
Play around with styles on android and check if inline styles override paragraph styles. 
For example: Inline code inside blockquote.

## Screenshots / Videos

https://github.com/user-attachments/assets/9bacb687-ad62-43fe-a0d0-4c135548f1ef

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

## Checklist

- [x] E2E tests are passing


android/src/main/java/com/swmansion/enriched/common/parser/EnrichedParser.java
```diff
@@ -467,10 +468,11 @@ public Spanned convert() {
       if (end == start) {
         mSpannableStringBuilder.removeSpan(obj[i]);
       } else {
-        // TODO: verify if Spannable.SPAN_EXCLUSIVE_EXCLUSIVE does not break anything.
+        // TODO: verify if Spannable.SPAN_EXCLUSIVE_EXCLUSIVE from
+        // EnrichedSpanFlags.paragraphSpanFlags does not break anything.
```
exploIF: Well, I've added this comment couple months ago. I guess it can be removed :D

android/src/main/java/com/swmansion/enriched/common/EnrichedSpanFlags.kt
```diff
@@ -0,0 +1,35 @@
...
+
+  @JvmField
+  val inlineSpanFlags: Int = applyPriority(Spannable.SPAN_EXCLUSIVE_EXCLUSIVE, INLINE_SPAN_PRIORITY)
+
+  fun forSpan(
+    span: EnrichedSpan,
+    baseFlags: Int,
+  ): Int {
+    val isInlineSpan = span is EnrichedInlineSpan
+    val priority = if (isInlineSpan) INLINE_SPAN_PRIORITY else PARAGRAPH_SPAN_PRIORITY
```
exploIF: I wonder if we shouldn't have a third case. For example `AlignmenSpan` is not marked as `EnrichedInlineSpan` either `EnrichedParagraphSpan`

exploIF (approved):
Looks good!

android/src/main/java/com/swmansion/enriched/common/EnrichedSpanFlags.kt
kacperzolkiewski: Good idea! I added also priority for alignment

PR #630, opened 2026-06-09
feat(web): HTML normalization

hejsztynx:
# Summary

Web implementation was missing the HTML normalizer

- implemented the normalizer, mirroring the one in the native code
- tests also mirror the native ones 

## Test Plan

Try pasting styled text from `Google Docs`, `MS Word` or any other platform

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ❌     |
| Web |    ✅      |

## Checklist

- [x] E2E tests are passing


yarn.lock
kacperzolkiewski, line 1: I think that you should update your `yarn.lock` because every CI job is failing

src/web/normalization/htmlNormalizer.ts
```diff
@@ -0,0 +1,609 @@
...
+function hasBlockOrBqChild(node: Element): boolean {
+  for (const child of Array.from(node.childNodes)) {
+    if (isBlockProducing(child) || isBlockquoteNode(child)) return true;
+  }
+  return false;
+}
+
+function findCssValue(style: string, prop: string): string | null {
+  // Returns the value of the last declaration with this property name.
+  // Mirrors GumboNormalizer's find_css_value scanning behavior.
```
kacperzolkiewski, line 140: I think that C `find_css_value` returns the first matching property declaration and stops, not the last one:
https://github.com/software-mansion/react-native-enriched/blob/12b073090a7ab4b4d115f470ec7239acba6a32c2/cpp/parser/GumboNormalizer.c#L273

kacperzolkiewski (approved):
Overall solid, and works fine!

src/web/normalization/htmlNormalizer.ts
hejsztynx: Great catch! TY

exploIF (approved):

PR #631, opened 2026-06-09
feat: adjust opacity settings

### LINKED ISSUE #627: Mentions on iOS have forced background opacity of 0.4
**Describe the bug**
On iOS, the backgroundColor of mentions has a forced opacity of 0.4 if the opacity set is above 0.0

**To Reproduce**
1. Create a very transparent mention.
```
<EnrichedTextInput
  htmlStyle={{
    mention: {
      backgroundColor: "rgba(255, 255, 255, 0.1)"
    }
  }} />
```
3. Create a mention. Notice it's quite white.

**Expected behavior**
iOS builds shouldn't override the user's choice of color.

 - Device: iPhone 16 Pro
 - OS: iOS 26.5
 - Version: 0.5.2

**Additional context**
I solved this for myself with the following patch. unfortunately don't have the time currently to set up a development environment for RNEnriched to test all checks pass.

```
diff --git a/ios/styles/MentionStyle.mm b/ios/styles/MentionStyle.mm
index 5d10772e501e6464b8eb0fb01f0bb0988f2c5235..b4cc9fd071f90ecf31d2a2bd5cd45b449deec032 100644
--- a/ios/styles/MentionStyle.mm
+++ b/ios/styles/MentionStyle.mm
@@ -198,8 +198,7 @@ static NSString *const MentionAttributeName = @"MentionAttributeName";
     NSForegroundColorAttributeName : styleProps.color,
     NSUnderlineColorAttributeName : styleProps.color,
     NSStrikethroughColorAttributeName : styleProps.color,
-    NSBackgroundColorAttributeName :
-        [styleProps.backgroundColor colorWithAlphaIfNotTransparent:0.4],
+    NSBackgroundColorAttributeName : styleProps.backgroundColor,
   } mutableCopy];
 
   if (styleProps.decorationLine == DecorationUnderline) {
@@ -238,8 +237,7 @@ static NSString *const MentionAttributeName = @"MentionAttributeName";
     NSForegroundColorAttributeName : styleProps.color,
     NSUnderlineColorAttributeName : styleProps.color,
     NSStrikethroughColorAttributeName : styleProps.color,
-    NSBackgroundColorAttributeName :
-        [styleProps.backgroundColor colorWithAlphaIfNotTransparent:0.4],
+    NSBackgroundColorAttributeName : styleProps.backgroundColor,
   } mutableCopy];
 
   if (styleProps.decorationLine == DecorationUnderline) {
```

exploIF:
# Summary

Fixes: https://github.com/software-mansion/react-native-enriched/issues/627
Do not apply additional opacity if user specified it manually. 

## Test Plan

First scenario:
- Run the app
- Adjust `htmlStyle` to use color with opacity
- Notice additional opacity is not applied

Second scenario:
- Run the app
- Adjust `htmlStyle` to use color without opacity
- Notice that additional opacity is applied

## Screenshots / Videos

n/a

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist

- [x] E2E tests are passing


ios/styles/InlineCodeStyle.mm
```diff
@@ -29,7 +29,7 @@ - (void)applyStyling:(NSRange)range {
     [self.host.textView.textStorage
         addAttribute:NSBackgroundColorAttributeName
                value:[[self.host.config inlineCodeBgColor]
-                         colorWithAlphaIfNotTransparent:0.4]
+                         colorWithDefaultAlpha:0.4]
```
kacperzolkiewski: Since 0.4 is used as the default value throughout the codebase, it might make sense to define it as a constant in `ColorExtension.h` and have `colorWithDefaultAlpha` use that constant internally instead of requiring the value to be passed in each time?

exploIF: Kind of out of scope change, as it has been used like that for a while, but I've fixed that

ios/extensions/ColorExtension.mm
```diff
@@ -27,10 +27,14 @@ - (BOOL)isEqualToColor:(UIColor *)otherColor {
   return [selfColor isEqual:otherColor];
 }
 
-- (UIColor *)colorWithAlphaIfNotTransparent:(CGFloat)newAlpha {
+- (UIColor *)colorWithDefaultAlpha {
```
szydlovsky: The name `colorWithDefaultAlpha` suggests it always sets some default alpha.

I'd rather have it named like `colorWithResolvedAlpha` and at least add a comment stating that if the alpha was manually set to something other than `1.0`/`255`, we respect it and otherwise use `0.4` because it lets us see the cursor nicely.

kacperzolkiewski (approved):
LGTM!

szydlovsky (approved):

PR #632, opened 2026-06-09
docs: jsdocs improvements

exploIF:
# Summary

Adds JSDocs


src/types.ts
```diff
@@ -259,8 +259,18 @@ export type TextShortcutStyle =
   | 'ordered_list'
   | 'checkbox_list';
 
+/**
+ * Defines a single text shortcut: a character sequence that, when typed at
+ * the start of a line, is replaced by the corresponding block or inline style.
```
szydlovsky:
```suggestion
 * the start of a line, is replaced by the corresponding paragraph or inline style.
```

src/types.ts
```diff
@@ -259,8 +259,18 @@ export type TextShortcutStyle =
   | 'ordered_list'
   | 'checkbox_list';
 
+/**
+ * Defines a single text shortcut: a character sequence that, when typed at
```
szydlovsky: I'd also add that it isn't always at the start of a line, but rather either at the start of a line for paragraph styles or around some text for inline styles..

src/types.ts
```diff
@@ -414,40 +424,144 @@ export interface OnSubmitEditing {
...
+  /** Focuses the editor, opening the software keyboard on mobile. */
   focus: () => void;
+
+  /** Removes focus from the editor and dismisses the software keyboard. */
   blur: () => void;
+
+  /** Replaces the entire editor content with the supplied HTML/plain-text `value`. */
   setValue: (value: string) => void;
+
+  /** Moves the text cursor (or extends the selection) to the range `[start, end]`. */
```
szydlovsky: Not 100% sure, but I think we don't include the `end` index? If so, maybe `[start, end)` should rather be the range supplied here.

src/types.ts
```diff
@@ -414,40 +424,144 @@ export interface OnSubmitEditing {
...
+  /** Toggles underline on the current selection (or toggles it for future typing if nothing is selected). */
   toggleUnderline: () => void;
+
+  /** Toggles strikethrough on the current selection (or toggles it for future typing if nothing is selected). */
   toggleStrikeThrough: () => void;
+
+  /** Toggles inline code on the current selection (or toggles it for future typing if nothing is selected). */
   toggleInlineCode: () => void;
+
+  /** Toggles an H1 heading on the current selection or the focused paragraph. */
```
szydlovsky: Maybe
```suggestion
  /** Toggles an H1 heading for the paragraph(s) at the current selection.*/
```
If you agree, let's change it for the rest of the paragraph styles too.

src/types.ts
```diff
@@ -414,40 +424,144 @@ export interface OnSubmitEditing {
...
   toggleUnorderedList: () => void;
+
+  /**
+   * Toggles a checkbox (task) list on the current selection or the focused
+   * paragraph. Pass `checked` as `true` to pre-check the inserted items.
+   */
   toggleCheckboxList: (checked: boolean) => void;
+
+  /**
+   * Inserts or updates a hyperlink over the character range `[start, end]`.
```
szydlovsky: Again, if the end is exclusive we could mark the range as `[start, end)`

src/types.ts
```diff
@@ -414,40 +424,144 @@ export interface OnSubmitEditing {
...
+   *
+   * @param start - Start offset of the linked text (inclusive).
+   * @param end   - End offset of the linked text (exclusive).
+   * @param text  - Display text for the link.
+   * @param url   - URL the link should navigate to.
+   */
   setLink: (start: number, end: number, text: string, url: string) => void;
+
+  /**
+   * Removes any hyperlink from the `[start, end]` range while
```
szydlovsky: `[start, end)` here too

szydlovsky (approved):

PR #637, opened 2026-06-14
fix: various mentions bugs and inconsistencies between platforms

hejsztynx:
# Summary

There were various bugs and inconsistencies with mentions functionalities across every platform.

Mainly it wasn't handled properly when you switched **directly** between different mentions, which caused improper event emission and the lists stayed visible - documented in the videos section

iOS:
- events not emitted when switching directly between mentions
- duplicated `onChange` events when just moving a cursor across a mention
- `_recentInputString` was not updated if no `onChangeText` prop was given

Android:
- events not emitted when switching directly between mentions
- on selection changes, the mentions were not reanalyzed and no events were emitted
- when `setMention` a duplicated space was inserted if there already was one
- `setMention` wasn't replacing the surrounding text
-  after `setMention` the selection wasn't moved to the trailing space

Web:
- refactor on the event subscription hook for consistency
- mention query was evaluated to a cursor position, not the whole written mention, which produced leftover text when inserting a mention
- unnecessarily considered a finalized mention into the query
- events not emitted when switching directly between mentions
- when `setMention` a duplicated space was inserted if there already was one

Example apps:
- updated example apps with more idiomatic code. Earlier `onChange` mention event would handle popup list opening - something what `onStart` is for

## Test Plan

Follow the text written in the attached videos, play around with selection changes, all events should emit correctly and the lists should function as expected.

## Screenshots / Videos

(all platforms) events not emitting when switching between mentions

https://github.com/user-attachments/assets/cc408357-ae3f-41ed-87ce-c6192149539b

(web) incorrect query evaluation and inserting, doubled space

https://github.com/user-attachments/assets/4981556a-c8c6-4139-855e-957fa9dc4754

(android) `setMention` not replacing surrounding text:

https://github.com/user-attachments/assets/37fb805c-aaa1-4947-a8dd-4d1427ab1629

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| Web |    ✅     |

## Checklist

- [x] E2E tests are passing


kacperzolkiewski (changes requested):
1. Something is wrong on Android right now:

https://github.com/user-attachments/assets/be40f980-886b-4a9f-848d-fc2f1a6dbb82

2. Should it emit event when we are with the cursor before indicator:

https://github.com/user-attachments/assets/409bba00-4bd1-41ee-bdb2-9dd0e4bcbce7

3. (If answer on second question is yes): On web the cursor does not go to the end of mention when we insert one like in native:

https://github.com/user-attachments/assets/71a83848-0329-4a56-ad18-f5aebdcb1bc1

hejsztynx:
> 1. Something is wrong on Android right now:
> 
>  Screen.Recording.2026-06-25.at.11.34.16.mov 
> 2. Should it emit event when we are with the cursor before indicator:
> 
>  Screen.Recording.2026-06-25.at.11.38.08.mov 
> 3. (If answer on second question is yes): On web the cursor does not go to the end of mention when we insert one like in native:
> 
>  Screen.Recording.2026-06-25.at.11.40.25.mov

ad 1 - I am aware of that, I fixed it, but I've decided to fix Android's `setMention` in a different PR, as it felt like something a bit bigger and this PR had already a lot in it. When this PR gets merged, I'll publish the one fixing it.

ad 2 - I believe we should emit the event before the indicator, that's how it was initially implemented  and the behavior is shared across all platforms.

ad 3 - Good catch, fixed it.

ios/styles/MentionStyle.mm
```diff
@@ -156,9 +158,25 @@ - (void)addMention:(NSString *)indicator
...
+        [self.host.textView.textStorage.string characterAtIndex:nextCharIndex];
+    if ([[NSCharacterSet whitespaceAndNewlineCharacterSet]
+            characterIsMember:nextChar]) {
+      hasSpaceAfter = YES;
+    }
+  }
+
+  NSString *newText =
+      hasSpaceAfter ? text : [NSString stringWithFormat:@"%@ ", text];
+
```
szydlovsky, line 178: I get the idea but the space was there for two purposes:
- to always have a space after a new mention (it's still this way, good)
- to make sure the user selection shifts away from the mention, previous code always added a space so the selection always moved afterwards via `TextInsertionUtils`. With this change, we can add a mention and still be directly behind it, see:

https://github.com/user-attachments/assets/88398ad6-ef48-47f3-8235-aeb42673cdc4

No strong opinion here, but we could either;
- revert this change
or
- make sure selection still goes behind the space (regardless whether we added it or it was there already)

ios/styles/MentionStyle.mm
```diff
@@ -534,7 +567,18 @@ - (void)removeActiveMentionRange {
...
     _activeMentionIndicator = nullptr;
     _activeMentionRange = nullptr;
-    [self.host emitOnMentionEvent:indicatorCopy text:nullptr];
+    [self emitOnMentionEvent:indicatorCopy text:nullptr];
+  }
+}
+
+- (void)emitOnMentionEvent:(NSString *)indicator text:(NSString *)text {
+  BOOL sameText = (_lastEmittedMentionText == text) ||
+                  [_lastEmittedMentionText isEqualToString:text];
```
szydlovsky: I'm pretty sure we need only the second comparison:

```suggestion
  BOOL sameText = [_lastEmittedMentionText isEqualToString:text];
```

as the `==` operator in terms of `NSStrings` essentially compares the pointer addresses.
Only `isEqualToString` should be used to compare the content.

ios/EnrichedTextInputView.mm
```diff
@@ -1722,12 +1722,11 @@ - (void)anyTextMayHaveBeenModified {
...
 
   if (![textView.textStorage.string isEqualToString:_recentInputString]) {
+    _recentInputString = [textView.textStorage.string copy];
+
     // emit onChangeText event
     auto emitter = [self getEventEmitter];
     if (emitter != nullptr && _emitTextChange) {
-      // set the recent input string only if the emitter is defined
-      _recentInputString = [textView.textStorage.string copy];
-
```
szydlovsky: I traced the code where it was added, it was in https://github.com/software-mansion/react-native-enriched-html/pull/270.

There apparently was a reason why the `_recentInputString` (then called differently but serving the same purpose) was set only after the emitter is defined; without it, it's written to before emitter is there, resulting in no first `onChangeText` event in a case where we have a default value:

https://github.com/user-attachments/assets/77ff2428-aff1-4057-b182-ccf711677cac

`defaultValue` results in initial `onChangeHtml` but no `onChangeText` and that's exactly why `_recentInputString` was set only when the emitter is defined.

So it's a regression on our end.

szydlovsky (changes requested):

ios/styles/MentionStyle.mm
hejsztynx: Now the selection moves to the available space, if there already was one.

ios/EnrichedTextInputView.mm
hejsztynx: As discussed, I branched the condition into two `if`'s, now both issues should be handled nicely.

ios/styles/MentionStyle.mm
hejsztynx: both `_lastEmittedMentionText` and `text` can be `nullptr`, so we have to handle that with the `==`

ios/EnrichedTextInputView.mm
```diff
@@ -1722,18 +1722,20 @@ - (void)anyTextMayHaveBeenModified {
   }
 
   if (![textView.textStorage.string isEqualToString:_recentInputString]) {
+
     // emit onChangeText event
     auto emitter = [self getEventEmitter];
-    if (emitter != nullptr && _emitTextChange) {
-      // set the recent input string only if the emitter is defined
```
szydlovsky, line 1723: Can we please still keep this comment. And maybe add some explanation that it has to be this way so that initial `onChangeText` event is properly emitted.

szydlovsky (approved):
iOS looks good to me, just a one tiny comment

.github/workflows/ci.yml
```diff
@@ -205,7 +205,7 @@ jobs:
       - name: Set up Xcode
         uses: maxim-lobanov/setup-xcode@v1
         with:
-          xcode-version: '16.4'
+          xcode-version: latest-stable
```
exploIF: I would prefer pinned version. Also, this is out of scope of this PR. Should be handled separately

exploIF (changes requested):

hejsztynx:
> ad 1 - I am aware of that, I fixed it, but I've decided to fix Android's `setMention` in a different PR, as it felt like something a bit bigger and this PR had already a lot in it. When this PR gets merged, I'll publish the one fixing it.

It turned out, that PR didn't produce much changes as I initially expected - I merged it and the fix is present on this PR.

kacperzolkiewski (approved):
LGTM!

android/src/main/java/com/swmansion/enriched/textinput/events/MentionHandler.kt
```diff
@@ -27,17 +27,30 @@ class MentionHandler(
     indicator: String,
     text: String?,
   ) {
+    var startMention = false
```
exploIF: Is it really needed? Why not handle `emitEvent(indicator, "")` directly inside `if (previousIndicator != indicator)` condition

android/src/main/java/com/swmansion/enriched/textinput/EnrichedTextInputView.kt
```diff
@@ -92,6 +92,7 @@ class EnrichedTextInputView :
   val alignmentStyles: AlignmentStyles? = AlignmentStyles(this)
   var isDuringTransaction: Boolean = false
   var isRemovingMany: Boolean = false
+  var recentInputString: String = ""
```
exploIF: Why it's needed? Ideally I would like to avoid having such property as it might be difficult in the future to keep it synchronized and correctly handle every single case.

apps/example/metro.config.js
exploIF, line 1: This change seems completely out of scope, why it's needed?

apps/example/metro.config.js
```diff
@@ -10,7 +10,21 @@ const root = path.resolve(__dirname, '../..');
...
   root,
   dirname: __dirname,
 });
+
+config.resolver = {
+  ...config.resolver,
+  blockList: [
+    ...(Array.isArray(config.resolver?.blockList)
+      ? config.resolver.blockList
+      : config.resolver?.blockList
```
exploIF: That's almost the same (or even the same taking into account fact that optional chaining operator is going into second, not into the first one)

src/web/pmPlugins/MentionPlugin/useMentionEvents.ts
```diff
@@ -0,0 +1,114 @@
...
+      editor.off('transaction', handleUpdate);
+      editor.off('blur', handleBlur);
+    };
+  }, [editor, getCallbacks]);
+}
+
+function getMentionInCurrentSelection(
+  editor: Editor
+): (OnMentionDetected & { key: string }) | null {
+  const { state } = editor;
```
exploIF:
```suggestion
function getMentionInCurrentSelection(
  { state }: Editor
): (OnMentionDetected & { key: string }) | null {
```

apps/example/metro.config.js
hejsztynx: Those lines differed in types, but that ternary operator flow was messy. I refactored that and it's much cleaner

apps/example/metro.config.js
hejsztynx: I introduced maestro e2e tests where I took screenshots of the whole app (to capture the user list at the bottom). That exposed an issue where if a screenshot was added to the `.maestro` directory, the metro bundler reloaded and you could see the "Refreshing..." metro banner at the top of the screen, which was then captured on the test screenshot

android/src/main/java/com/swmansion/enriched/textinput/EnrichedTextInputView.kt
hejsztynx: I wanted to mirror the iOS behavior that uses `_recentInputString` to know if a selection change handler runs on **only** selection change, or it's a text+selection change. We run `detectActiveMention` on text change, so it would make `detectActiveMention` run twice. But I see your point and I dropped that `recentInputString`. The `detectActiveMention` now runs twice, but it's effectively idempotent. It's a valid trade-off 👍

exploIF (approved):

PR #639, opened 2026-06-15
chore: Release 1.0.0-related renames

szydlovsky:
# Summary

- All the needed renames related to `1.0.0`'s `react-native-enriched`-> `react-native-enriched-html` change.
- Also updated all the needed docs.
- Links are updated but no repo rename has been done yet, so it needs checking.
- iOS needed some generated code imports updates because of `Podspec` rename.

## Test Plan
--
## Screenshots / Videos
--
## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |


README.md
```diff
@@ -46,9 +46,8 @@ We can help you build your next dream product –
 
 ## Prerequisites
 
-- `react-native-enriched` currently supports only Android and iOS platforms
+- `react-native-enriched-html` currently supports Android and iOS, Web implementation is still experimental.
 - It works only with [the React Native New Architecture (Fabric)](https://reactnative.dev/architecture/landing-page) and supports following React Native releases: `0.81`, `0.82`, `0.83`, `0.84` and `0.85`.
-- If you would like to use `react-native-enriched` in React Native `0.79` or `0.80` use `react-native-enriched 0.4.x`.
```
szydlovsky, line 51: I intentionally removed this line, as we would have tell users to use deprecated `react-native-enriched` package for the `0.79` or `0.80` RN versions.

package.json
kacperzolkiewski, line 212:
```suggestion
    "name": "ReactNativeEnrichedHtmlSpec",
```

Should't we change also name here?

README.md
```diff
@@ -1,27 +1,27 @@
-<img src="https://github.com/user-attachments/assets/c8ba03bc-4ea8-48f4-9566-02ebac0c19d3" alt="react-native-enriched by Software Mansion" width="100%">
+<img src="https://github.com/user-attachments/assets/c8ba03bc-4ea8-48f4-9566-02ebac0c19d3" alt="react-native-enriched-html by Software Mansion" width="100%">
 
 [![Ad](https://swm-delivery.com/www/images/zone-gh-react-native-enriched-1?n=1)](https://swm-delivery.com/www/delivery/ck-slug.php?zoneid=zone-gh-react-native-enriched-1&n=1)
```
kacperzolkiewski, line 3: What about those links, they would not change if the package will change to `react-native-enriched-html`?

kacperzolkiewski (approved):
Example app works properly after rename

exploIF (commented):
Have you tested building library and installing it in the app from the tarball?

package.json
```diff
@@ -1,7 +1,7 @@
 {
-  "name": "react-native-enriched",
-  "version": "0.8.0",
-  "description": "Rich Text Editor component for React Native",
+  "name": "react-native-enriched-html",
+  "version": "1.0.0",
```
exploIF, line 3: Not the biggest fan of introducing so many changes together with version bump. I would prefer to have two commits on main (rename and version upgrade)

exploIF (approved):

package.json
szydlovsky: @kacperzolkiewski we intentionally don't want to change the name of the spec, as it would come with a ton of native changes and it's not essential.

README.md
szydlovsky: Ad stays the same. The main image pane WILL change - we're just waiting for it.

PR #640, opened 2026-06-15
fix(android, ios): Link copy-pasting inconsistencies

hejsztynx:
# Summary

There were some inconsistencies when copy-pasting created links

iOS:
- when you add a link via setting input's value or copy-pasting an already existing manual link where `href == textValue`, but the `href` does not match the given `linkRegex`, it gets marked as an autolink, which immediately removes the link, as the regex match fails

Android:
- copying pasting an autolink resulted in creating a manual link unnecessarily 

## Videos

The said iOS bug

https://github.com/user-attachments/assets/cd43e1c9-882a-4c76-a1da-2128176d0f72

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅    |
| Android |    ✅     |

## Checklist

- [x] E2E tests are passing


.maestro/enrichedInput/flows/manual_link_not_matching_set_regex.yaml
```diff
@@ -0,0 +1,17 @@
+appId: swmansion.enriched.example
```
kacperzolkiewski: Could you combine these test cases into a single one with one screenshot? They verify very similar behavior, and we're gradually accumulating more and more test cases.

.maestro/enrichedInput/screenshots/ios/link_auto_detection.png
kacperzolkiewski, line 1: I think that screenshot should be without red underline

ios/htmlParser/HtmlParser.mm
```diff
@@ -415,6 +416,12 @@ + (NSString *_Nullable)initiallyProcessHtml:(NSString *_Nonnull)html
 }
 
 + (NSArray *_Nonnull)getTextAndStylesFromHtml:(NSString *_Nonnull)fixedHtml {
```
kacperzolkiewski: I think that you can delete this function and inside `TextHtmlParser` simply use `[HtmlParser getTextAndStylesFromHtml:fixedHtml config:nil]; `
or 
`[HtmlParser getTextAndStylesFromHtml:fixedHtml config:_view.config];`
If we need config also there?

kacperzolkiewski (commented):
> copying pasting an autolink resulted in creating a manual link unnecessarily

https://github.com/user-attachments/assets/dd0dfc49-192c-4811-b5fb-5b1a3d3eff13

It still recognizes it as a link on Android when we paste it directly next to another letter, until we type something.

android/src/main/java/com/swmansion/enriched/common/parser/EnrichedParser.java
```diff
@@ -862,12 +873,19 @@ private static void startA(Editable text, Attributes attributes) {
     start(text, new Href(href));
   }
 
-  private static <T> void endA(Editable text, T style, EnrichedSpanFactory<T> spanFactory) {
+  private static boolean urlMatchesLinkRegex(String url, Pattern linkRegex) {
+    if (linkRegex == null) return false;
+    Matcher m = linkRegex.matcher(url);
+    return m.find() && m.start() == 0 && m.end() == url.length();
+  }
```
kacperzolkiewski, line 875:
```suggestion
  private static boolean urlMatchesLinkRegex(String url, Pattern linkRegex) {
    if (linkRegex == null) return false;
    return linkRegex.matcher(url).matches();
  }
```

I think that `matches()` should work in this case?

hejsztynx:
> > copying pasting an autolink resulted in creating a manual link unnecessarily
> 
>  Screen.Recording.2026-06-17.at.15.27.31.mov 
> It still recognizes it as a link on Android when we paste it directly next to another letter, until we type something.

fixed, the same issue happened when we would paste text inside a finalized mention - the fix was shared

kacperzolkiewski (approved):
Works well!

ios/EnrichedTextInputView.mm
```diff
@@ -719,6 +719,15 @@ - (void)updateProps:(Props::Shared const &)props
...
 
+  // linkRegex
+  LinkRegexConfig *oldRegexConfig =
+      [[LinkRegexConfig alloc] initWithLinkRegexProp:oldViewProps.linkRegex];
+  LinkRegexConfig *newRegexConfig =
+      [[LinkRegexConfig alloc] initWithLinkRegexProp:newViewProps.linkRegex];
+  if (![newRegexConfig isEqualToConfig:oldRegexConfig]) {
+    [config setLinkRegexConfig:newRegexConfig];
+  }
+
```
szydlovsky, line 730: Any particular reason to move it?

szydlovsky (changes requested):
unsure about the iOS fixes..

ios/htmlParser/HtmlParser.mm
```diff
@@ -747,7 +750,9 @@ + (NSArray *_Nonnull)getTextAndStylesFromHtml:(NSString *_Nonnull)fixedHtml {
       LinkData *linkData = [[LinkData alloc] init];
       linkData.url = url;
       linkData.text = text;
-      linkData.isManual = ![text isEqualToString:url];
+      linkData.isManual = !([text isEqualToString:url] &&
+                            [LinkStyle matchesLinkRegexWithConfig:url
+                                                           config:config]);
```
szydlovsky: Let's just change it to a corrseponding:


```suggestion
      linkData.isManual = ![text isEqualToString:url] || ![LinkStyle matchesLinkRegexWithConfig:url config:config]);
```

the brackets confused me, like a lot

szydlovsky (approved):
Sorry for the mistake with the `isManual` in there, brackets fooled me.

ios/EnrichedTextInputView.mm
szydlovsky: okay makes sense. Default value parsing needs a newer regex config

ios/htmlParser/HtmlParser.mm
```diff
@@ -759,7 +762,7 @@ + (NSArray *_Nonnull)getTextAndStylesFromHtml:(NSString *_Nonnull)fixedHtml {
       LinkData *linkData = [[LinkData alloc] init];
       linkData.url = url;
       linkData.text = text;
-      linkData.isManual = ![text isEqualToString:url];
+      linkData.isManual = ![text isEqualToString:url] || ![LinkStyle matchesLinkRegexWithConfig:url config:config]);
```
exploIF: iOS builds are failing

exploIF (approved):

PR #641, opened 2026-06-16
feat(iOS): custom style

kacperzolkiewski:
# Summary

Add custom style for foreground and background colors handling

## Test Plan
Run the example app and experiment with applying foreground and background colors through the toolbar across different text styles.

## Screenshots / Videos

https://github.com/user-attachments/assets/b922cb42-4911-4285-9385-4c4f6a6d7b9c

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist

- [x] E2E tests are passing


hejsztynx:
Currently you can't see the selection when background color is set

https://github.com/user-attachments/assets/2f9b9a85-365e-49c3-9434-187c0e113cc5

The background and foreground colors are dimmed up in a codeblock, it doesn't happen on other platforms

image

Probably not in this PR's scope, but when text alignment is set and the text wraps, the background color fills up the whole line. This also applies to mentions.

https://github.com/user-attachments/assets/36fe813b-e3b7-47e5-83fc-491f84c79be4

kacperzolkiewski:
@hejsztynx,

1. I applied the same behavior we use for other editor colors (e.g. inside code blocks): if the user doesn't provide an opacity value, one is added automatically.
2. I would leave it like this in first iteration. Android behaves similar to iOS:

image

The problem is not straightforward to resolve. On iOS, custom style backgrounds are applied via `NSBackgroundColorAttributeName` on the text storage, while code block backgrounds are drawn manually as `UIBezierPath` filled rects in `LayoutManagerExtension` one per line, stretched to the full container width. These two layers are completely independent, so making them work together would require splitting the code block rects around the custom-colored ranges.

3. Yes there is a bug but it should be fix in separate PR the same issue occur with inline code:
image

ios/styles/CustomStyle.mm
```diff
@@ -0,0 +1,213 @@
...
+                  return;
+
+                NSMutableDictionary *attrs = [NSMutableDictionary dictionary];
+                if (data.foregroundColor != nil) {
+                  attrs[NSForegroundColorAttributeName] =
+                      [data.foregroundColor colorWithResolvedAlpha];
+                  attrs[NSUnderlineColorAttributeName] =
+                      [data.foregroundColor colorWithResolvedAlpha];
+                  attrs[NSStrikethroughColorAttributeName] =
+                      [data.foregroundColor colorWithResolvedAlpha];
```
hejsztynx: I would not change the opacity of the `foregroundColor`, we don't do that with the inline code style. It doesn't interfere with the selection visibility.

```suggestion
                  attrs[NSForegroundColorAttributeName] =
                      data.foregroundColor;
                  attrs[NSUnderlineColorAttributeName] =
                      data.foregroundColor;
                  attrs[NSStrikethroughColorAttributeName] =
                      data.foregroundColor;
```

hejsztynx (approved):
Functionally looks great!

ios/styles/CustomStyle.mm
kacperzolkiewski: Nice catch!

kacperzolkiewski: Fixed

cpp/tests/GumboParserTest.cpp
szydlovsky, line 1: Maybe a silly question, but do we even run these tests anywhere?

ios/customStyleData/CustomStyleData.h
szydlovsky, line 1: I'd move both the header and implementation files to `/interfaces` catalogue, no need to have them separate like this.

ios/extensions/ColorExtension.h
```diff
@@ -5,4 +5,7 @@
 - (BOOL)isEqualToColor:(UIColor *)otherColor;
 - (UIColor *)colorWithResolvedAlpha;
 - (UIColor *)colorWithResolvedAlpha:(CGFloat)newAlpha;
+- (NSString *)hexString;
+
```
szydlovsky, line 9:
```suggestion
```

ios/styles/CustomStyle.mm
```diff
@@ -0,0 +1,211 @@
...
+- (void)applyStyling:(NSRange)range {
+  if (range.length == 0)
+    return;
+
+  NSUInteger storageLength = self.host.textView.textStorage.length;
+  if (storageLength == 0)
+    return;
+
+  NSRange safeRange = NSMakeRange(
+      range.location, MIN(range.length, storageLength - range.location));
```
szydlovsky, line 42: Why these checks? No other style does this in their `applyStyling` method.

ios/styles/CustomStyle.mm
```diff
@@ -0,0 +1,211 @@
...
+
+- (void)setCustomStyleData:(CustomStyleData *)data
+                     range:(NSRange)range
+                withTyping:(BOOL)withTyping
+            withDirtyRange:(BOOL)withDirtyRange {
+  if (range.length > 0) {
+    if (data == nil || data.isEmpty) {
+      [self remove:range withDirtyRange:withDirtyRange];
+      return;
+    }
```
szydlovsky: Is this an actual used case or a safeguard? I mean; do the call sites of these method even allow for nil/empty data here.

ios/styles/CustomStyle.mm
```diff
@@ -0,0 +1,211 @@
...
+                                           range:range];
+    if (withDirtyRange) {
+      [self.host.attributesManager addDirtyRange:range];
+    }
+  }
+
+  if (withTyping) {
+    if (data == nil || data.isEmpty) {
+      [self removeTyping];
+    } else {
```
szydlovsky: Same here, is it even a possible scenario or rather some overzealous safeguard?

ios/styles/CustomStyle.mm
```diff
@@ -0,0 +1,211 @@
...
+  if (!withTyping) {
+    // Enumerate each existing sub-range and merge the partial update into its
+    // own data so per-character differences (e.g. fg color on some chars) are
+    // preserved when only one field (e.g. bg color) is being changed.
+    NSUInteger storageLength = self.host.textView.textStorage.length;
+    if (storageLength == 0)
+      return;
+
+    NSRange safeRange = NSMakeRange(
+        range.location, MIN(range.length, storageLength - range.location));
```
szydlovsky, line 176: Looks like a super overzealous check, not needed here and the call site should guarantee that the range is proper.

ios/utils/StyleUtils.mm
```diff
@@ -91,7 +91,8 @@ + (NSDictionary *)conflictMap {
       @([CheckboxListStyle getType])
     ],
     @([ImageStyle getType]) :
-        @[ @([LinkStyle getType]), @([MentionStyle getType]) ]
+        @[ @([LinkStyle getType]), @([MentionStyle getType]) ],
+    @([CustomStyle getType]) : @[]
```
szydlovsky, line 95: why do images conflict with the fg/bg color? I think there's nothing stopping these working together.

ios/utils/StyleUtils.mm
```diff
@@ -123,23 +124,35 @@ + (NSDictionary *)blockingMap {
     @([AlignmentStyle getType]) : @[],
     @([BlockQuoteStyle getType]) : @[],
     @([CodeBlockStyle getType]) : @[],
-    @([ImageStyle getType]) : @[ @([InlineCodeStyle getType]) ]
+    @([ImageStyle getType]) : @[ @([InlineCodeStyle getType]) ],
+    @([CustomStyle getType]) : @[]
```
szydlovsky, line 128: Here as well, I don't see why image can't be used within fg/bg color.

ios/EnrichedTextInputView.mm
```diff
@@ -1920,8 +2003,9 @@ - (bool)textView:(UITextView *)textView
             replacementText:(NSString *)text {
   // Capture the attributes at range.location that are being replaced
   // (autocorrect / predictive) so didProcessEditing: can re-stamp them onto the
-  // replacement. Only capture for genuine replacements (text.length > 0), not
-  // for deletions/backspace (text.length == 0).
+  // replacement. Skip pure deletions (text.length == 0) — there is no incoming
+  // text to receive attributes, and capturing here would cause the deleted
+  // character's CustomStyleData to be re-stamped onto the widened editedRange.
```
szydlovsky, line 2008: That comment change is too specific and actually such issue touches all the styles - not only the custom style. I'd rather we stayed with the old comment here.

src/types.ts
```diff
@@ -1,13 +1,13 @@
...
-  ViewProps,
+import {
+  type ColorValue,
+  type DimensionValue,
+  type NativeMethods,
+  type NativeSyntheticEvent,
+  type ReturnKeyTypeOptions,
+  type TargetedEvent,
+  type TextStyle,
+  type ViewProps,
```
szydlovsky, line 10: Is that a linter change or intentional? I think the previous looks better..

ios/EnrichedTextInputView.mm
szydlovsky, line 1: I think the way we decide the style value over a range is a bit wrong. IMO:
- when there is only ONE value of fg/bg color within the range, then we return this value
- when there is more than one value, we should IMO just return `nil`

This would at least align with how we decide it for other styles, e.g. an inline style is shown as active only if it spans the whole selected range without any exceptions.

Image

Image

szydlovsky (changes requested):

cpp/tests/GumboParserTest.cpp
kacperzolkiewski: Yes we run them if something changed inside parser here is an action for that: https://github.com/software-mansion/react-native-enriched-html/actions/workflows/test-cpp.yml

ios/utils/StyleUtils.mm
kacperzolkiewski: Wait, but images don't have a CustomStyle set as conflicting/blocking?

ios/utils/StyleUtils.mm
kacperzolkiewski: As above

src/types.ts
kacperzolkiewski: linter change, fixed

ios/EnrichedTextInputView.mm
kacperzolkiewski: I restored old comment

ios/styles/CustomStyle.mm
kacperzolkiewski: I deleted those check, I think that you are sure we do not need them

ios/styles/CustomStyle.mm
kacperzolkiewski: I deleted those checks as above

ios/interfaces/CustomStyleData.h
kacperzolkiewski: Moved

ios/EnrichedTextInputView.mm
kacperzolkiewski: Fixed:
image

image

PR #644, opened 2026-06-16
fix: disable autocorrection in mobile e2e tests

hejsztynx:
# Summary

Disabled the autocorrection in e2e maestro testing devices to prevent unexpected behavior.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅    |

## Checklist

- [x] E2E tests are passing


szydlovsky (approved):
Looks alright!

.maestro/scripts/run-tests.sh
```diff
@@ -144,6 +144,11 @@ run_maestro() {
   return "$rc"
 }
 
+if [ "$PLATFORM" = ios ]; then
+  echo "=== Running iOS device setup ==="
+  run_maestro --device "$DEVICE_ID" "$MAESTRO_ROOT/setup/ios_keyboard_settings.yaml"
+fi
```
kacperzolkiewski: Shouldn't we call this only when the rebuild flag is enabled? Right now, every time we run the test cases, it also executes the flow responsible for disabling autocorrection.

hejsztynx: now we use `pListBuddy` instead of a setup flow

kacperzolkiewski (approved):
Works properly!

PR #648, opened 2026-06-18
feat(android): custom style

kacperzolkiewski:
# Summary

Add custom style to support foreground and background colors

## Test Plan
Run the example app and experiment with applying foreground and background colors through the toolbar across different text styles.

## Screenshots / Videos

https://github.com/user-attachments/assets/d3ae479f-b274-415a-9bec-a0fc39e7d9cd

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

## Checklist

- [x] E2E tests are passing


hejsztynx:
Currently you can't see the selection and the cursor inside a text with some background color set


https://github.com/user-attachments/assets/9de6d860-e7f6-46b1-8f61-bad62cc7796c

kacperzolkiewski:
> Currently you can't see the selection and the cursor inside a text with some background color set

@hejsztynx, 
I applied the same behavior we use for other editor colors (e.g. inside code blocks): if the user doesn't provide an opacity value, one is added automatically.

hejsztynx (approved):
Functionally looks great!

android/src/main/java/com/swmansion/enriched/common/spans/EnrichedCustomStyleSpan.kt
```diff
@@ -0,0 +1,30 @@
...
+  private val foregroundColor: Int?,
+  private val backgroundColor: Int?,
+) : CharacterStyle(),
+  EnrichedInlineSpan {
+  fun getForegroundColor(): Int? = foregroundColor
+
+  fun getBackgroundColor(): Int? = backgroundColor
+
+  override fun updateDrawState(textPaint: TextPaint) {
+    foregroundColor?.let { textPaint.color = withOpacity(it, 80) }
```
hejsztynx: I wouldn't change the opacity of the `foregroundColor` though, it doesn't interfere with selection or cursor visibility. Also maybe let's use already available `parseColorWithOpacity` in the input's `HtmlStyle.kt` and text's `TextStyle.kt`, and apply the opacity during props parsing - consistent with how the inline code style's `backgroundColor` is computed.

```suggestion
    foregroundColor?.let { textPaint.color = it }
```

hejsztynx (changes requested):

android/src/main/java/com/swmansion/enriched/common/spans/EnrichedCustomStyleSpan.kt
kacperzolkiewski: Nice catch!

kacperzolkiewski: Fixed

hejsztynx (approved):

PR #651, opened 2026-06-21
feat(web): enriched text component

hejsztynx:
# Summary

The `EnrichedText` component has been implemented with the support of basic props:

- `style`
- `htmlStyle`
-  `children`
-  `selectionColor`

The core displaying feature of the component is fully functional and should be identical to the `EnrichedTextInput`. The `EnrichedText` specific parts of `htmlStyle` is not yet implemented, as these are _press_ specific actions - outside of this PR's scope. 'htmlStyle` is currently only handled on the shared part with the `EnrichedTextInput`.

As we are using `dangerouslySetInnerHtml`, proper sanitization of the html was provided using `dompurify`.

## Test Plan

On the root page a "Set Enriched Text" button was added to easily test the component. You write something in the input, press the button to apply the html to the `EnrichedText` component and the visible content should look identical

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| Web |    ✅    |

## Checklist

- [x] E2E tests are passing


src/web/normalization/prepareHtmlForWeb.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  doc.querySelectorAll('ul[data-type="checkbox"]').forEach((ul) => {
+    ul.querySelectorAll('li').forEach((li) => {
+      const checked = li.hasAttribute('checked');
+      const labelContent = li.innerHTML;
+
+      li.removeAttribute('checked');
+      li.innerHTML =
+        `<input type="checkbox"${checked ? ' checked' : ''}>` +
+        `<label>${labelContent}</label>`;
```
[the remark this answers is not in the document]
hejsztynx: this will  be handled in the pr with the other press events

src/web/normalization/prepareHtmlForWeb.ts
```diff
@@ -0,0 +1,43 @@
...
+ *       <input type="checkbox" checked>
+ *       <label>foo</label>
+ *     </li>
+ *     <li>
+ *       <input type="checkbox">
+ *       <label>bar</label>
+ *     </li>
+ *   </ul>
+ */
+function checkboxHtmlToWeb(html: string): string {
```
kacperzolkiewski, line 24: I wonder if we could maybe use the normalizer we already have for checkboxes: `checkboxHtmlForTiptap`?

kacperzolkiewski:
1. Broken images does not render properly inside `EnrichedText`:

https://github.com/user-attachments/assets/4a34edf6-49d6-4ff7-9c05-b86f9588789d

2. When we press the link, the page re-renders. I think it shouldn’t do anything until `onLinkPress` is implemented.

https://github.com/user-attachments/assets/654ee927-58a6-4a13-b5a0-e4dffad592d3

3. I think that it would be good to clear input in example app when we push html to `EnrichedText` like in native example app

src/web/normalization/prepareHtmlForWeb.ts
hejsztynx: Probably not, `tiptap` expects `data-checked` attribute where on web we need the `checked` attribute in the `input` tag - schema differs. More importantly `checkboxHtmlForTiptap` does not render the `input` and `label` tags needed to natively render the checkboxes on web.

.playwright/tests/enrichedTextVisual.spec.ts
```diff
@@ -0,0 +1,235 @@
...
+        '<html>',
+        '<ul><li>Bullet item</li></ul>',
+        '<ol><li>Numbered item</li></ol>',
+        '<ul data-type="checkbox"><li checked>Checked item</li><li>Unchecked item</li></ul>',
+        '</html>',
+      ].join(''),
+    },
+  ];
+
+  for (const c of cases) {
```
kacperzolkiewski: NIT: loop is redundant here because we have only one case in this test case

src/web/styleConversion/htmlStyleToCSSVariables.ts
```diff
@@ -12,65 +20,106 @@ import { toColor } from './toColor';
 import { isMentionStyleRecord } from '../../utils/isMentionStyleRecord';
 
 export function mergeWithDefaultHtmlStyle(
```
kacperzolkiewski, line 22: NIt: Right now it not always merge htmlStyle with `DEFAULT_HTML_STYLE`, I would change the name to something like in native: `assignDefaultValues`

kacperzolkiewski (approved):
Works properly!

docs/WEB.md
```diff
@@ -18,20 +20,33 @@ Web support is still experimental. APIs and behavior can change in future releas
...
+
+- Customizing the styling using props: `style`, `htmlStyle`, `selectionColor`.
+
+### Unsupported
+
+- **`selectable`**: ignored on web.
+- **`useHtmlNormalizer`**: ignored on web.
+- **`ellipsizeMode`**: ignored on web.
+- **`numberOfLines`**: ignored on web.
+- **Press events**: `onLinkPress` and `onMentionPress` callbacks are ignored on web.
```
exploIF, line 48: What's the plan for this props? Is it something that we implement later on?

src/web/EnrichedText.tsx
```diff
@@ -0,0 +1,82 @@
...
+import { ENRICHED_TEXT_CLASSNAME } from './constants/classNames';
+import { enrichedInputThemingToCSSProperties } from './styleConversion/enrichedThemingToCSSProperties';
+import { buildMentionRulesCSS } from './styleConversion/buildMentionRulesCSS';
+import { sanitizeHtml } from './sanitization/htmlSanitizer';
+import { prepareHtmlForWeb } from './normalization/prepareHtmlForWeb';
+import { INLINE_IMAGE_CSS_VARIABLES } from './styleConversion/inlineImageCSSVariables';
+import { useImageErrorFallback } from './useImageErrorFallback';
+import { usePressInteractions } from './usePressInteractions';
+
+export const EnrichedText = memo(
```
exploIF, line 18: Do we really need `memo` there? Nowdays, most of the projects should already use react-compiler, so I don't think any manual memoization inside the library it's needed.

src/web/useImageErrorFallback.ts
```diff
@@ -0,0 +1,42 @@
...
+
+    container.addEventListener('error', handleImageError, true);
+
+    return () => {
+      container.removeEventListener('error', handleImageError, true);
+    };
+  }, [containerRef]);
+
+  // handle the cached unloaded images that will not emit the error event
+  useEffect(() => {
```
exploIF: Is it on purpose that there is no dependency array in this effect? If so, why it's done like that?

src/web/usePressInteractions.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+      const anchor = target.closest('a');
+      if (anchor && container.contains(anchor)) {
+        e.preventDefault();
+      }
+    };
+
+    container.addEventListener('click', handleClick);
+    return () => container.removeEventListener('click', handleClick);
+  }, [containerRef]);
```
exploIF, line 21: Is it needed? Ref should be stable, so I guess it can be removed. Unless eslint complains

package.json
```diff
@@ -105,7 +105,8 @@
     "@tiptap/extension-underline": "3.20.4",
     "@tiptap/extensions": "3.20.4",
     "@tiptap/pm": "3.20.4",
-    "@tiptap/react": "3.20.4"
+    "@tiptap/react": "3.20.4",
+    "dompurify": "^3.4.11"
```
exploIF: Let's pin this version


```suggestion
    "dompurify": "3.4.11"
```

src/web/sanitization/htmlSanitizer.ts
```diff
@@ -0,0 +1,8 @@
+import DOMPurify from 'dompurify';
+
+export function sanitizeHtml(html: string) {
+  return DOMPurify.sanitize(html, {
+    ADD_TAGS: ['mention', 'codeblock'],
+    ADD_ATTR: ['text', 'indicator'],
```
exploIF, line 6: I wonder if we really need custom attributes here 🤔   But if we need them, keep in mind that mentions allow setting custom attributes

docs/WEB.md
hejsztynx: Yes, their implementation will come in the next PRs

src/web/EnrichedText.tsx
hejsztynx: From what I've read, projects typically have React Compiler set up in such way that it doesn't touch `node_modules` and without `memo` there's currently a bug where on every rerender ``'s fetches will rerun which will cause the image placeholder to flicker, so I think it's better to keep it as it is.

src/web/useImageErrorFallback.ts
hejsztynx: I revisited this logic and refined it. Removed that `useEffect` and put its logic into the first one, as we only need it when component initially mounts.

src/web/usePressInteractions.tsx
hejsztynx: yes, eslint complains

src/web/sanitization/htmlSanitizer.ts
hejsztynx: As discussed, the sanitization of those custom attributes needs to be researched and implemented. I would consider it to be outside of this PR's scope, let's keep it as it is, so core mention functionality works

exploIF (approved):

PR #652, opened 2026-06-22
feat: add getCaretRect() to read the caret bounding rect

dennistan95:
## Summary

Adds an async `getCaretRect()` to the `EnrichedTextInput` imperative handle so consumers can anchor UI — link popovers, floating toolbars, mention menus — to the live caret.

`onChangeSelection` exposes only character offsets, so there's currently no way to position an overlay at the caret. Native already has the pixel geometry (`UITextView caretRectForPosition:` on iOS, `Layout` on Android); this just surfaces it to JS, mirroring the existing `getHTML()` / `requestHTML` request→response round-trip exactly.

## API

```ts
interface CaretRect { x: number; y: number; width: number; height: number; }

// on EnrichedTextInputInstance:
getCaretRect: () => Promise<CaretRect | null>;
```

Resolves the caret rect in the editor view's coordinate space (RN dp / iOS points), or `null` when unavailable. A built-in timeout resolves `null` rather than leaking a pending promise, so callers always get a definite result.

```tsx
const rect = await ref.current?.getCaretRect();
if (rect) setPopover({ left: rect.x, top: rect.y - POPOVER_HEIGHT }); // e.g. above the caret
```

## Changes

| Layer | Change |
|---|---|
| `src/spec/EnrichedTextInputNativeComponent.ts` | `OnCaretRectEvent` event, `onCaretRect` prop, `requestCaretRect` command (+ `supportedCommands`) |
| `src/types.ts` | `CaretRect` type + `getCaretRect()` on `EnrichedTextInputInstance` |
| `src/native/EnrichedTextInput.tsx` | `getCaretRect()` (requestId→promise with timeout fallback), `onCaretRect` handler, unmount cleanup |
| `ios/EnrichedTextInputView.mm` | `requestCaretRect:` → `caretRectForPosition:` converted into the component view's space; emits a `valid` flag when the rect is null/infinite |
| `android/.../EnrichedTextInputView.kt` | `requestCaretRect()` → `Layout` primary-horizontal + line top/bottom, padding/scroll-adjusted, px→dp for parity with iOS points |
| `android/.../EnrichedTextInputViewManager.kt` | `requestCaretRect` override + event registration |
| `android/.../events/OnCaretRectEvent.kt` | new direct event (`requestId`, `x`, `y`, `width`, `height`, `valid`) |

Both platforms emit a `valid` flag so JS can distinguish "no caret / unknown" from a real zero-origin rect.

## Note for maintainers

This package ships pre-generated Fabric codegen (`includesGeneratedCode: true`) and a compiled `lib/`. After review, the generated artifacts (`ios/generated`, `android/generated`) and `lib/` will need regenerating (codegen + `bob` build) before a release picks up the new command/event.

## Testing

- iOS: caret rect tracks the insertion point across lines/wrapping; `valid = false` with no selection.
- Android: rect matches iOS within rounding (dp-normalized).
- Degradation: on a build without the native side, `getCaretRect()` resolves `null` via the timeout (no hang).

### DIFF
--- a/android/src/main/java/com/swmansion/enriched/textinput/EnrichedTextInputView.kt
+++ b/android/src/main/java/com/swmansion/enriched/textinput/EnrichedTextInputView.kt
@@ -1025,7 +1025,10 @@ class EnrichedTextInputView :
     var widthDp = 0.0
     var heightDp = 0.0
     val currentLayout = layout
-    val pos = selectionStart
+    // Anchor to the selection END (the active insertion point), matching iOS'
+    // selectedTextRange.end — so a non-empty selection reports the live caret,
+    // not the start of the range.
+    val pos = selectionEnd
     if (currentLayout != null && pos >= 0) {
       try {
         val line = currentLayout.getLineForOffset(pos)
--- a/android/src/main/java/com/swmansion/enriched/textinput/EnrichedTextInputView.kt
+++ b/android/src/main/java/com/swmansion/enriched/textinput/EnrichedTextInputView.kt
@@ -43,6 +43,7 @@ import com.swmansion.enriched.common.GumboNormalizer
 import com.swmansion.enriched.common.parser.EnrichedParser
 import com.swmansion.enriched.common.pixelFromSpOrDp
 import com.swmansion.enriched.textinput.events.MentionHandler
+import com.swmansion.enriched.textinput.events.OnCaretRectEvent
 import com.swmansion.enriched.textinput.events.OnContextMenuItemPressEvent
 import com.swmansion.enriched.textinput.events.OnInputBlurEvent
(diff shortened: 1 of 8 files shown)


dennistan95:
Thanks for the review — all three addressed in the latest commit:

1. **`setCustomSelection` change** — that was unrelated and slipped in from a packaging diff; reverted to the original `getActualIndex` (ZWSP-aware) mapping.
2. **Android anchoring to `selectionStart`** — switched to `selectionEnd` so it reports the active caret/insertion point, matching iOS `selectedTextRange.end`.
3. **Web target missing `getCaretRect`** — implemented it via ProseMirror `coordsAtPos(selection.to)`, converted to editor-element coordinates, so the cross-platform `EnrichedTextInputInstance` contract is fully satisfied.

PR #653, opened 2026-06-22
feat(web): custom style

kacperzolkiewski:
# Summary

Add custom style for foreground and background colors handling

## Test Plan

- Run the example web app and experiment with applying foreground and background colors through the toolbar across different text styles.
- Run yarn test:e2e:web to see if all the tests pass

## Screenshots / Videos

https://github.com/user-attachments/assets/5490888b-7f2b-431d-99b0-36c15da1fd7d

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ❌     |
| Web |    ✅     |

## Checklist

- [X] E2E tests are passing


hejsztynx:
1. When we set the background color of a `<code>` the colors clip

image

Also maybe they sohuld be overwritten? Just like it happens in `blockquote` or `codeblock`. Or the inline styles should not be overwritten? (mention, link). I'm just checking if this is intentional.

2. When there's a space in the beginning of a foreground or a background style, on backspace, both styles are stripped

https://github.com/user-attachments/assets/703e666e-607d-4b2a-8793-a65b4bb954f1

src/web/EnrichedTextInput.tsx
```diff
@@ -368,7 +370,29 @@ export const EnrichedTextInput = ({
...
+        runFocused(editor, (c) => {
+          if (!resolvedColor && !resolvedBg) {
+            return c.unsetCustomStyle();
+          }
+
+          return c.setCustomStyle({
+            foregroundColor: resolvedColor,
+            backgroundColor: resolvedBg,
+          });
+        });
```
hejsztynx: Looks like this logic is duplicated. The same thing happens inside `setCustomStyle` in `EnrichedCustomStyle.ts`. We should probably remove that logic here and let the plugin handle it.

hejsztynx (changes requested):

src/web/EnrichedTextInput.tsx
kacperzolkiewski: Good catch! fixed

kacperzolkiewski:
> 1. When we set the background color of a `<code>` the colors clip
> 
> image
> Also maybe they sohuld be overwritten? Just like it happens in `blockquote` or `codeblock`. Or the inline styles should not be overwritten? (mention, link). I'm just checking if this is intentional.
> 
> 2. When there's a space in the beginning of a foreground or a background style, on backspace, both styles are stripped
> 
>  Screen.Recording.2026-06-25.at.13.56.58.mov

1. Custom styles have higher priority than paragraph styles, but lower priority than inline styles. This is why they override paragraph styles but do not override inline styles.
2. Fixed.

src/web/formats/EnrichedCustomStyle.ts
```diff
@@ -0,0 +1,121 @@
...
+              foregroundColor: resolvedColor ?? null,
+              backgroundColor: resolvedBg ?? null,
+            })
+            .run();
+        },
+
+      unsetCustomStyle:
+        () =>
+        ({ chain }) =>
+          chain().unsetMark('customStyle').run(),
```
hejsztynx: I think this can be now safely removed

hejsztynx (approved):
Works great!

PR #654, opened 2026-06-24
feat(web): text alignment

kacperzolkiewski:
# Summary

Text alignment for Web

## Test Plan

- Run example web app and play with `textAlignment` when different styles are toggled.
- Check if paragraphs/lists with different `textAlignment` than `auto` are properly parsed and contain `style="text-align:center|right|justify"` inside HTML
- Run yarn test:e2e:web to see if all the tests pass

## Screenshots / Videos

https://github.com/user-attachments/assets/92a4b539-6ff0-423b-af68-3bc2f3f787a5

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ❌     |
| Web |    ✅     |

## Checklist

- [x] E2E tests are passing


hejsztynx:
https://github.com/user-attachments/assets/777167e7-0830-4eac-97fc-1a47b945df98

When we select all content with `command + A`, then toggle a paragraph style, the alignment isn't persisted

hejsztynx:
https://github.com/user-attachments/assets/cd40ac5d-3a10-4676-8341-f8c14056e9ee

It looks like the issue is still there - again when using `cmd + A`, but only when switching from `blockquote` or `codeblock` to a different paragraph style.

hejsztynx (approved):

hejsztynx:
Works flawlessly 🚀

apps/example-web/src/components/Toolbar.tsx
```diff
@@ -218,6 +228,18 @@ export function Toolbar({
             }}
           />
         ))}
+        {alignmentItems.map((item) => (
+          <ToolbarButton
+            key={`alignment-${item.value}`}
+            label={item.label}
+            testId={`toolbar-button-alignment-${item.value}`}
+            isActive={s?.alignment === item.value}
+            isDisabled={false}
```
exploIF: Let's make it optional and don't pass at all when false

exploIF (approved):

PR #662, opened 2026-06-26
fix(iOS): rendering images from html with link

### LINKED ISSUE #661: Image before link crashes renderer
**Describe the bug**

Using react-native-enriched, email that feature an image before a link crash the renderer.

**To Reproduce**

Pass this html into `EnrichedTextInput`
```
<html><img src="data:," /><img src="data:," /><img src="data:," /><p>See <a href="https://example.com">a link</a></p></html>
```

Get this crash
```
*** Terminating app due to uncaught exception 'NSInvalidArgumentException',
    reason: '-[__NSCFString substringWithRange:]: Range {7, 6} out of bounds; string length 10'
```

Root cause: HtmlParser.mm:698 does [plainText substringWithRange:tagRangeValue.rangeValue] where the range was built at :166 in final-NSTextStorage coords (counts +1 per ), but plainText doesn’t contain the image placeholders. Any link finalized while precedingImageCount > 0 overruns plainText by that count. Triggered in our app on a real ~5.3KB email with 3 images and a trailing link (overrun by 3, matching the math). Fix is a one-site clamp before :698.

**Expected behavior**
HTML should not crash

kacperzolkiewski:
# Summary

Fixes: #661 

## Test Plan

Insert html by `Set input's value` button into example app:
`<html><p></p><p>See  <a href="https://example.com">a link</a></p></html>`
The html should be parsed properly.

## Screenshots / Videos
before:

after:

https://github.com/user-attachments/assets/eb79f199-a5d5-4edc-ba4d-cc65074644aa

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |
| Web     |    ❌     |

## Checklist

- [x] E2E tests are passing

### DIFF
--- a/ios/htmlParser/HtmlParser.mm
+++ b/ios/htmlParser/HtmlParser.mm
@@ -645,6 +645,9 @@
 
   // process tags into proper StyleType + StylePair values
   NSMutableArray *processedStyles = [[NSMutableArray alloc] init];
+  // Tracks the number of processed images to remove their pre-generated
+  // placeholder offsets from tag ranges when reading from plainText
+  // (which does not contain those placeholders).
   NSInteger secondPassImageCount = 0;
 
   for (NSArray *arr in initiallyProcessedTags) {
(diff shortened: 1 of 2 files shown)


ios/htmlParser/HtmlParser.mm
```diff
@@ -645,6 +645,7 @@ + (NSArray *_Nonnull)getTextAndStylesFromHtml:(NSString *_Nonnull)fixedHtml {
 
   // process tags into proper StyleType + StylePair values
   NSMutableArray *processedStyles = [[NSMutableArray alloc] init];
+  NSInteger secondPassImageCount = 0;
```
szydlovsky, line 651: Can we add some comment what it exactly counts as the "second pass" and what it's used for

ios/htmlParser/HtmlParser.mm
```diff
@@ -742,7 +744,23 @@ + (NSArray *_Nonnull)getTextAndStylesFromHtml:(NSString *_Nonnull)fixedHtml {
...
+      NSRange adjustedRange = tagRangeValue.rangeValue;
+      // Calculate the shifted location, ensuring it doesn't drop below 0
+      NSUInteger rawLoc = (adjustedRange.location >= secondPassImageCount)
+                              ? adjustedRange.location - secondPassImageCount
+                              : 0;
+      // Cap the location so it can never exceed the plain text length
+      NSUInteger plainLoc = MIN(rawLoc, plainText.length);
+      // Cap the length based on whatever space is left in the string
+      NSUInteger plainLen =
+          MIN(adjustedRange.length, plainText.length - plainLoc);
```
szydlovsky: Why do we even need to cap these? The new, plain-text oriented range is anyway being pushed to the left by the number of images. If we somehow still exceed plainText length, then it's another bug in the parser and this capping just effectively hides it :/

ios/htmlParser/HtmlParser.mm
kacperzolkiewski: added

ios/htmlParser/HtmlParser.mm
kacperzolkiewski: removed

szydlovsky (approved):

PR #665, opened 2026-06-30
feat(web): textShortcuts support

kacperzolkiewski:
# Summary

Add `textShortcuts` support for web

## Test Plan

-  Verify paragraph shortcuts (headings, blockquote, code block, lists)
-  Verify inline shortcuts (bold, italic, code, strikethrough)
-  Verify default behavior (no prop = -  and 1.  shortcuts work)
-  Verify empty array disables all shortcuts

## Screenshots / Videos

https://github.com/user-attachments/assets/ee0b09fc-cefe-463d-878e-8c93e5f73d95

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ❌     |
| Web     |    ✅     |

## Checklist

- [X] E2E tests are passing


src/web/pmPlugins/TextShortcutsPlugin.ts
```diff
@@ -0,0 +1,327 @@
...
+    const contentLength = closeDelimPrefixDocStart - contentDocStart;
+    const finalStart = openDocStart;
+    const finalEnd = openDocStart + contentLength;
+    tr.addMark(finalStart, finalEnd, markType.create());
+
+    // place cursor at end of content
+    tr.setSelection(TextSelection.create(tr.doc, finalEnd));
+
+    view.dispatch(tr);
+    view.dispatch(view.state.tr.setStoredMarks([]));
```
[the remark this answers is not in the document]
kacperzolkiewski: It did not work like that properly

hejsztynx (changes requested):
Currently applying shortcut styles, strips the current inline styles

https://github.com/user-attachments/assets/8521a551-33fd-4d8e-9965-ffafac87fcb2

hejsztynx (approved):
Works great!

docs/WEB.md
```diff
@@ -18,6 +18,7 @@ Web support is still experimental. APIs and behavior can change in future releas
 - Input theming via `placeholderTextColor`, `cursorColor` and `selectionColor` props
 - Keyboard shortcuts for formatting
 - `useHtmlNormalizer`
+- `textShortcuts`
```
szydlovsky, line 24: Might be a long shot here, but we could maybe link all of these features to their corresponding sections in the api reference?

kacperzolkiewski: I’d rather not do that for now, since we’re planning to improve the documentation in the near future.

src/web/pmPlugins/TextShortcutsPlugin.ts
```diff
@@ -0,0 +1,337 @@
...
+    // Trigger must be anchored at paragraph start
+    if (offsetInBlock !== prefixLen) continue;
+
+    // Verify the prefix characters already in the doc match the trigger prefix
+    if (prefixLen > 0) {
+      const docPrefix = view.state.doc.textBetween(blockStart, from);
+      if (docPrefix !== trigger.slice(0, prefixLen)) continue;
+    }
+
+    if (isFormatBlocked(style, editor, htmlStyle)) continue;
```
szydlovsky: The `style` argument here is incompatible with the method, e.g. the `isFormatBlocked` expects `codeBlock` but the passed `TextShortcutStyle` type has `codeblock`. We should ideally create same "translation table" for the paragraph styles, as the inline styles have now - `INLINE_MARK_NAME`. Or maybe even a shared table. 

I know that we currently have no active blocks for paragraph styles but it could change in the future and then this code would be a source of issues because of misaligned string format.

szydlovsky (commented):
Found one more thing that I'm kinda uncertain about; do we always check only for the previous shortcut delimiters? As you can see on the recording here, only the trailing ones actually work and I'm unsure if it's expected (what do mobile platforms do?):

https://github.com/user-attachments/assets/b17a1b0c-6666-4918-8475-3e4ae08933fa

kacperzolkiewski:
> Found one more thing that I'm kinda uncertain about; do we always check only for the previous shortcut delimiters? As you can see on the recording here, only the trailing ones actually work and I'm unsure if it's expected (what do mobile platforms do?):

@szydlovsky The mobile implementations work the same way. When implementing this feature, I used Slack's mobile implementation as a reference, and it behaves the same way.

src/web/pmPlugins/TextShortcutsPlugin.ts
kacperzolkiewski: fixed

szydlovsky (approved):

apps/example-web/src/testScreens/VisualRegression.tsx
```diff
@@ -49,6 +51,17 @@ export function VisualRegression() {
     }
   }, [htmlStyleOverrideJson]);
 
+  const textShortcuts = useMemo<TextShortcut[] | undefined>(() => {
+    const raw = textShortcutsJson.trim();
+    if (!raw) return undefined;
+    try {
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-return
+      return JSON.parse(raw) as TextShortcut[];
```
exploIF: Could you elaborate why eslint is complaining here and why it's safe to do it that way?

docs/WEB.md
exploIF, line 7: Before making web stable we should completely remove this section, as by default everything should work (other than known limitations)

apps/example-web/src/testScreens/VisualRegression.tsx
kacperzolkiewski: It's inside the `example-web` app. I noticed an eslint warning, so I thought we could simply disable eslint for that line since it's only used in our internal example app.

kacperzolkiewski: image

Problem solved. ESLint no longer reports any issues :D

docs/WEB.md
kacperzolkiewski: I save the note about it

exploIF (approved):

PR #668, opened 2026-07-02
fix: checkbox list normalization and stripped empty list elements

hejsztynx:
# Summary

There were two issues:

1. The normalizer didn't understand the Tiptap's checkbox list structure and the Google Docs' or MS Word's one, so when pasting their HTML, the lists were broken. This also included a case where we just copy-pasted a checkbox list internally in the web's `EnrichedTextInput` - the list became a normal unordered list.
2. The empty `<li></li>` were stripped by the normalizer, which was incorrect, as we allow this structure.

## Test Plan

- Write some checkbox lists in the web `EnrichedTextInput` and copy-paste them. The pasted list should be parsed and displayed correctly now.
- Go to Google Docs and MS Word, create some checkbox lists and paste them into the `EnrichedTextInput` on a chosen platform. They should be correctly parsed now.
- Try creating a list with empty elements on the web and copy-paste it to the same input or to an input on a mobile platform. The empty elements should not be stripped now.

## Screenshots / Videos

The internal Tiptap copy-paste issue

Before:

https://github.com/user-attachments/assets/db47c1bd-c0ae-481d-9389-1f2fd3a62240

After:

https://github.com/user-attachments/assets/c7033f9d-9c5e-465b-b9c5-5816b7346de1

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅    |
| Android |    ✅     |
| Web     |    ✅     |

## Checklist

- [x] E2E tests are passing

### DIFF
--- a/apps/example/ios/Podfile.lock
+++ b/apps/example/ios/Podfile.lock
@@ -2026,7 +2026,7 @@ EXTERNAL SOURCES:
 
 SPEC CHECKSUMS:
   FBLazyVector: c00c20551d40126351a6783c47ce75f5b374851b
-  hermes-engine: b4dad6ba67535bb03c8ff1006b337cba14db16cb
+  hermes-engine: d80056af9dcb5ba3db80f37943f96751037c7381
   RCTDeprecation: 3bb167081b134461cfeb875ff7ae1945f8635257
   RCTRequired: 74839f55d5058a133a0bc4569b0afec750957f64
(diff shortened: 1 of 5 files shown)


kacperzolkiewski (approved):
Works properly!

szydlovsky (approved):
Ship it!

exploIF (approved):

PR #677, opened 2026-07-06
feat(web): EnrichedText selectable, useHtmlNormalizer, ref, onFocus, onBlur

hejsztynx:
# Summary

Implemented following props:

- `selectable`
- `useHtmlNormalizer`
- native `ref` methods, where only `focus` and `blur` are functional, the rest of them are no-op
- `onFocus` and `onBlur` callback props

Added `EnrichedTextInstance` to the exported API. Refactored the web example app.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ❌     |
| Web     |    ✅     |

## Checklist

- [x] E2E tests are passing


src/web/EnrichedText.tsx
```diff
@@ -1,4 +1,10 @@
-import { memo, useMemo, useRef, type CSSProperties } from 'react';
+import {
+  memo,
+  useImperativeHandle,
+  useMemo,
+  useRef,
+  type CSSProperties,
+} from 'react';
 import type { EnrichedTextProps } from '../types';
 import './EnrichedText.css';
```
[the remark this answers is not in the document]
hejsztynx: that's only relevant to older React versions

src/web/EnrichedText.tsx
```diff
@@ -72,8 +102,19 @@ export const EnrichedText = memo(
...
         <div
           ref={containerRef}
+          tabIndex={-1}
           style={finalStyle}
           className={ENRICHED_TEXT_CLASSNAME}
+          onFocus={
+            onFocus
+              ? (event) => onFocus(adaptWebToNativeEvent(event, { target: -1 }))
+              : undefined
+          }
```
exploIF:
```suggestion
          onFocus={(event) => onFocus?.(adaptWebToNativeEvent(event, { target: -1 }))}
```

src/web/EnrichedText.tsx
```diff
@@ -72,8 +102,19 @@ export const EnrichedText = memo(
...
+          onBlur={
+            onBlur
+              ? (event) => onBlur(adaptWebToNativeEvent(event, { target: -1 }))
+              : undefined
+          }
```
exploIF: Same here, it can be simplified

src/web/EnrichedText.tsx
```diff
@@ -72,8 +102,19 @@ export const EnrichedText = memo(
         {mentionRulesCSS ? <style>{mentionRulesCSS}</style> : null}
         <div
           ref={containerRef}
+          tabIndex={-1}
```
exploIF, line 109: Why it's needed?

exploIF (approved):

src/web/styleConversion/enrichedThemingToCSSProperties.ts
```diff
@@ -20,15 +29,28 @@ export function enrichedInputThemingToCSSProperties({
...
+  selectable?: boolean;
+}
+
+export function enrichedTextThemingToCSSProperties({
+  selectionColor,
+  selectable,
+}: EnrichedTextThemingOptions): CSSProperties {
+  const extra: EnrichedThemingStyle = {};
+
+  if (selectable !== undefined) extra.userSelect = selectable ? 'text' : 'none';
```
kacperzolkiewski, line 53: Shouldn't `extra.userSelect` also be set to `none` when selectable is `undefined`?

src/web/EnrichedText.tsx
```diff
@@ -7,23 +13,47 @@ import {
...
+    onFocus,
+    onBlur,
+  }: EnrichedTextProps) => {
     const containerRef = useRef<HTMLDivElement>(null);
 
+    useImperativeHandle(ref, () => ({
+      measureInWindow: () => {},
+      measure: () => {},
+      measureLayout: () => {},
+      setNativeProps: () => {},
```
szydlovsky, line 44: Maybe we could throw some `console.warns` stating that these are no-ops on Web in case someone used it?

src/web/EnrichedText.tsx
```diff
@@ -7,23 +13,47 @@ import {
...
+      focus: () => {
+        containerRef.current?.focus();
+      },
+      blur: () => {
+        containerRef.current?.blur();
+      },
```
szydlovsky, line 50: Maybe a stupid question, but it seems like the native `EnrichedText` doesn't have neither `focus` nor `blur` events to it - do we even need them here?

szydlovsky (commented):
In general nice, but I have a few uncertainties

src/web/EnrichedText.tsx
hejsztynx: If I see correctly, the native `EnrichedText` defines `blur` and `focus` methods on its `ref`

src/web/EnrichedText.tsx
hejsztynx: Not sure about that, no-op behavior is described in the docs and there are no `console.warn`'s in the `EnrichedInput` `ref` definition, but maybe it's worth looking back at and discussing

src/web/styleConversion/enrichedThemingToCSSProperties.ts
hejsztynx: To me it looks fine and consistent with other parts of this file - if something is not provided and `undefined`, we do nothing and allow it to fall back to its default value, similarly as `caretColor`

src/web/EnrichedText.tsx
hejsztynx: It makes a `div` programmatically focusable. Currently it's done the same way in the web's `EnrichedTextinput`

kacperzolkiewski (approved):

src/web/EnrichedText.tsx
szydlovsky: Right, looked in the wrong place. Still feels weird for a readonly component but let's keep it

src/web/EnrichedText.tsx
szydlovsky: Fine, no need to dig more

szydlovsky (approved):

PR #678, opened 2026-07-06
feat(web): onLinkPress and onMentionPress functionality

hejsztynx:
# Summary

Implemented the `onLinkPress` and `onMentionPress` callback functionality.

## Test Plan

- write a mention or a link using the input in the example app
- press on it
- a relevant callback should be visible in the logs

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ❌     |
| Web     |    ✅     |

## Checklist

- [x] E2E tests are passing


apps/example-web/src/testScreens/TestEnrichedText.tsx
```diff
@@ -18,6 +27,12 @@ export function TestEnrichedText() {
         <EnrichedText
           style={enrichedTextStyle}
           htmlStyle={WEB_DEFAULT_HTML_STYLE}
+          onLinkPress={(event) => {
+            setLastLinkPress(event);
+          }}
+          onMentionPress={(event) => {
+            setLastMentionPress(event);
+          }}
```
exploIF:
```suggestion
          onLinkPress={setLastLinkPress}
          onMentionPress={setLastMentionPress}
```

exploIF (approved):

kacperzolkiewski (approved):
LGTM!

szydlovsky (commented):
Overall looks solid but I noticed one strange thing; the press colors are prone to transitioning which looks pretty weird IMO. We could add a `transition: none;` to link's and mention's css  rules which should disable it for them.

here's a video where I'm going over a press frame-by-frame:

https://github.com/user-attachments/assets/1e822962-2c2b-4cb9-9680-09d7ba8d97ef

src/web/usePressInteractions.ts
```diff
@@ -0,0 +1,62 @@
...
+        const url = anchor.getAttribute('href');
+        if (url && onLinkPressRef.current) {
+          onLinkPressRef.current({ url });
+        }
+      }
+
+      const mention = target.closest('mention');
+      if (mention && container.contains(mention)) {
+        if (onMentionPressRef.current) {
+          e.preventDefault();
```
szydlovsky: Shouldn't we also `preventDefault` regardless whether `onMentionPressRef.current` is defined or not?
Links do exactly that - they first `preventDefault` and only then check the press ref and fire proper method.
As mentions are treated pretty much the same way, I'd expect them to work the same in this manner too.

src/web/styleConversion/buildMentionRulesCSS.ts
```diff
@@ -50,9 +54,27 @@ export function buildMentionRulesCSS(
...
+  background-color: var(${ET_MENTION_CSS_VARS.backgroundColor(indicator)}, var(${ET_MENTION_CSS_VARS.backgroundColor(MENTION_STYLE_DEFAULT_KEY)}));
+  text-decoration-line: var(${ET_MENTION_CSS_VARS.textDecorationLine(indicator)}, var(${ET_MENTION_CSS_VARS.textDecorationLine(MENTION_STYLE_DEFAULT_KEY)}));
+}`.trim()
+    );
+  }
+
+  // Press-state styling - only the read-only EnrichedText handles presses.
+  lines.push(
+    `${mentionSelector(ENRICHED_TEXT_CLASSNAME, MENTION_STYLE_DEFAULT_KEY)} {
+  cursor: pointer;
```
szydlovsky, line 71: We are now making mention look interactable at all times - even if the `onMentionPress` callback is not defined. I think we could make the `cursor` property use a variable that depends on the `onMentionPress` prop, like:

```TSX
'--et-mention-cursor': onMentionPress ? 'pointer' : 'default',
```

src/web/styleConversion/htmlStyleToCSSVariables.ts
```diff
@@ -261,17 +263,94 @@ function expandMentionStylesForIndicatorsIncludeDefault(
...
+    ET_LINK_PRESS_COLOR_VAR,
+    anchor?.pressColor ?? DEFAULT_ENRICHED_TEXT_STYLE.a.pressColor
+  );
+}
+
+function expandVarsWithEnrichedTextMention(
+  vars: Record<string, string>,
+  mention?: EnrichedTextHtmlStyle['mention']
+): void {
+  const mentionIndicators = isMentionStyleRecord(mention)
```
szydlovsky: `isMentionStyleRecord(mention)` is stable here and called multiple times in this method. Let's extract it to a separate variable.

szydlovsky (commented):
A few things I noticed, but overall nice work

src/web/usePressInteractions.ts
hejsztynx: I looked back and realized that `preventDefault` is not doing anything anyway, as `mention` is a custom HTML tag. Removed that completely

szydlovsky (approved):

PR #680, opened 2026-07-06
feat(web): HTML sanitization

hejsztynx:
# Summary

- implemented sanitization on all ends of the `EnrichedTextInput` and `EnrichedText`
- we were allowing all custom mention attributes names, inluding potentially malicious ones. Now the attributes are sanitized and if one doesn't start with the `data-` prefix, we print a runtime warning that it might get stripped by a sanitizer
- the `'default'` mention style functionality wasn't documented
- the normalizer let through only a predefined set of `mention` attributes: `id`, `text` and `indicator` - now it allows every custom attribute
- implemented a new web-only prop `sanitizationConfig`, currently with only `linkRegex` field, allowing to persist custom links which would be otherwise stripped by the sanitizer 

## Test Plan

Try to create a mention with an illegal HTML attribute, eg. `onClick`. It should be stripped and a suitable console warning should be visible.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ❌     |
| Web     |    ✅     |

## Checklist

- [x] E2E tests are passing


src/web/normalization/tiptapHtmlNormalizer.ts
```diff
@@ -8,6 +9,7 @@ export function prepareHtmlForTiptap(
   html: string,
   useHtmlNormalizer: boolean | undefined
 ): string {
+  html = sanitizeHtml(html);
```
kacperzolkiewski: Right now, when setting HTML in the input using the "Set Input's Value" button, some mention attributes (such as `data-type`, `type`) are stripped. Shouldn't those attributes be preserved, just like they are when using `setMention`?

When I insert: 
`<html><p><mention text="#engineering" indicator="#" id="c2" data-type="channel" type="channel">#engineering</mention></p></html>`

I receive in `onChangeHtml`:
`<html><p><mention text=\"#engineering\" indicator=\"#\" id=\"c2\">#engineering</mention></p></html>`

hejsztynx: Thanks, I missed that. That issue had been present for some time now and it also came up when you tried to copy-paste a mention. The normalizer only let through predefined set of attributes: `id`, `text` and `indicator`. Now it allows all of them and leaves the sanitization to the part that's responsible for that.

kacperzolkiewski (approved):

src/web/__tests__/sanitization.test.ts
```diff
@@ -0,0 +1,113 @@
...
+  it('strips javascript: URLs from links', () => {
+    const out = sanitizeHtml('<a href="javascript:alert(1)">x</a>');
+    // eslint-disable-next-line no-script-url
+    expect(out).not.toContain('javascript:');
+  });
+
+  it('strips unknown protocol URLs from links', () => {
+    const out = sanitizeHtml('<a href="custom://link">x</a>');
+    expect(out).not.toContain('custom');
+  });
```
exploIF, line 112: This is pretty important limitation, as such values are totally valid urls in mobile environment (including web browser on mobile devices). We should consider adding such support somehow

exploIF (changes requested):
In general looks good. However custom domains support is a blocker from my perspective. Happy to discuss potential solutions, but we should at least give an option for the users

apps/example-web/src/components/TextRenderer.tsx
```diff
@@ -44,6 +47,9 @@ export function TextRenderer({ htmlValue }: TextRendererProps) {
         onBlur={handleTextBlur}
         onLinkPress={handleLinkPress}
         onMentionPress={handleMentionPress}
+        sanitizationConfig={{
+          linkRegex: LINK_REGEX,
+        }}
```
exploIF: It's a good practice to move it out of the rendered component. That way we can avoid some overhead coming from the prop parsing on every render

apps/example-web/src/App.tsx
```diff
@@ -278,6 +278,9 @@ function App() {
           mentionIndicators={['@', '#']}
           htmlStyle={WEB_DEFAULT_HTML_STYLE}
           linkRegex={LINK_REGEX}
+          sanitizationConfig={{
+            linkRegex: LINK_REGEX,
+          }}
```
exploIF: Same here

exploIF (approved):

szydlovsky (approved):
:shipit:

PR #682, opened 2026-07-06
chore: upgrade example app and library deps to RN 0.86

kacperzolkiewski:
# Summary

- Upgrades example app to RN 0.86
- Upgrade project deps to RN 0.86

## Test Plan

- Ensure example app builds and works fine
- Ensure example-web app builds and works fine
- Ensure library builds properly

## Screenshots / Videos

n/a

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist

- [x] E2E tests are passing


src/native/EnrichedTextInput.tsx
```diff
@@ -39,7 +38,7 @@ const warnMentionIndicators = (indicator: string) => {
   );
 };
 
-type ComponentType = (Component<NativeProps, {}, any> & NativeMethods) | null;
+type ComponentType = React.ElementRef<HostComponent<NativeProps>>;
```
exploIF: In other places we import directly from `react` instead of using `React` default, for consistency we can change it here as well
```suggestion
type ComponentType = ElementRef<HostComponent<NativeProps>>;
```

src/native/EnrichedText.tsx
```diff
@@ -22,7 +21,7 @@ import { normalizeEnrichedTextHtmlStyle } from '../utils/normalizeHtmlStyle';
 import type { EnrichedTextProps } from '../types';
 import type { DirectEventHandler } from 'react-native/Libraries/Types/CodegenTypes';
 
-type ComponentType = (Component<NativeProps, {}, any> & NativeMethods) | null;
+type ComponentType = React.ElementRef<HostComponent<NativeProps>>;
```
exploIF: Same as in https://github.com/software-mansion/react-native-enriched-html/pull/682/changes#r3538445387

package.json
```diff
@@ -132,10 +133,10 @@
...
     "playwright": "1.58.2",
     "prettier": "^3.0.3",
-    "react": "19.1.0",
-    "react-dom": "19.1.0",
-    "react-native": "0.81.0",
-    "react-native-builder-bob": "^0.40.18",
+    "react": "19.2.3",
+    "react-dom": "19.2.3",
+    "react-native": "0.86.0",
+    "react-native-builder-bob": "^0.41.0",
```
exploIF: Most recent version is [0.43.0](https://github.com/callstack/react-native-builder-bob/releases/tag/react-native-builder-bob%400.43.0), is there any particular reason why it's not upgraded to the newest version?

kacperzolkiewski: Actually, I was looking at Reanimated's package.json and noticed that it uses React Native 0.86.0 with `react-native-builder-bob` version 0.41.0. I'll try 0.43.0 and see if it works the same.

kacperzolkiewski: I changed the version to 0.43.0 and it builds properly so I will leave it as is.

src/native/EnrichedTextInput.tsx
kacperzolkiewski: I'm using a direct import from React now. Also, I changed ElementRef into ComponentRef because ElementRef is deprecated.

exploIF (approved):
Works fine in the example app 🙌 

Just to confirm, have you run the library in clean react-native apps, with older RN versions?

szydlovsky (approved):

kacperzolkiewski:
> Works fine in the example app 🙌
> 
> Just to confirm, have you run the library in clean react-native apps, with older RN versions?

@exploIF Yes, I tested it with older versions on clean React Native apps. ;)

PR #686, opened 2026-07-08
test: add --shards option to mobile e2e tests

kacperzolkiewski:
# Summary

This PR adds a `--shards` flag to the Maestro test runner. The default value is `1`.
To speed up test execution, you can split the test suite across multiple emulators. For example:
`yarn test:e2e:android --shards 3`
`yarn test:e2e:ios --shards 3`
This will distribute the Maestro tests across three emulators, reducing the overall execution time.

When I run tests for example on iOS with 3 shards it reduces the time from 14min to 5min.

## Test Plan

1. Run commands:

`yarn test:e2e:android --shards N`
`yarn test:e2e:ios --shards N`
With different N value

2. Run Commands:

`yarn test:e2e:android`
`yarn test:e2e:ios`
Tests should run on single emulator like before this PR

## Screenshots / Videos

https://github.com/user-attachments/assets/b0c98b42-0fea-48dc-8d41-7ea862676e34

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| Web     |    ❌     |

## Checklist

- [x] E2E tests are passing


.maestro/scripts/run-tests.sh
```diff
@@ -38,21 +41,29 @@ REPO_ROOT="$(cd "$SCRIPT_DIR/../.." && pwd)"
 MAESTRO_ROOT="$REPO_ROOT/.maestro"
 SCREENSHOT_ROOT="$MAESTRO_ROOT"
 BUNDLE_ID="swmansion.enriched.example"
+EXAMPLE_DIR="$REPO_ROOT/apps/example"
 
 PLATFORM=""
 UPDATE_SCREENSHOTS=""
 REBUILD=""
+SHARDS=1
```
szydlovsky, line 49: Don't we want to then use some larger shards number as a default? I think the main motivation behind this PR is to, after all, speed up all the test runs. Or am I missing something and it works only for manual runs, when CI ones need to still go with `SHARDS=1`?

kacperzolkiewski: The intention was for this to have no effect by default. Developers only need to use the `--shards` option when they want to run the tests with sharding.

exploIF: I agree with @kacperzolkiewski, let's merge basic config first and adjust it later

PR #687, opened 2026-07-08
feat(web): ellipsizeMode and numberOfLines in EnrichedText

hejsztynx:
# Summary

Implemented `ellipsizeMode` and `numberOfLines` props in `EnrichedText` web component.

The RN `ellipsizeMode` can often cause confusion, as it only works when the ellipsized text is a continuous text that wraps the lines. If try to ellipsize text that consists of paragraphs, the `ellipsizeMode` bails, often just showing the given `numberOfLines` limit without actually rendering the ellipsis, which is pretty confusing. The web's DOM allows us to compute the text truncation more precisely and here it does so.

- `tail`, `head`, `middle`, `clip` are implemented
-  `middle` works a bit differently, as we can't really merge two different paragraph styles into one line, so if such situation occurs, we fallback to the `head` behavior
-  Generally, each mode creates an invisible sandbox, which is a copy of computed component, and finds the ellipsized HTML that matches given constraints in `useLayoutEffect`
- Multiple visual E2E tests were added, with various edge-case handling

## Test Plan

I've modified example app's code so it's easier to test repeatedly - the changes are marked with a `//` comment, and will be removed before the merge:

- there is a second `EnrichedText`, but without `numberOfLines` set to see how would it render without the ellipsizing,
- "Push text" does not clear the input, so re-push changes quicker

Play around with the modes, make sure the HTML is computed and rendered as expected. Inline images, line brakes, long words were often something that would brake the algorithm, so try working with that.

## Screenshots / Videos

image

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ❌     |
| Web     |    ✅     |

## Checklist

- [x] E2E tests are passing


src/web/EnrichedText.tsx
```diff
@@ -74,7 +90,7 @@ export const EnrichedText = memo(
           ref={containerRef}
           style={finalStyle}
           className={ENRICHED_TEXT_CLASSNAME}
-          dangerouslySetInnerHTML={{ __html: finalHtml }}
+          dangerouslySetInnerHTML={{ __html: clampedHtml ?? '' }}
```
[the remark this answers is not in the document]
hejsztynx: added a docs note about that

.playwright/tests/ellipsizeVisual.spec.ts
```diff
@@ -0,0 +1,313 @@
...
+
+  for (const c of middleEllipsizeCases) {
+    test(c.name, async ({ page }) => {
+      await gotoTestEllipsize(page);
+      await setValue(page, c.html);
+      await setNumberOfLines(page, c.numberOfLines);
+      await setMode(page, 'middle');
+
+      await expect(displayLocator(page)).toHaveScreenshot(
+        `ellipsize-middle-${c.slug}.png`
```
kacperzolkiewski: I would rename it to:

```suggestion
        `ellipsize-${c.slug}-middle.png`
```
Then it will be consistent with the rest of ellipsis modes

apps/example-web/src/components/TextRenderer.tsx
```diff
@@ -44,25 +49,24 @@ export function TextRenderer({ htmlValue }: TextRendererProps) {
...
+      >
+        {htmlValue}
+      </EnrichedText>
+      <EnrichedTextActions
+        ellipsizeMode={ellipsizeMode ?? 'tail'}
+        numberOfLines={numberOfLines}
+        onChangeEllipsizeMode={setEllipsizeMode}
+        onChangeNumberOfLines={setNumberOfLines}
+      />
+      {/*temporary second component to make testing easier*/}
```
kacperzolkiewski: Reminder: Remove it before merging

apps/example-web/src/testScreens/TestEllipsize.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+type EllipsizeMode = NonNullable<EnrichedTextProps['ellipsizeMode']>;
+
+const ELLIPSIZE_MODES: EllipsizeMode[] = ['head', 'middle', 'tail', 'clip'];
+
+export function TestEllipsize() {
+  const [htmlInput, setHtmlInput] = useState('');
+  const [value, setValue] = useState('');
+  const [numberOfLines, setNumberOfLines] = useState(2);
+  const [ellipsizeMode, setEllipsizeMode] = useState<EllipsizeMode>('tail');
```
kacperzolkiewski, line 18: Do you also have an eslint issue here?

Image

apps/example-web/src/App.tsx
```diff
@@ -234,7 +234,8 @@ function App() {
       ?.getHTML()
       .then((html) => {
         setEnrichedTextValue(html);
-        ref.current?.setValue('');
+        // temporary for making the testing easier
+        // ref.current?.setValue('');
```
kacperzolkiewski: Reminder: uncomment it before merging

src/web/EnrichedText.tsx
```diff
@@ -95,6 +99,14 @@ export const EnrichedText = memo(
       [textStyle, themingStyle, cssVars]
     );
 
+    useEllipsizeMode({
```
kacperzolkiewski, line 113: I think that we should react on some layout changes, because right now if someone for example will change `fontSize` for `EnrichedText` the content of it will be wrong:

https://github.com/user-attachments/assets/520b3e55-b314-4213-babe-c7e58542528a

src/web/ellipsizeMode/middleEllipsize.ts
```diff
@@ -0,0 +1,213 @@
...
+
+      if (past) toRm.push(rmWalker.currentNode);
+    }
+
+    toRm.forEach((n) => removeAndCleanUp(n, sandbox));
+
+    targetNode = backRightSplit;
+  }
+
+  const ellipsisNode = document.createTextNode('...');
```
kacperzolkiewski: There is dedicated ellipsis character: `\u2026`
Maybe we could use it?

apps/example-web/src/testScreens/TestEllipsize.tsx
hejsztynx: It's fine for me 🤔

src/web/ellipsizeMode/middleEllipsize.ts
```diff
@@ -0,0 +1,214 @@
...
+// If it fits, we keep it as it is. Otherwise:
+//   - A true inline middle only works when the front (line N) and the tail (last
+//     line) live in the same block with no <br> between them. If not (different
+//     blocks, or a <br> in the way), we bail and delegate to headEllipsize, which
+//     handles block-structured truncation.
+//   - Otherwise we split the front line at the centre mark, and - when more than
+//     one line overflows - also split at the second-to-last line and safely delete
+//     every node between the two split points. We insert "..." at the join, then eat
+//     forward one unit at a time until the surviving tail fits back onto line N.
+export function middleEllipsize(
```
kacperzolkiewski, line 29: `middle` does not work as expected on Safari for this html:
`<html><p>asdasd asd as dasdmaklsdmklasmdklma klsdmkl asmkld amskl dmaklsd mkalsm dklasmd klasm kldmask ldm plasm kldamskl damask dmklas mlkdamskl dmklasmk ldmalsk dmlkasmlkd amlskd mlkasd sa dadas dasd asd asd asd as das dasdasda sd asd asd asd </p></html>`

https://github.com/user-attachments/assets/e88e9210-7abe-4289-b606-e0b063922e35

Another example:
`<html><blockquote><p>asdnasjkdn ajksnd kansjkd najksd njkasd asd asd asdmaskld maskld maskl dmaskl dmaskl dmaskl mdklasmdklamskldmaslkdmlkasmdkl amsdlk masld maklsm dlkasm dlkasml kdmaskl dmalksm dlkasmdl alskdm lkasmdlk amslkd maklsmdl amsldk malksmklasdl asmdlk</p></blockquote></html>`

https://github.com/user-attachments/assets/bd25e0da-afc2-457b-996d-6d50a71dc89d

src/web/ellipsizeMode/utils.ts
```diff
@@ -0,0 +1,533 @@
...
+  while (isOverflowing) {
+    const lastNode = lastRenderedNode(sandbox);
+    if (!lastNode) break;
+
+    // handling text nodes: trim one character at a time (re-appending "..." each iteration)
+    if (lastNode.nodeType === Node.TEXT_NODE) {
+      const textNode = lastNode as Text;
+      let text = textNode.nodeValue || '';
+      if (withEllipsis && text.endsWith(ELLIPSIS_CHAR))
+        text = text.slice(0, -3);
```
szydlovsky: Isn't ellipsis char now only a single character rather than three?

src/web/EnrichedText.tsx
```diff
@@ -115,7 +129,7 @@ export const EnrichedText = memo(
           onBlur={(event) =>
             onBlur?.(adaptWebToNativeEvent(event, { target: -1 }))
           }
-          dangerouslySetInnerHTML={{ __html: finalHtml }}
+          dangerouslySetInnerHTML={{ __html: clampedHtml ?? '' }}
```
szydlovsky, line 143: `clampedHtml` starts as `null` and only gets set inside `useEllipsizeMode`'s `useLayoutEffect` - even when no `numberOfLines` or `ellipsizeMode` is set. So in this case there might be a first render layout flash because we initially render `''` and wait for `useLayoutEffect` even though the target html is already available for a while.

I think we could do that:

```suggestion
          dangerouslySetInnerHTML={{ __html: clampedHtml ?? finalHtml }}
```

src/web/ellipsizeMode/utils.ts
```diff
@@ -0,0 +1,533 @@
...
+  return last;
+}
+
+// Repeatedly removes the trailing node in the sandbox - trimming one character
+// at a time from text, or dropping a whole img / br / empty block - until the
+// remaining content no longer overflows past `lastBottom` (the bottom of the
+// last kept line). When `withEllipsis` is set, an "..." is (re)anchored onto the
+// trailing content as it shrinks. Otherwise content is simply clipped.
+// Shared by the tail and clip modes.
+export function eatBackwardUntilFits(
```
szydlovsky, line 359: In theory this node could be really really long and then we're essentially measuring char-by-char a lot of times. I wonder if we could employ a simple binary search here over this node's content? I think the difference would be really noticeable.

src/web/ellipsizeMode/utils.ts
```diff
@@ -0,0 +1,533 @@
...
+      }
+    }
+  }
+}
+
+// Repeatedly removes content immediately after the ellipsis - a single
+// character, or a whole img / br / empty block - until `checkFits` reports the
+// trailing content fits on the ellipsis line.
+// Shared by the head and middle modes.
+export function eatForwardUntilFits(
```
szydlovsky: Same as with `eatBackwardUntilFits`, we could use a binary search to ommit many unneeded measuring steps.

src/web/ellipsizeMode/utils.ts
hejsztynx: Thanks for that catch 🥇

src/web/ellipsizeMode/middleEllipsize.ts
hejsztynx: That should be fixed now, Safari's special

src/web/EnrichedText.tsx
hejsztynx: I don't think ther would be a flash. If there was `useEffect` used, then yes, but we have a synchronous `useLayoutEffect`.

I looked back at it at decided to tweak it a little anyway. Now when no `ellipsize` is needed, the `clampedHtml` is already set to `finalHtml`, otherwise the `div` was filled with empty content and then refilled with `finalHtml` after the `useLayoutEffect`. Now there should be no re-mounts in that case.

kacperzolkiewski (approved):
Nice work 🚀

src/web/ellipsizeMode/utils.ts
```diff
@@ -375,7 +375,8 @@ export function eatBackwardUntilFits(
     const lastNode = lastRenderedNode(sandbox);
     if (!lastNode) break;
 
-    // handling text nodes: trim one character at a time (re-appending "..." each iteration)
+    // handling text nodes: keep the longest prefix that still fits (re-appending
+    // "..." in ellipsis mode). With find the maximum amount of content to keep with binary-search.
```
szydlovsky: `With find the maximum amount of content to keep with binary-search.`
this sentence doesn't make sense to me 😛

szydlovsky (approved):
Binsearch looks good!

PR #703, opened 2026-07-16
fix(android): enriched text initial height measure

### LINKED ISSUE #700: [Android] EnrichedText adds large phantom trailing space when external HTML contains long attributes (e.g. a long href)
## Version
`react-native-enriched-html@1.0.0`, Android, New Architecture.

## Summary
When `EnrichedText` renders **external** HTML (a string not wrapped in `<html>…</html>`) with `useHtmlNormalizer`, the Android view is measured much taller than it renders. The extra height shows up as a large empty gap after the content. The gap grows with the amount of markup — a single anchor with a long `href` is enough to add ~15–20 phantom lines.

iOS renders the identical content with the correct height.

## Root cause
The **render** path normalizes/parses the HTML, but the **measure** path does not.

`EnrichedTextView.parseText` (render) handles non-`<html>` strings via `normalizeHtmlIfNeeded` → `GumboNormalizer.normalizeHtml` → `EnrichedParser.fromHtml`, so tags/attributes are stripped and only the visible text remains.

`MeasurementStore.getInitialText` (measure) only parses when the string is already wrapped in `<html>…</html>`:

```kotlin
val text = props?.getString("text") ?: ""
val isHtml = text.startsWith("<html>") && text.endsWith("</html>")
if (!isHtml) return text   // returns RAW markup, unparsed
```

Since `EnrichedText.tsx` passes `text={children}` verbatim (no wrapper), external HTML falls into the `!isHtml` branch and the raw markup — including the full `href` attribute value — is laid out by `StaticLayout` as literal text. Measured height ≫ rendered height ⇒ trailing gap.

## Reproduction
```tsx
<EnrichedText useHtmlNormalizer>
  {'Hello,\n\n<a href="https://example.com/path?a=1&b=2&c=…(700+ chars)…">Link</a>'}
</EnrichedText>
```
- **Android:** renders "Hello," + "Link", followed by a large empty gap whose size tracks the `href` length.
- **iOS:** renders correctly with no gap.

Minimal trigger: any external HTML with substantial markup relative to visible text; a long `href` makes it obvious.

## Expected
Android measurement should match the rendered content height (as iOS does).

## Suggested fix
Have `MeasurementStore.getInitialText` normalize/parse external HTML the same way the render path does before measuring — i.e. when the string is not internal `<html>` and `useHtmlNormalizer` is set, run it through `GumboNormalizer` + `EnrichedParser.fromHtml` (mirroring `EnrichedTextView.parseNormalizedHtml`) rather than returning the raw string.

## Workaround (for other users hitting this)
Wrap external HTML in `<html>…</html>` on Android only, which forces both the measure and render paths onto the internal-HTML parse branch:
```ts
const content = Platform.OS === 'android' ? `<html>${html}</html>` : html;
```
Verified to remove the gap while rendering identically.

kacperzolkiewski:
# Summary
Fixes: #700 

## Test Plan

Run reproduction steps from issue: #700
EnrichedText height should be measured properly.

## Screenshots / Videos

https://github.com/user-attachments/assets/5cf47a83-320b-4542-9714-bc8ab1054c8b

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |
| Web     |    ❌     |

## Checklist

- [x] E2E tests are passing

### DIFF
--- a/android/src/main/java/com/swmansion/enriched/text/MeasurementStore.kt
+++ b/android/src/main/java/com/swmansion/enriched/text/MeasurementStore.kt
@@ -17,6 +17,7 @@ import com.facebook.react.views.text.ReactTypefaceUtils.parseFontWeight
 import com.facebook.yoga.YogaMeasureMode
 import com.facebook.yoga.YogaMeasureOutput
 import com.swmansion.enriched.common.EnrichedConstants
+import com.swmansion.enriched.common.GumboNormalizer
 import com.swmansion.enriched.common.allowFontScalingFromProps
 import com.swmansion.enriched.common.parser.EnrichedParser
 import com.swmansion.enriched.common.pixelFromSpOrDp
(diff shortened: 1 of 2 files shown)


android/src/main/java/com/swmansion/enriched/text/MeasurementStore.kt
```diff
@@ -104,21 +105,77 @@ object MeasurementStore {
...
+  ): CharSequence? {
+    if (isInternalHtml) {
+      try {
+        val factory = EnrichedTextSpanFactory()
+        val parsed = EnrichedParser.fromHtml(text, style, factory)
+        return parsed.trimEnd('\n')
+      } catch (e: Exception) {
+        Log.w("MeasurementStore", "Error parsing initial HTML text: ${e.message}")
+        return normalizeHtmlIfNeeded(text, style, useHtmlNormalizer)
+      }
```
hejsztynx: Is it possible to inverse the condition? So we have something like this? I believe it would get cleaner, especially by removing duplicated  `EnrichedParser.fromHtml` logic and currently we have `normalizeHtmlIfNeeded` which does not imply that parsing is done inside of it.


```suggestion
    val toParse = isInternalHtml ? text : normalizeHtml(text, style, useHtmlNormalizer)
    
    try {
        val factory = EnrichedTextSpanFactory()
        val parsed = EnrichedParser.fromHtml(toParse, style, factory)
        return parsed.trimEnd('\n')
    } catch (e: Exception) {
        Log.w("MeasurementStore", "Error parsing initial HTML text: ${e.message}")
        return toParse
   }
```

hejsztynx (approved):
Overall, great! Just a small suggestion

android/src/main/java/com/swmansion/enriched/text/MeasurementStore.kt
kacperzolkiewski: Fixed

hejsztynx (approved):

exploIF (changes requested):
In general, it looks good. However, I'm not sure about diff size, I have a feeling that most of this changes, like separate `parseText` or `getEnrichedStyle` methods are completely redundant. I don't think they are improving code readability, rather making code review more complicated

exploIF (approved):

PR #704, opened 2026-07-16
fix: whitespaces parsing inside normalizer

kacperzolkiewski:
# Summary

This PR fixes whitespaces parsing between tags in normaliser.

## Test Plan

Insert some external html into example mobile/web app by `Set input's value` button for example:
```
<p>Asdasd</p>
<p>Asdasd</p>
<p>Asdasd</p>
```
And see if the output html is the same.

## Screenshots / Videos

https://github.com/user-attachments/assets/b112bfb8-0cc5-4fa7-b106-8b9978f70fa1

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| Web     |    ✅     |

## Checklist

- [X] E2E tests are passing


src/web/normalization/htmlNormalizer.ts
```diff
@@ -323,11 +323,27 @@ function escapeText(s: string): string {
 
 // --- Blockquote content flattening ---
 
+function isWhitespaceOnly(value: string): boolean {
+  for (let i = 0; i < value.length; i++) {
+    const c = value.charCodeAt(i);
+    // space, tab, LF, CR, FF — mirrors GumboNormalizer.c
```
hejsztynx: NIT: i would remove the mention about the mirroring. The whole file is mirrored overall and it's already stated in the very top of the file.
```suggestion
    // space, tab, LF, CR, FF
```

cpp/parser/GumboNormalizer.c
```diff
@@ -674,10 +689,13 @@ static void walk_children(GumboNode *node, buffer_t *out) {
...
         if (is_br_node(child)) {
-          if (ib.len > 0)
+          /* Whitespace-only buffer is layout noise; treat like empty → <br> */
+          if (ib.len > 0 && !is_whitespace_only(ib.data, ib.len))
             flush_inline_p(&ib, out);
-          else
+          else {
+            buffer_clear(&ib);
             buffer_append_str(out, "<br>");
+          }
```
hejsztynx: If we modified the `flush_inline_p` to return boolean, depending if it run the internal `if` branch, we could clean this up as the `ib.len` `is_whitespace_only` checks and `buffer_clean` is already done inside `flush_inline_p`. The same suggestion would apply to the web's normalzier. I don't insist on that, just a suggestion

```suggestion
if (!flush_inline_p(&ib, out))
     buffer_append_str(out, "<br>");
```

hejsztynx (approved):
Works great! Just left some minor comments.

cpp/parser/GumboNormalizer.c
kacperzolkiewski: fixed

exploIF (approved):

PR #707, opened 2026-07-17
feat(web): browser environment assertion

hejsztynx:
# Summary

Both components `EnrichedText` and `EnrichedTextInput` are using DOM - `DOMPurify`, `DOMParser`, `TipTap`. An attempt of rendering them in a non-browser environment, e.g. in SSR applications, would result in a crash. Only `DOMParser` availability check was present, but it was silent. Now we check for the browser environment before any other component logic and throw a proper error. Documentation was also updated, to explicitly point the client-only nature of the components.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ❌     |
| Web     |    ✅     |

## Checklist

- [x] E2E tests are passing


exploIF:
`TipTap` works SSR, why our input cannot work there as well?

hejsztynx:
> `TipTap` works SSR, why our input cannot work there as well?

We've decided to move the SSR implementation to the backlog, I'm re-opening this PR for reviews.

kacperzolkiewski (approved):
LGTM!

szydlovsky (approved):

PR #712, opened 2026-07-20
docs: project configuration

hejsztynx:
# Summary

- wired up ESLint and TS, as we now have interactive examples' code visible directly in the docs.
- `react-native-enriched-html` is now linked locally.
- as the library is resolved locally, the CI build scripts need its dependencies installed,so the root `node_modules` has become available now in the CI scripts.


eslint.config.mjs
```diff
@@ -33,6 +33,16 @@ export default defineConfig([
...
+    ignores: [
+      'node_modules/',
+      'lib/',
+      'apps/example-web/dist/',
+      'cpp/build/',
+      'docs/**/*',
+      '!docs/src',
+      '!docs/src/examples',
+      '!docs/src/examples/**',
+      '!docs/*',
```
kacperzolkiewski, line 48: What exactly do we want to include in the ESLint configuration from the docs at this point?

hejsztynx: Yeah, good point, it looks messy. I added a clarifying comment what is included in the end.

I could not get ESLint to work as needed until this `ignores` list worked. `'docs/**/*'` ignores all files and the last two`'!docs/src/examples/**'` with `'!docs/*'` includes what we want. The intermediate `'!docs/src'` and '!docs/src/examples'` are for the ESLint to be able to just traverse those directories (as the first line excludes them) one by one and reach `examples`, where we then re-include it.

If you find a better way of getting this to work, I'm happy to apply it.

kacperzolkiewski (approved):
LGTM

package.json
```diff
@@ -45,10 +45,11 @@
     "example": "yarn workspace react-native-enriched-example",
     "example-web": "yarn workspace react-native-enriched-web-example",
     "test": "jest --passWithNoTests",
-    "typecheck": "yarn typecheck:lib && yarn typecheck:example && yarn typecheck:example-web",
+    "typecheck": "yarn typecheck:lib && yarn typecheck:example && yarn typecheck:example-web && yarn typecheck:docs",
     "typecheck:lib": "tsc -p tsconfig.build.json",
     "typecheck:example": "tsc -p apps/example/tsconfig.json",
     "typecheck:example-web": "tsc -p apps/example-web/tsconfig.app.json",
+    "typecheck:docs": "tsc -p docs/tsconfig.json",
```
exploIF, line 54: Imo, we should run it before docs publish

package.json
exploIF, line 1: Could you add two scripts in root package.json:
- install docs dependencies
- start docs locally

hejsztynx: added `docs-install` and `docs-start` scripts

szydlovsky (approved):

exploIF (approved):

PR #713, opened 2026-07-20
docs: Rich text formatting section

hejsztynx:
# Summary

`Rich text formatting` section with:
- basic styles,
- links,
- mentions,
- lists,
- inline images,
- text alignment,
- text shortcuts

PR #716, opened 2026-07-21
docs: add api reference section

kacperzolkiewski:
# Summary

Add Api reference section to docs


docs/docs/api-reference/enriched-text-input.md
```diff
@@ -4,4 +4,1395 @@ sidebar_position: 1
 
 # EnrichedTextInput
 
-<!-- TODO: write content for this page -->
+`EnrichedTextInput` is a rich text editor that styles text live as you type.
+It is uncontrolled — content lives on the native side and you talk to it through
+a `ref`. See [Core concepts](/fundamentals/core-concepts) for the mental model.
```
hejsztynx: NIT: Maybe let's shorten this? I feel like api-reference section should be more straight-forward and that description can already be found in the `core-concepts`
```suggestion
`EnrichedTextInput` is a rich text editor that styles text live as you type.
See [Core concepts](/fundamentals/core-concepts) for the mental model.
```

docs/docs/api-reference/enriched-text-input.md
```diff
@@ -4,4 +4,1395 @@ sidebar_position: 1
...
+    selection: { start: number; end: number };
+    styleState: OnChangeStateEvent;
+  }) => void;
+  visible?: boolean;
+}
+```
+
+- `text` is the title displayed in the menu
+- `onPress` is the callback invoked when the item is tapped
+- `visible` controls whether the item is shown; defaults to `true`
```
hejsztynx: Maybe let's move the details of `onPress` after it is introduced
```suggestion
```ts
interface ContextMenuItem {
  text: string;
  onPress: (args: {
    text: string;
    selection: { start: number; end: number };
    styleState: OnChangeStateEvent;
  }) => void;
  visible?: boolean;
}
```

- `text` is the title displayed in the menu
- `onPress` is the callback invoked when the item is tapped
- `visible` controls whether the item is shown; defaults to `true`

The `onPress` callback receives a single object argument with:

- `text` — the currently selected text
- `selection` — an object with `start` and `end` indices of the current selection
- `styleState` — the latest `OnChangeStateEvent` payload reflecting active styles
  at the time of the tap
```

File: docs/docs/api-reference/enriched-text-input.md [hunk 3]
```diff
@@ -4,4 +4,1395 @@ sidebar_position: 1
...
+A custom regex pattern for detecting links in the input. If not provided, a
+default regex is used. You can customize which patterns are recognized as links
+— for example only `https://` URLs, or custom schemes.
+
+Not all JS regex features are supported; for example variable-width lookbehinds
+won't work.
+
+| Type             | Default                       | Platforms    |
+| ---------------- | ----------------------------- | ------------ |
+| `RegExp \| null` | default native platform regex | Android, iOS |
```
hejsztynx: I believe it is implemented on web

```suggestion
| `RegExp \| null` | default native platform regex | Android, iOS, Web |
```

docs/docs/api-reference/enriched-text-input.md
```diff
@@ -4,4 +4,1395 @@ sidebar_position: 1
...
+| ---------------------------------------------------------- | ------- | ----------------- |
+| `(event: NativeSyntheticEvent<OnChangeHtmlEvent>) => void` | —       | Android, iOS, Web |
+
+:::tip
+
+Specifying `onChangeHtml` may have performance implications, especially with
+large documents, as it requires continuous HTML parsing. If you only need the
+HTML at specific moments (for example when saving), use the
+[`getHTML`](#gethtml) ref method instead. When `onChangeHtml` is not provided,
+the component avoids unnecessary HTML parsing.
```
hejsztynx: It feels like it's already written in the first sentence

```suggestion
HTML at specific moments (for example when saving), use the
[`getHTML`](#gethtml) ref method instead.
```

docs/docs/api-reference/enriched-text-input.md
```diff
@@ -4,4 +4,1395 @@ sidebar_position: 1
...
+
+:::
+
+### `onChangeMention` <Optional /> {#onchangemention}
+
+Called whenever the user changes a mention that is being edited.
```
hejsztynx: Wording feels wrong, as it could imply the change of edited mentions, as switching between two different ones. Where `onChangeMention` fires when the query/text inside a single mention changes.

docs/docs/api-reference/enriched-text-input.md
```diff
@@ -4,4 +4,1395 @@ sidebar_position: 1
...
+Called whenever the input is focused.
+
+| Type                      | Default | Platforms         |
+| ------------------------- | ------- | ----------------- |
+| `(e: FocusEvent) => void` | —       | Android, iOS, Web |
+
+### `onLinkDetected` <Optional /> {#onlinkdetected}
+
+Called when a new link has been added, or the user has moved the
+cursor/selection onto a link.
```
hejsztynx: that's not 100% true. I believe it doesn't fire on a pure link addition, but rather from the fact that after it is added, the cursor sits in it. E.g. we paste some text which has a link, then no `onLinkDetected` is fired.

```suggestion
Called when the user has moved the
cursor/selection onto a link.
```

docs/docs/api-reference/enriched-text-input.md
```diff
@@ -4,4 +4,1395 @@ sidebar_position: 1
...
+- `end` is the first index after the ending index of the link
+
+| Type                              | Default | Platforms         |
+| --------------------------------- | ------- | ----------------- |
+| `(event: OnLinkDetected) => void` | —       | Android, iOS, Web |
+
+### `onMentionDetected` <Optional /> {#onmentiondetected}
+
+Called when a mention has been detected — either a new mention was added, or the
+user moved the cursor/selection onto a mention.
```
hejsztynx: The same case with `onLinkDetected`
```suggestion
Called when the
user moved the cursor/selection onto a mention.
```

docs/docs/api-reference/enriched-text-input.md
```diff
@@ -4,4 +4,1395 @@ sidebar_position: 1
...
+- `indicator` is the indicator of the mention
+- `attributes` are the additional user-defined attributes stored with the mention
+
+| Type                                 | Default | Platforms         |
+| ------------------------------------ | ------- | ----------------- |
+| `(event: OnMentionDetected) => void` | —       | Android, iOS, Web |
+
+### `onStartMention` <Optional /> {#onstartmention}
+
+Called whenever mention editing starts (after placing the indicator).
```
hejsztynx: It also fires when we re-enter an unfinished mention.

```suggestion
Called whenever mention editing starts.
```

docs/docs/api-reference/enriched-text-input.md
```diff
@@ -4,4 +4,1395 @@ sidebar_position: 1
...
+| `(event: NativeSyntheticEvent<OnSubmitEditing>) => void` | —       | Android, iOS, Web |
+
+### `onPasteImages` <Optional /> {#onpasteimages}
+
+Called when the user pastes one or more images or GIFs into the input.
+
+- `images` — an array of objects with URI, MIME type, and dimensions for each
+  pasted image/GIF
+- **Web:** each `uri` is a `blob:` URL (`URL.createObjectURL`). If you retain
+  URIs, call `URL.revokeObjectURL` when finished so blobs can be released
```
hejsztynx: that's already in the `:::note` below

and maybe we should use `:::caution` instead? But I'm not sure about that though
```suggestion
```

docs/docs/api-reference/enriched-text-input.md
```diff
@@ -4,4 +4,1395 @@ sidebar_position: 1
...
+fire when a closing delimiter is typed around text (e.g. `**text**` → bold). The
+trigger is the delimiter string (e.g. `**`, `*`, `~~`). Supported styles:
+`bold`, `italic`, `underline`, `strikethrough`, `inline_code`.
+
+:::note
+
+Style rules still apply to shortcut-triggered styles: if the target style is
+**blocked** by another currently active style (e.g. bold inside a codeblock),
+the shortcut has no effect. If the target style **conflicts** with another
+active style, the conflicting style is removed when the new one is applied. See
```
hejsztynx: Currently this might imply that e.g. `h1` shortcut in a `h2` paragraph, will conflict, and replace the `h2` into `h1`. Maybe it's just worth clarifying that the conflicting behavior only applies to inline style shortcuts. But honestly I'm not sure 🤔 

Also I wrote similar stuff in the #713 `Text shortcuts` section, I wonder where that should ultimately be present. What do you think?

docs/docs/api-reference/enriched-text-input.md
```diff
@@ -4,4 +4,1395 @@ sidebar_position: 1
...
+
+### `.removeLink()`
+
+```ts
+removeLink: (start: number, end: number) => void;
+```
+
+Removes link styling from any links found within the given range. The text
+content is preserved; only the link attributes are stripped. Out-of-bounds
+values are clamped to a valid range.
```
hejsztynx: NIT: I wonder if that's relevant and worth mentioning.

```suggestion
content is preserved; only the link attributes are stripped.
```

docs/docs/api-reference/enriched-text-input.md
```diff
@@ -4,4 +4,1395 @@ sidebar_position: 1
...
+
+Sets the currently edited mention with a given indicator, displayed text, and
+custom attributes.
+
+- `indicator: string` — indicator of the set mention
+- `text: string` — text displayed for the mention; anything the user typed is
+  replaced by that text. The mention indicator isn't added to that text
+- `attributes?: Record<string, string>` — additional custom attributes for the
+  mention, preserved through parsing to and from HTML
+
```
hejsztynx, line 869: It might be nice to add a note about using `data-` preifx

for example (I added such note to mantion-only-input guide):

```suggestion

:::note

The attributes you pass to `setMention` ride along in the HTML
and survive a round-trip through `getHTML` / `setValue`. Prefix custom keys with
`data-` if they need to outlive a sanitizer - see the note in
[Mentions](/rich-text-formatting/mentions).

:::

```

docs/docs/api-reference/enriched-text-input.md
```diff
@@ -4,4 +4,1395 @@ sidebar_position: 1
...
+- The input is uncontrolled. Change content and formatting through ref methods;
+  observe changes through events. See
+  [Core concepts](/fundamentals/core-concepts#the-input-is-uncontrolled).
+- Prefer [`getHTML`](#gethtml) over continuous `onChangeHtml` when you only need
+  HTML at specific moments.
+- Sanitizing HTML is your responsibility. See
+  [Core concepts](/fundamentals/core-concepts#html-is-the-source-of-truth).
+- For the full list of supported tags and style conflicts, see
+  [Supported tags](/fundamentals/html-format-and-supported-tags).
```
hejsztynx: Maybe this would be better?
```suggestion
  [HTML format and supported tags](/fundamentals/html-format-and-supported-tags).
```

docs/docs/api-reference/enriched-text-input.md
```diff
@@ -4,4 +4,1395 @@ sidebar_position: 1
...
+  useHtmlNormalizer?: boolean;
+  allowFontScaling?: boolean;
+}
+```
+
+</details>
+
+## Props
+
+### `allowFontScaling` <Optional /> {#allowfontscaling}
```
hejsztynx: I'm not sure about that `Optional` mark. Pretty much every prop we support is optional and IMO it creates unnecessary noise when reading, especially as its style it is pretty vibrant. What about changing the approach and marking the non-optional fields with a `<Required />` instead?

docs/docs/api-reference/enriched-text-input.md
```diff
@@ -4,4 +4,1395 @@ sidebar_position: 1
...
+
+### a (link)
+
+- `color` — color of link text; takes a
+  [color](https://reactnative.dev/docs/colors) value, defaults to `blue`
+- `textDecorationLine` — whether links are underlined; `'underline'` or
+  `'none'`, defaults to `'underline'`
+
+### mention
+
```
hejsztynx, line 1184: we should probably mention the `'default'` style, that user can create. I added this in `styling-the-input` section:

```
:::tip

You can also create a default `mention` style config, by using the `'default'` key.

htmlStyle={{
  mention: {
    'default': { color: '#2563eb', backgroundColor: '#dbeafe' },
    '#': { color: '#16a34a', backgroundColor: '#dcfce7' },
  },
}}

This way you can create a style for any mention indicator to fallback if it doesn't have one fully defined.

:::
```

the same would apply to the `EnrichedText` api ref

hejsztynx (changes requested):

docs/docs/api-reference/enriched-text-input.md
kacperzolkiewski: OK but few lines above this section we have:
```
:::note

Paragraph shortcuts are only effective on plain paragraphs. If the paragraph
already has an active paragraph style (for example it is already a heading or a
list item), typing the trigger pattern has no effect.

:::
```

hejsztynx: I know, but it might be worth clarifying, as I got confused by it myself, when I was reading that. If you see it as sufficient, I see your point, I don't insist ;)

hejsztynx (approved):

docs/docs/api-reference/enriched-text-input.md
kacperzolkiewski: Fixed

docs/docs/api-reference/enriched-text-input.md
kacperzolkiewski: Lets wait and see what @szydlovsky and @exploIF think about it ;)

szydlovsky: @hejsztynx @kacperzolkiewski I agree that most if not of all the props we have are optional. Do I see it right that the `EnrichedText`'s `children` is the only required prop? If so, I'm okay with using just a single `<Required />` and also emphasizing this fact in the docs.

docs/docs/api-reference/enriched-text-input.md
```diff
@@ -4,4 +4,1468 @@ sidebar_position: 1
...
+| ----------------------------------------------------------- | ------- | ----------------- |
+| `(event: NativeSyntheticEvent<OnPasteImagesEvent>) => void` | -       | Android, iOS, Web |
+
+:::note
+
+On Web, `uri` is a blob URL (`blob:...`). Blob URLs hold memory until explicitly
+released. Call `URL.revokeObjectURL(uri)` once you no longer need the image
+(for example after the upload completes).
+
+:::
```
szydlovsky, line 511: I'd make it a tip, as it rather suggests some good programming practice rather than explains some unusual or confusing API implications.

docs/docs/api-reference/enriched-text-input.md
```diff
@@ -4,4 +4,1468 @@ sidebar_position: 1
...
+
+| Type             | Default   | Platforms         |
+| ---------------- | --------- | ----------------- |
+| `TextShortcut[]` | see above | Android, iOS, Web |
+
+:::note
+
+Pass an empty array to disable all shortcuts.
+
+:::
```
szydlovsky: Also should be a `tip` IMO.

docs/docs/api-reference/enriched-text-input.md
```diff
@@ -4,4 +4,1468 @@ sidebar_position: 1
...
+  mention, preserved through parsing to and from HTML
+
+:::note
+
+The attributes you pass to `setMention` ride along in the HTML
+and survive a round-trip through `getHTML` / `setValue`. Prefix custom keys with
+`data-` if they need to outlive a sanitizer - see the note in
+[Mentions](/rich-text-formatting/mentions).
+
+:::
```
szydlovsky: Should be a `tip` or even a `caution` IMO - user that doesn't know this will suddenly lose their custom mention attributes.

szydlovsky (approved):

docs/docs/api-reference/enriched-text-input.md
kacperzolkiewski: Yes, you are right. I removed all the `Optional` badges and added a `Required` badge for `children` in `EnrichedText`.

docs/docs/api-reference/enriched-text-input.md
```diff
@@ -4,4 +4,1470 @@ sidebar_position: 1
...
+The `onPress` callback receives a single object argument with:
+
+- `text` - the currently selected text
+- `selection` - an object with `start` and `end` indices of the current selection
+- `styleState` - the latest `OnChangeStateEvent` payload reflecting active styles
+  at the time of the tap
+
+| Type                | Default | Platforms    |
+| ------------------- | ------- | ------------ |
+| `ContextMenuItem[]` | -       | Android, iOS |
```
exploIF, line 144: That's a general note, but let's remove `Default` column for callback, there is no option to have a default callback implemented

docs/docs/api-reference/enriched-text-input.md
```diff
@@ -4,4 +4,1470 @@ sidebar_position: 1
...
+
+### `onChangeState` {#onchangestate}
+
+Called when any of the styles within the selection changes. Use this to drive
+toolbar button state. See
+[The style state model](/fundamentals/core-concepts#the-style-state-model).
+
+```ts
+interface OnChangeStateEvent {
+  bold: { isActive: boolean; isConflicting: boolean; isBlocking: boolean };
```
exploIF: Can we make it more readable? I think it makes sense to extract `{ isActive: boolean; isConflicting: boolean; isBlocking: boolean }` to a separate interface and reuse it whenever possible. Currently it's hard to see difference between styles

docs/docs/api-reference/enriched-text-input.md
```diff
@@ -4,4 +4,1470 @@ sidebar_position: 1
...
+| Type                                                       | Default | Platforms         |
+| ---------------------------------------------------------- | ------- | ----------------- |
+| `(event: NativeSyntheticEvent<OnChangeTextEvent>) => void` | -       | Android, iOS, Web |
+
+:::tip
+
+If you don't need the plain text value, omit `onChangeText` - continuous text
+extraction can have performance implications.
+
+:::
```
exploIF: Let's remove it. It's true for html, but impact when using raw text should be minimal, not worth mentioning it imo

docs/docs/api-reference/enriched-text-input.md
```diff
@@ -4,4 +4,1470 @@ sidebar_position: 1
...
+    height: number;
+  }[];
+}
+```
+
+| Type                                                        | Default | Platforms         |
+| ----------------------------------------------------------- | ------- | ----------------- |
+| `(event: NativeSyntheticEvent<OnPasteImagesEvent>) => void` | -       | Android, iOS, Web |
+
+:::tip
```
exploIF: Let's make it `note` instead

docs/docs/api-reference/enriched-text-input.md
```diff
@@ -4,4 +4,1470 @@ sidebar_position: 1
...
+  observe changes through events. See
+  [Core concepts](/fundamentals/core-concepts#the-input-is-uncontrolled).
+- Prefer [`getHTML`](#gethtml) over continuous `onChangeHtml` when you only need
+  HTML at specific moments.
+- Sanitizing HTML is your responsibility. See
+  [Core concepts](/fundamentals/core-concepts#html-is-the-source-of-truth).
+- For the full list of supported tags and style conflicts, see
+  [HTML format and supported tags](/fundamentals/html-format-and-supported-tags).
+
+## Platform compatibility
```
exploIF: I think we can remove it

docs/docs/api-reference/enriched-text.md
```diff
@@ -4,4 +4,268 @@ sidebar_position: 2
...
+  onLinkPress?: (event: OnLinkPressEvent) => void;
+  onMentionPress?: (event: OnMentionPressEvent) => void;
+}
+```
+
+</details>
+
+## Props
+
+All props are optional except `children`, which is required.
```
exploIF:
```suggestion
All props are optional except `children`.
```

`except` already desires that this one is required ;)

docs/docs/api-reference/enriched-text.md
```diff
@@ -4,4 +4,268 @@ sidebar_position: 2
...
+
+:::
+
+### `numberOfLines` {#numberoflines}
+
+Limits the number of displayed lines. Set to `0` for unlimited lines.
+
+| Type     | Default | Platforms    |
+| -------- | ------- | ------------ |
+| `number` | `0`     | Android, iOS |
```
exploIF, line 159: Really default is `0`? It sounds weird to me, isn't it undefined?

docs/docs/api-reference/enriched-text.md
```diff
@@ -4,4 +4,268 @@ sidebar_position: 2
...
+Same as in `HtmlStyle`, if only a single config is given the style applies to
+all mention types. To style each indicator separately, pass a record with
+indicators as keys and configs as values. Use the `'default'` key for a fallback
+style applied to any indicator without its own config.
+
+## Remarks
+
+- `EnrichedText` is read-only. Use [`EnrichedTextInput`](/api-reference/enriched-text-input)
+  when the user needs to edit content.
+- Pass the HTML string as `children`, not as a `value` prop.
```
exploIF:
```suggestion
```

I think It's redundant

docs/docs/api-reference/enriched-text.md
```diff
@@ -4,4 +4,268 @@ sidebar_position: 2
...
+- Sanitizing HTML is your responsibility. See
+  [Core concepts](/fundamentals/core-concepts#html-is-the-source-of-truth).
+
+## Platform compatibility
+
+<PlatformCompatibility android ios web />
```
exploIF: I think it's not needed 
```suggestion
```

exploIF (commented):
🔥 🔥 🔥

docs/docs/api-reference/enriched-text.md
kacperzolkiewski: No, when you look at `EnrichedText` implementation the default value is `0`

docs/docs/api-reference/enriched-text-input.md
kacperzolkiewski: I extracted `{ isActive: boolean; isConflicting: boolean; isBlocking: boolean }` into `StyleState` interface

docs/docs/api-reference/enriched-text-input.md
kacperzolkiewski: Removed from all callbacks

docs/docs/api-reference/enriched-text-input.md
kacperzolkiewski: done

docs/docs/api-reference/enriched-text-input.md
kacperzolkiewski: done

exploIF (approved):

PR #719, opened 2026-07-22
docs: htmlStyle for interactive examples components

hejsztynx:
# Summary

Provided an `htmlStyle` for the `EnrichedTextInput` and `EnrichedText` components in the interactive examples. The internal default values didn't look too well, so this `htmlStyle` provides colorings that match the docs' style both in dark and light mode.

That makes the interactive examples' code not directly copiable, as `htmlStyle` is not present there to reduce noise. Added a note explaining that in the `your-first-editor` section.

### DIFF
--- a/docs/docs/fundamentals/your-first-editor.mdx
+++ b/docs/docs/fundamentals/your-first-editor.mdx
@@ -204,4 +204,13 @@ Here's that editor running live on this page. Select some text and hit **Bold** 
 notice the button turns green whenever the cursor sits on bold text. Switch to
 the **Code** tab to see the exact source.
 
+:::note
+
+The `htmlStyle` you'll spot in the Code tab is only there to make the rendered
+text match these docs' color palette, so
+don't worry about it here. If want to learn more, see
+[Styling the input](/core-functionalities/styling-the-input).
+
+:::
+
 <InteractiveExample src={FirstEditorSrc} component={FirstEditor} />
--- a/docs/docs/fundamentals/your-first-editor.mdx
+++ b/docs/docs/fundamentals/your-first-editor.mdx
@@ -204,13 +204,4 @@ Here's that editor running live on this page. Select some text and hit **Bold** 
 notice the button turns green whenever the cursor sits on bold text. Switch to
 the **Code** tab to see the exact source.
 
-:::note
-
-The `htmlStyle` you'll spot in the Code tab is only there to make the rendered
-text match these docs' color palette, so
-don't worry about it here. If you want to learn more, see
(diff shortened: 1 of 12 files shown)


docs/src/examples/FirstEditor.tsx
```diff
@@ -15,6 +16,7 @@ export default function App() {
       <EnrichedTextInput
         ref={ref}
         style={styles.input}
+        htmlStyle={htmlStyle}
```
kacperzolkiewski, line 23: I don't think we need to use `htmlStyle` in `FirstEditor` example, since we're only using the bold style there.

docs/docs/fundamentals/your-first-editor.mdx
```diff
@@ -198,4 +198,13 @@ Here's that editor running live on this page. Select some text and hit **Bold**
 notice the button turns green whenever the cursor sits on bold text. Switch to
 the **Code** tab to see the exact source.
 
+:::note
+
+The `htmlStyle` you'll spot in the Code tab is only there to make the rendered
```
kacperzolkiewski: This note is redundant here if we would remove `htmlStyle` from `FirstEditor` example?

docs/src/examples/htmlStyle.ts
```diff
@@ -0,0 +1,78 @@
...
+    backgroundColor: '#e2e5ff',
+    borderRadius: 8,
+  },
+  code: {
+    color: '#782aeb',
+    backgroundColor: '#e8dafc',
+  },
+  a: {
+    color: '#38acdd',
+    textDecorationLine: 'none',
```
kacperzolkiewski:
```suggestion
    textDecorationLine: 'underline',
```

?

docs/src/examples/FirstEditor.tsx
hejsztynx: I would keep it as it is, users can still type `-` and create a list, or just paste something which would look bad. I'd keep it consistent in the whole docs

kacperzolkiewski (approved):
LGTM

szydlovsky (approved):
lgtm

docs/docs/fundamentals/your-first-editor.mdx
```diff
@@ -204,4 +204,13 @@ Here's that editor running live on this page. Select some text and hit **Bold**
 notice the button turns green whenever the cursor sits on bold text. Switch to
 the **Code** tab to see the exact source.
 
+:::note
```
exploIF: This can be a comment in the code. There is no need to inform user about it if he is not going to open code tab at all

exploIF (approved):

PR #737, opened 2026-07-31
fix(android): preserve inline styles during IME composition

### LINKED ISSUE #734: [Android] Inline styles are stripped from previously-styled text when toggling a style OFF mid-word with predictive text enabled
## Description

On Android, when **predictive text / word suggestions are enabled** on the soft keyboard (so the current word is held in an active IME **composing region**), toggling an inline style (bold / italic / underline / strikethrough) **OFF** and then continuing to type **without a space** causes the **previously styled text in the same composing word to lose its style**.

The style formatting bit is dropped, not just the visual rendering — the earlier characters that were styled become unstyled.

## Steps to reproduce

Using the library's example app (raw `<EnrichedTextInput>`, no wrapper):

1. Use a soft keyboard with **predictive text / suggestions ON** (e.g. Gboard default).
2. Focus the editor and start typing a word, e.g. `abc`.
3. Toggle **bold ON**, continue typing **without a space** → `def` (now `def` is bold, still one composing word `abcdef`).
4. Toggle **bold OFF**, continue typing **without a space** → `ghi`.
5. Observe the result.

**Expected:** `def` stays bold; only `ghi` is unstyled → `abc` + **`def`** + `ghi`.

**Actual:** the previously bolded `def` loses its bold → the whole word renders unstyled.

Notes:
- Reproduces the same way for italic / underline / strikethrough.
- The trigger is the **active composing region** covering the styled text. If the IME commits each key immediately (`composing=[-1,-1]`), the bug does **not** occur. `hw.keyboard` turned out not to be the determining factor.
- Typing a **space** (which commits the composing region) before toggling avoids it.

## Environment

- `react-native-enriched-html`: reproduced on **latest `main`** (cloned repo example app) and on `1.1.0-nightly-20260724-982c0ca15`.
- React Native: 0.86
- Platform: Android only (iOS unaffected).
- Reproduced on:
  - Samsung Galaxy S9 — Android 10, **Samsung Keyboard**, English
  - Samsung Galaxy S20 Ultra — Android 13, **Samsung Keyboard**, English
  - Emulator: Pixel 8a, API 36 (Android 16), **Microsoft SwiftKey**, English

> **Correction (2026-08-03).** This section originally said "Gboard predictive text / suggestions ON" for all three environments. That was wrong: Gboard is not installed on the Galaxy devices, which use Samsung Keyboard. With **Gboard in English the bug does not reproduce at all** — Gboard sends every character via `commitText` with no composing region. It reproduces with any IME that keeps an active composing region, e.g. Microsoft SwiftKey on a plain emulator. See the analysis in the comments below.

## Likely cause (hypothesis)

While an IME composing region is active over the text, applying/removing an inline style span interacts with the composing region such that when the composing text is subsequently replaced (`setComposingText` → `SpannableStringBuilder.replace`), the character-level style spans within the replaced range are dropped.

Possible directions:
- Commit the composing region (`InputConnection.finishComposingText()`) before applying/removing an inline-style span, so the style is applied to committed (non-composing) text.
- Or re-apply / preserve existing character-level style spans across the composing-text replacement (span flags / re-span after `setComposingText`).

https://github.com/user-attachments/assets/0532fc28-f8b0-42d8-935d-033e1f647ca8

woosanggyu:
# Summary

Fixes #734.

On Android, predictive keyboards replace the entire active composing word on each `setComposingText` call. When an inline style was toggled off in the middle of that word, the replacement discarded enriched inline spans that belonged to the unchanged prefix.

This change:

- enables inline-style preservation only when a style is toggled during an active IME composition;
- snapshots enriched inline spans from the current composing range before `setComposingText` or `commitText`;
- restores spans only over unchanged prefix and suffix ranges, keeping newly typed text consistent with the current style state;
- clears preservation state when the composition is committed or finished; and
- queries all enriched inline spans in one pass to avoid overhead during normal typing.

The change is limited to the Android text input implementation.

## Test Plan

Manual verification in the example Android app with Gboard predictive text / suggestions enabled:

1. Type `abc`.
2. Toggle bold on and type `def` without a space.
3. Toggle bold off and type `ghi` without a space.
4. Verify that only `def` remains bold.
5. Repeat with italic, underline, and strikethrough.
6. Commit the word and verify that the earlier inline styles are still preserved.

Automated checks:

- Android Kotlin compilation passed.
- Android lint and formatting checks passed.
- `yarn lint` passed.
- `yarn typecheck` passed.
- `yarn test --maxWorkers=2` passed (415 tests).

## Screenshots / Videos

See the reproduction video in #734. The fix does not change layout or visual design.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |     ❌      |
| Android |     ✅      |
| Web     |     ❌      |

## Checklist

- [ ] E2E tests are passing (not run; the regression requires an active predictive-keyboard composing region)
- [x] Required E2E tests have been added (not applicable to the current automated input environment)


hejsztynx (changes requested):
Hi @woosanggyu !
First of all, thanks for taking your time to implement that fix.

Unfortunately, it doesn't work quite right - when following the initial test case you provided, it's fine, but when you move the cursor elsewhere and return to the end of that typed word, the inline styles are not preserved.

https://github.com/user-attachments/assets/7f851b3a-8d9a-40a4-a7bc-fa5f216f319a

woosanggyu:
Thanks for catching that — you're right, and the root cause was the gate I used.

The old version only armed preservation when an inline style was **toggled during
an active composition**. Moving the caret ends the composition and clears the
flag, so when you return to the end of the word the IME re-establishes a composing
region and replaces the whole word with no snapshot taken — the styled run inside
it is dropped.

The trigger isn't the toggle, it's **a composing replacement over text that
carries inline spans**, which happens with or without a toggle. So I moved the
gate from the event to the content: `setComposingText`/`commitText` now snapshot
whenever the composing region contains inline spans, and restore afterwards.
`captureComposingText()` returns null when there's nothing to preserve, so plain
typing still costs a single `getSpans` call over the composing range.

That removed the whole state machine along with the view-side hook, so
`EnrichedTextInputView.kt` is no longer touched by this PR.

Verified on an emulator with Microsoft SwiftKey (English):

1. `abc` → **B** → `def` → **B** → `ghi` — `def` stays bold (original case, no regression)
2. Same, then move the caret away, return to the end of the word, keep typing —
   `def` stays bold (your case)
3. Bold a word, unbold part of it via selection, then keep typing — the unbolded
   part stays unbolded (the snapshot is taken fresh before each replacement, so
   deliberate removals aren't resurrected)

android/src/main/java/com/swmansion/enriched/textinput/EnrichedTextInputConnectionWrapper.kt
```diff
@@ -61,6 +84,143 @@ class EnrichedTextInputConnectionWrapper(
...
+        val start = editable.getSpanStart(span).coerceAtLeast(composingStart)
+        val end = editable.getSpanEnd(span).coerceAtMost(composingEnd)
+
+        if (start < end) {
+          InlineSpanSnapshot(style, start - composingStart, end - composingStart)
+        } else {
+          null
+        }
+      }
+
```
hejsztynx: It would look cleaner and we would get rid of the "guessing" `style` lookup. What do you think?

```suggestion
    EnrichedSpans.inlineSpans.flatMap { (style, config) ->
      editable.getSpans(composingStart, composingEnd, config.clazz).mapNotNull { span ->
        val start = editable.getSpanStart(span).coerceAtLeast(composingStart)
        val end = editable.getSpanEnd(span).coerceAtMost(composingEnd)

        if (start < end) {
          InlineSpanSnapshot(style, start - composingStart, end - composingStart)
        } else {
          null
        }
      }
    }

```

android/src/main/java/com/swmansion/enriched/textinput/styles/InlineStyles.kt
```diff
@@ -161,6 +161,24 @@ class InlineStyles(
     setAndMergeSpans(spannable, type, start, end)
   }
 
+  fun restoreStyleOnRange(
```
hejsztynx: It strongly feels like duplicated logic, maybe instead of defining a new function, let's reuse `applyStyleOnRange` that we have above?

hejsztynx (changes requested):

android/src/main/java/com/swmansion/enriched/textinput/EnrichedTextInputConnectionWrapper.kt
woosanggyu: Good, I applied it. The isInstance lookup would also have picked the wrong style name if these span classes ever gained a hierarchy — this removes that.

android/src/main/java/com/swmansion/enriched/textinput/styles/InlineStyles.kt
woosanggyu: Agreed, dropped it and reused `applyStyleOnRange`. The `start >= end` guard was
redundant since the caller already checks it, so `InlineStyles.kt` is untouched
by this PR now.

hejsztynx (approved):
Looks good and works great. Thank you @woosanggyu for your work!

PR #730, opened 2026-07-28
fix(web): list styling

hejsztynx:
# Summary

Some of the web's list styling was working poorly

- the padding for the `unordered` and `checkbox` lists was calculated wrong. The `marginLeft`, `gapWidth` and `bulletSize` / `checkboxSize` were used improperly,
- the bullets and checkboxes should now be centered on the Y axis, to the text baseline, regardless for their sizes,
- `ol` marks are now aligned to the right of the `marginLeft` space (they sit within it), making the `gapWidth` always identical,
- `ol` marks get additional margin, if their values are too large to fit within the initial `marginLeft`

## Test Plan

Experiment with different values inside `defaultHtmlStyle.ts` inside `example-web/`

## Screenshots / Videos

The `ol` markers overflowing bug:

Before:

https://github.com/user-attachments/assets/1ca03de3-b732-42b6-8027-7a913cba531d

After:

https://github.com/user-attachments/assets/d7b04078-4223-4f37-b00d-9641e4c9be85

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ❌     |
| Web     |    ✅     |

## Checklist

- [x] E2E tests are passing
- [x] Required E2E tests have been added (if applicable)

PR #738, opened 2026-07-31
fix(iOS, Android): list styling

hejsztynx:
# Summary

There were some inconsistencies regarding the list styling between native platforms

- on iOS the `bulletSize` was not applied when computing the list indent
- implemented dynamic `ol` margins on both platforms. By default it uses the margin provided via `htmlStyle`, but if the rendered ordinal doesn't fit in it, it gets sufficiently increased. This solves an issue where markers would overlap the text. Equivalent fix was introduced in #730 
- the Android used to increase the `marginLeft`, to suppress the issue from the previous point - now this mechanism is removed, as no longer needed
- added relevant e2e tests, and regenerated the baselines, as list markers are now placed in a different manner

## Test Plan

experiment around with the htmlStyle in example app's `editorConfig.ts`

## Screenshots / Videos

Before:

image

After:

You can see the margin got extended, after introducing double-digit ordinals

image

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| Web     |    ❌     |

## Checklist

- [x] E2E tests are passing


ios/styles/OrderedListStyle.mm
```diff
@@ -58,10 +142,86 @@ - (void)applyStylingToTypingAttrs:(NSMutableDictionary *)attributes {
...
+// walks paragraphs backward and forward from the given range to find the full
+// contiguous run of ordered-list items it belongs to, and counts them
+- (NSRange)contiguousOrderedListRangeContaining:(NSRange)range
+                                      itemCount:(NSInteger *)outCount {
+  NSString *fullText = self.host.textView.textStorage.string;
+  NSUInteger length = fullText.length;
+  if (length == 0) {
+    if (outCount != nullptr) {
+      *outCount = 0;
+    }
```
kacperzolkiewski, line 179: Do we need this if block? 
We always call this function with `itemCount = 0`

hejsztynx: I would keep that as a safety net, as otherwise this function's stability would depend on how the function is called. We still can technically pass a pointer with uninitialized memory

android/src/main/java/com/swmansion/enriched/textinput/EnrichedTextInputView.kt
```diff
@@ -448,6 +449,9 @@ class EnrichedTextInputView :
 
     runAsATransaction {
       val newText = if (shouldParseHtml) parseText(value) else value
+      // Compute ordered-list marker column widths before setText so the first layout reserves
+      // the correct leading margin (getLeadingMargin runs during layout, before drawing).
+      (newText as? Spannable)?.let { updateOrderedListColumnMargins(it, paint) }
       setText(newText)
```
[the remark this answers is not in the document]
hejsztynx: this is fine. The `SpanWatcher` handles that, the margins are recalculated when spans are removed/added

docs/docs/api-reference/enriched-text-input.md
```diff
@@ -1236,6 +1236,14 @@ By marker, we mean the number that denotes consecutive lines of the list.
...
   the input's [color](#style)
 
+:::note
+
+As you may have noticed, there is no way to define an ordered list marker’s width (unlike other list styles).
+This is because the marker value changes, so its width is dynamic.
+The marker is rendered _inside_ the space defined by `marginLeft` (right-aligned within it), so keep this in mind when styling your UI.
+
+:::
+
```
exploIF, line 1252: This is expected and does not need any additional explanation

apps/example/src/constants/editorConfig.ts
```diff
@@ -111,7 +111,7 @@ export const htmlStyle = {
   },
   ol: {
     gapWidth: 16,
-    marginLeft: 24,
+    marginLeft: 16,
```
exploIF, line 114: What's the purpose of example app change?

exploIF (commented):
I'm afraid this PR will have many performance implications. Did you collected any metrics? Changes looks kind of risky given that issue they are trying to fix is not crucial

exploIF (changes requested):

android/src/main/java/com/swmansion/enriched/textinput/styles/ListStyles.kt
```diff
@@ -115,6 +116,8 @@ class ListStyles(
       val index = getOrderedListIndex(text, spanStart)
       span.setListIndex(index)
     }
+
+    updateOrderedListColumnMargins(text, view.paint)
```
exploIF, line 120: We are already going span by span in the `updateOrderedListIndexes`, then you iterate over them multiple times in the `updateOrderedListColumnMargins`. This could be definitely done in a more efficient way

android/src/main/java/com/swmansion/enriched/common/OrderedListMargins.kt
```diff
@@ -0,0 +1,73 @@
...
+  if (listSpans.isEmpty()) return
+
+  // Because the list was populated during a reversed() loop, the elements
+  // are in descending order. last() is the start of the list, first() is the end.
+  val start = text.getSpanStart(listSpans.last())
+  val end = text.getSpanEnd(listSpans.first())
+
+  if (start < 0 || end < 0 || start > end) return
+
+  class EmptySpan : ParagraphStyle
```
exploIF: We don't use such approach anywhere in the code. Instead, helper span class should be implemented, as we do in other places

exploIF (changes requested):

apps/example/src/constants/editorConfig.ts
hejsztynx: the margin got smaller, so we can easily see the dynamic margin when we just hit double digits, e.g. `10.`. It's easier to test the feature that way

docs/docs/api-reference/enriched-text-input.md
hejsztynx: I don't think it's that expected. E.g. in Google Docs it seems like the marker is rendered inside the equivalent of our `gapWidth`. One might assume a "margin" should contain effectively nothing. Anyway, I'm open to both options

android/src/main/java/com/swmansion/enriched/textinput/styles/ListStyles.kt
hejsztynx: How would you see it? iteration inside `updateOrderedListColumnMargins` is backwards, where we propagate the list's highest index, which was calculated in `updateOrderedListIndexes`. I guess we could add some kind of context to `EnrichedOrderedListSpan` objects, so they are aware of the its list other elements, but it feels like a difficult thing to maintain. Trying to reuse that `updateOrderedListIndexes` iteration would also couple those two functions together and make it harder to use `updateOrderedListColumnMargins` on its own, like we currently do in other places.

android/src/main/java/com/swmansion/enriched/common/OrderedListMargins.kt
hejsztynx: Maybe I've misunderstood, but there is this code in `EnrichedSpanWatcher`:

```kotlin
  // After adding/removing heading or inline MetricAffectingSpan spans, we have to manually set empty paragraph span to the following text
  // This allows us to update the layout (as it's not updated automatically - looks like an Android issue)
  private fun updateNextLineLayout(
    what: Any,
    text: Spannable,
    end: Int,
  ) {
    class EmptySpan : ParagraphStyle

    if (what is EnrichedHeadingSpan || (what is EnrichedInlineSpan && what is MetricAffectingSpan)) {
      val finalStart = (end + 1)
      val finalEnd = text.length
      val (safeStart, safeEnd) = text.getSafeSpanBoundaries(finalStart, finalEnd)
      text.setSpan(EmptySpan(), safeStart, safeEnd, Spannable.SPAN_EXCLUSIVE_EXCLUSIVE)
    }
  }
```

But I agree, it's better to define a helper class

docs/docs/api-reference/enriched-text-input.md
szydlovsky: I think it's worth leaving at least the 2nd part of the note that the marker is right-aligned within the margin. Having said that, I'm okay with leaving both.

ios/inputAttributesManager/InputAttributesManager.mm
```diff
@@ -66,16 +66,27 @@ - (void)clearRemovedTypingAttributes {
...
     NSRange dirtyRange = [rangeObj rangeValue];
 
+    // deletion (0-length dirtyRange) means we need to refresh the ordered
+    // list margins. If the ordered list style is not present itself,
+    // it might mean that we have just deleted a list element and need
+    // to recalculate the potentially split list (current adjacent lists)
+    if (dirtyRange.length == 0) {
+      if (orderedListStyle != nil) {
+        if ([orderedListStyle detect:dirtyRange]) {
+          [orderedListStyle applyStyling:dirtyRange];
```
szydlovsky: Do we really need to reapply the list styling if the list still exists in this dirty range? We could have anyway removed a list point but still landed on a different list.
My point is; isn't it safer to just always recalculate lists around the edited range? Regardless of existing lists in the dirty range.

ios/inputAttributesManager/InputAttributesManager.mm
```diff
@@ -90,6 +101,13 @@ - (void)handleDirtyRangesStyling {
       presentStyles[@([[style class] getType])] = [style all:dirtyRange];
     }
 
+    // it's possible that ordered list style has just got removed,
+    // so we have to refresh adjacent ordered lists
+    if (orderedListStyle != nil &&
+        [presentStyles[@([OrderedListStyle getType])] count] == 0) {
+      [orderedListStyle recalculateListsAroundEditedRange:dirtyRange];
+    }
+
```
szydlovsky: Holy moly we're checking for the lists here too? At this point, why don't we just go over all the dirty ranges (as the very first thing done in this `for` statement, before any `range==0` checks) and just always fix possible list around a dirty range. I think it's simpler in the long run and we're effectively doing the exact same work - just slam a `always check for lists needing recalculation` every time dirty ranges are handled.

ios/inputAttributesManager/InputAttributesManager.mm
szydlovsky, line 1: I see we're doing a lot of crazy hacks for the lists to be properly invalidated. I have an idea; why don't we just override `addDirtyRange` so that it will always check for surrounding lists; if the range is within a list, extend the range to the whole ordered list. Except for that, also scan for the ordered lists before/after the dirty range, and if such exists they should also go into the current dirty ranges.

After this, when handling the ordered lists within `handleDirtyRangesStyling`, we could be just be manually launching the proper method for recalculating indents (that you already added) when reapplying attributes (precisely, between `reapplyFromStylePair` and `applyStyling` calls) and I think we're essentially reaching the same fix, but being much closer to the current flow. The fixes made in this PR are just - in my opinion - stretching our implementation principles a bit too much.

ios/styles/OrderedListStyle.mm
szydlovsky, line 1: If we went for my suggested approach, we could still keep the `applyStyling` as a relatively simple method (as it should be). The idea is that when invalidating the margins (which happens when handling `dirtyRanges`), we also keep the calculated margin somewhere in the textLists value string (or as a separate attribute - both can be okay). This way `applyStyling` just reads the textLists/the new attribute in the range and applies proper margin.

If we do it, the proper flow is preserved:
- we first add just the meta attributes alone via `add` or `toggle`
- adding the meta attributes adds a properly larger dirty range so that the whole list/surrounding lists are always taken into account
- when we handle these dirty ranges, just after `reapplyFromStylePair` but before `applyStyling`, we recalculate the ordered lists' indents and keep this info somewhere (either `textLists` or some new attribute)
- then `applyStyling` comes and, using the margin data we provided, properly sets the indent

My main point is to keep the style file itself relatively simple and integral - I believe the current changes just make it hazy and we lose the flow somewhere along the way.

hejsztynx: Updated the flow, so it matches the current model. The margin data is now persisted in the attribute markers, which we then just read during `applyStyling`

PR #742, opened 2026-08-04
fix(web): img attributes sanitization

### LINKED ISSUE #741: [web] <img> tag attributes are ignored during HTML export
When exporting HTML, react-native-enriched-html ignores `` attributes such as the image link (src), width, and height. Only the bare `` tag gets exported.

Image

hejsztynx:
# Summary

- web's sanitization stripped `img`'s `height` and `weight` attributes when `sanitizationConfig.linkRegex` was defined, as those attributes were tested as URIs
- img placeholder now is displayed, even when no `src` is present in the `img` element
- added relevant regression tests
- updated the web example-app with a more practical `sanitizationConfig.linkRegex` #741 

## Test Plan

add some images in the input and push them to the `EnrichedText` - that will trigger sanitization. Everything should be at its place

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ❌     |
| Web     |    ✅     |

## Checklist

- [x] E2E tests are passing

### DIFF
--- a/apps/example-web/src/App.tsx
+++ b/apps/example-web/src/App.tsx
@@ -39,7 +39,7 @@ const LINK_REGEX =
   /^(?:enriched:\/\/\S+|(?:https?:\/\/)?(?:www\.)?swmansion\.com(?:\/\S*)?)$/i;
 
 const SANITIZATION_CONFIG = {
-  linkRegex: LINK_REGEX,
+  linkRegex: /^(?:enriched:\/\/\S+|https?:\/\/\S+)$/i,
 };
 
 function App() {
--- a/apps/example-web/src/App.tsx
+++ b/apps/example-web/src/App.tsx
@@ -39,7 +39,8 @@ const LINK_REGEX =
   /^(?:enriched:\/\/\S+|(?:https?:\/\/)?(?:www\.)?swmansion\.com(?:\/\S*)?)$/i;
 
 const SANITIZATION_CONFIG = {
-  linkRegex: /^(?:enriched:\/\/\S+|https?:\/\/\S+)$/i,
+  linkRegex:
+    /^(?:enriched:\/\/\S+|(?:https?:\/\/)?(?:www\.)?swmansion\.com(?:\/\S*)?|https?:\/\/\S+)$/i,
 };
 
 function App() {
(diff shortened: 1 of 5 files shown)


kacperzolkiewski:
Is this correct behaviour that href attribute is stripped from a tag?

https://github.com/user-attachments/assets/e60de648-408e-462b-b1ef-350cb316ed71

apps/example-web/src/components/TextRenderer.tsx
```diff
@@ -11,8 +11,7 @@ import {
 import { WEB_DEFAULT_HTML_STYLE } from '../defaultHtmlStyle';
 import { EnrichedTextActions } from './EnrichedTextActions';
 
-const LINK_REGEX =
-  /^(?:enriched:\/\/\S+|(?:https?:\/\/)?(?:www\.)?swmansion\.com(?:\/\S*)?)$/i;
+const LINK_REGEX = /^(?:enriched:\/\/\S+|https?:\/\/\S+)$/i;
```
exploIF: Why this change is needed?

apps/example-web/src/App.tsx
```diff
@@ -39,7 +39,7 @@ const LINK_REGEX =
   /^(?:enriched:\/\/\S+|(?:https?:\/\/)?(?:www\.)?swmansion\.com(?:\/\S*)?)$/i;
 
 const SANITIZATION_CONFIG = {
-  linkRegex: LINK_REGEX,
+  linkRegex: /^(?:enriched:\/\/\S+|https?:\/\/\S+)$/i,
```
exploIF: Same here

hejsztynx:
> Is this correct behaviour that href attribute is stripped from a tag?
> 
>  Screen.Recording.2026-08-04.at.13.15.42.mov

Good catch, now `sanitizationConfig.linkRegex` should be equivalent to the autolink detection `linkRegex` + any links with  `http` and `https` protocols, so `img`'s `src` are preserved

apps/example-web/src/components/TextRenderer.tsx
hejsztynx: the `sanitizationConfig.linkRegex` defines what URIs are preserved in the whole html, not only the `href` in the `<a>` tag, but also e.g. `src` in `img`, so we extend the sanitized URIs with `http` and `https` to keep those `img`s valid. The auto link detection is untouched, just the sanitization.

apps/example-web/src/App.tsx
hejsztynx: same as above

kacperzolkiewski (approved):

PR #743, opened 2026-08-05
fix: normalizer keeping empty src

hejsztynx:
# Summary

Now if no `src` attribute is present inside ``, the normalizer adds an empty one - `src=""`. This is done for the sake of structural consistency, which also fixed a couple of issues:

- on web, when you inserted an `` with no `src` attribute, the inline image would get dropped, whereas we want to still display the placeholder
- on iOS and Android, the parser would break if passed an empty `src` or no at all

## Test Plan

Set input's value with a `` tag and a `src=""`, `src="something" and no attribute at all. Each case should have the same effect - a placeholder inline image.

## Screenshots / Videos

iOS

Before:

https://github.com/user-attachments/assets/5fbd6bf4-1170-497e-a539-45fdc7509111

After:

https://github.com/user-attachments/assets/b794d6d9-e3f1-4622-88e3-d677c32812da

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| Web     |    ✅     |

## Checklist

- [x] E2E tests are passing

### DIFF
--- a/cpp/parser/GumboNormalizer.c
+++ b/cpp/parser/GumboNormalizer.c
@@ -491,7 +491,12 @@ static void emit_attributes(GumboElement *el, const char *tag_name,
   if (strcmp(tag_name, "a") == 0) {
     emit_one_attr(out, el, "href");
   } else if (strcmp(tag_name, "img") == 0) {
-    emit_one_attr(out, el, "src");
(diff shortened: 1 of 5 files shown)


kacperzolkiewski (approved):
LGTM

PR #745, opened 2026-08-05
fix(ios): apply obliqueness if italics are not supported

### LINKED ISSUE #739: [iOS] CJK text remains visually upright after toggleItalic despite <i> output
## Before submitting a new issue

- [x] Tested with the latest stable library version (`1.1.0`).
- [x] Tested with a supported React Native version (`0.86.2`).
- [x] Searched existing issues and found no duplicate for CJK/Chinese italic rendering.

## Bug summary

On iOS, `EnrichedTextInput.toggleItalic()` applies the expected `<i>` markup to selected Chinese, English, and numeric text, but Chinese glyphs remain visually upright. English letters and numbers in the same editor visibly render in italic.

The reproduction intentionally uses the system font and does not configure a custom font family or application-level faux-italic fallback. I am reporting this to clarify whether this is an expected platform-font limitation that should be documented, or whether the native renderer should provide consistent visual feedback for CJK text.

## Library version

`react-native-enriched-html@1.1.0`

## Environment info

```shell
System:
  OS: macOS 26.5.2
  CPU: (10) arm64 Apple M4
  Memory: 32.00 GB
Binaries:
  Node: 22.22.2
  npm: 11.18.0
  Watchman: 2026.07.20.00
Managers:
  CocoaPods: 1.17.0
SDKs:
  iOS SDK: iOS 26.5
IDEs:
  Xcode: 26.6/17F113
npmPackages:
  react: 19.2.3
  react-native: 0.86.2
  react-native-enriched-html: 1.1.0
Runtime:
  Expo SDK: 57.0.0
  New Architecture: enabled
Device:
  iPhone 17 simulator
  iOS 26.5
```

## Steps to reproduce

1. Clone the reproduction repository.
2. Run `npm install`.
3. Run `npm run ios`.
4. Tap **Italicize Chinese**, **Italicize English**, and **Italicize Numbers**.
5. Tap **Read generated HTML**.
6. Compare the visual rendering with the generated HTML.

## Expected behavior

All three selected ranges should have clear italic visual feedback, or the unsupported CJK italic behavior should be documented/exposed so apps can avoid presenting a control that appears ineffective.

## Actual behavior

The generated HTML correctly contains:

```html
<html>
<p><i>中文测试</i> <i>Hello</i> <i>123</i></p>
</html>
```

However, `Hello 123` visibly renders in italic while `中文测试` remains visually upright.

## Reproducible example repository

https://github.com/zhenyu-hong/react-native-enriched-html-cjk-italic-repro

## Screenshot

Chinese text remains upright while English and numbers render in italic

## Additional context

The reproduction uses only the library's public `setSelection()`, `toggleItalic()`, and `getHTML()` APIs. It has no custom toolbar, navigation, state management, or custom font configuration.

hejsztynx:
# Summary

this fixes #739 

- now `NSObliqueness` attribute is applied, when given font does not support proper italics
- added `stylingPriority` trait for base styles, so italics can be applied last

## Test Plan

Paste some non-latin characters, e.g. "冰淇淋", apply the italic style, everything should look as expected

## Screenshots / Videos

The italics are now properly applied

https://github.com/user-attachments/assets/ed73f153-98b6-4d6c-bb2b-7d17eed75c25

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |
| Web     |    ❌     |

## Checklist

- [x] E2E tests are passing


ios/interfaces/StyleBase.h
```diff
@@ -14,6 +14,7 @@
 - (BOOL)isParagraph;
 - (BOOL)needsZWS;
 - (BOOL)appliesStylingToTyping;
+- (NSInteger)stylingPriority;
```
kacperzolkiewski: There wili be conflict with `customStyle` I applied there `stylePriority` property in `StyleBase`: https://github.com/software-mansion/react-native-enriched-html/pull/641/changes#diff-7d290ee647218a0cde80f6b721d4476384026b9a060c8906cc4964c36f28ba0dR43

ios/inputAttributesManager/InputAttributesManager.mm
```diff
@@ -97,16 +105,21 @@ - (void)handleDirtyRangesStyling {
 
     // Sort style types so paragraph styles come first. Their broad visual
     // attributes (e.g. foreground color, font) are laid down before inline
-    // styles override them on their specific sub-ranges.
+    // styles override them on their specific sub-ranges. Inline styles among
```
kacperzolkiewski, line 102: And I also added sorting in here by `stylePriority` in `customStyle`.  And that would fix problem with inline style applied in wrong order because there inline styles are applied as last. https://github.com/software-mansion/react-native-enriched-html/pull/641/changes#diff-f0ba56d65d53cacb60b6d3b46e66fc464076119a4d5df36d92fcdd98b2a95847R98

ios/inputAttributesManager/InputAttributesManager.mm
```diff
@@ -97,16 +105,21 @@ - (void)handleDirtyRangesStyling {
...
-          BOOL aPara = [_input->stylesDict[a] isParagraph];
-          BOOL bPara = [_input->stylesDict[b] isParagraph];
-          if (aPara == bPara)
-            return NSOrderedSame;
-          return aPara ? NSOrderedAscending : NSOrderedDescending;
-        }];
+        sortedArrayWithOptions:NSSortStable
+               usingComparator:^NSComparisonResult(NSNumber *a, NSNumber *b) {
+                 NSInteger aOrder =
+                     [self stylingOrderFor:_input->stylesDict[a]];
```
kacperzolkiewski: I would get rid of `stylingOrderFor` function in here and simply use:
`NSInteger aOrder = [_input->stylesDict[a] stylePriority];
NSInteger bOrder = [_input->stylesDict[b] stylePriority];`
If there will be `nil` in here it would be treated as `0` so we do not need to add some extra check for that.

ios/interfaces/StyleBase.mm
```diff
@@ -45,6 +45,11 @@ - (BOOL)appliesStylingToTyping {
   return NO;
 }
 
+// determines the order in which the styles are applied
+- (NSInteger)stylePriority {
+  return 0;
```
kacperzolkiewski: And then here we would need to add check if the style is paragraph or not something like:
`return [self isParagraph] ? 0 : 2;`
I would do this in this way because those priorities are already implemented in `CustomStyle` and there will be less conflicts.

ios/textHtmlParser/TextHtmlParser.mm
```diff
@@ -153,8 +153,20 @@ - (void)applyProcessedStyles:(NSArray *_Nonnull)processedStyles {
...
+  NSArray *sortedInlineApply = [pendingInlineApply
+      sortedArrayWithOptions:NSSortStable
+             usingComparator:^NSComparisonResult(NSArray *a, NSArray *b) {
+               NSInteger aPriority = [((StyleBase *)a[0]) stylePriority];
+               NSInteger bPriority = [((StyleBase *)b[0]) stylePriority];
+               if (aPriority == bPriority)
+                 return NSOrderedSame;
+               return aPriority < bPriority ? NSOrderedAscending
+                                            : NSOrderedDescending;
+             }];
```
kacperzolkiewski: This comparator is almost the same as in `InputAttributesManager` maybe we could extract this logic?

ios/utils/ItalicUtils.mm
```diff
@@ -0,0 +1,179 @@
+#import "ItalicUtils.h"
```
kacperzolkiewski, line 1: Do we need to extract this whole logic into a separate file? It's only related to `ItalicStyle`, so I think it could stay there. All other styles for example`LinkStyle`, `MentionStyle` contains all of its logic in the same file without a separate utility.

ios/textHtmlParser/TextHtmlParser.mm
hejsztynx: done

ios/utils/ItalicUtils.mm
hejsztynx: good point, I agree

ios/styles/ItalicStyle.mm
```diff
@@ -1,6 +1,46 @@
...
+  static NSCharacterSet *nonNeutral = nil;
+  static dispatch_once_t onceToken;
+  dispatch_once(&onceToken, ^{
+    NSMutableCharacterSet *set =
+        [[NSCharacterSet whitespaceAndNewlineCharacterSet] mutableCopy];
+    [set formUnionWithCharacterSet:[NSCharacterSet controlCharacterSet]];
+    // ZWS
+    [set addCharactersInString:[NSString
+                                   stringWithFormat:@"%C", (unichar)0x200B]];
+    nonNeutral = [[set invertedSet] copy];
```
szydlovsky: We could add a comment that this set is all characters EXCEPT whitespaces, newlines, control characters, ZWS

ios/styles/ItalicStyle.mm
```diff
@@ -17,21 +57,147 @@ - (BOOL)isParagraph {
...
-              }];
+  [self applyItalicInTextStorage:self.host.textView.textStorage inRange:range];
+}
+
+// some styles might apply a new font (inline code), so we need to apply
+// the italic last, that way knowing if the used font supports italics
+// or we need to apply a slant
+- (NSInteger)stylePriority {
+  return 3;
+}
```
szydlovsky, line 24: NIT; let's move it before `applyStyling`.

ios/styles/ItalicStyle.mm
```diff
@@ -17,21 +57,147 @@ - (BOOL)isParagraph {
...
-                             value:newFont
-                             range:range];
-                }
-              }];
+  [self applyItalicInTextStorage:self.host.textView.textStorage inRange:range];
+}
+
+// some styles might apply a new font (inline code), so we need to apply
+// the italic last, that way knowing if the used font supports italics
+// or we need to apply a slant
```
szydlovsky, line 21:
```suggestion
// some styles might apply a new font (inline code), so we need to apply
// the italic last, that way we know if the used font supports italics
// or we need to apply a slant
```

ios/styles/ItalicStyle.mm
```diff
@@ -17,21 +57,147 @@ - (BOOL)isParagraph {
...
+- (NSInteger)stylePriority {
+  return 3;
+}
+
+- (void)applyItalicInTextStorage:(NSTextStorage *)textStorage
```
szydlovsky: As far as I see, we only use this in `applyStyling`. Can't we just move all of the code from this method? I don't like such redundant helpers.

ios/utils/StyleUtils.h
```diff
@@ -30,4 +30,7 @@
 + (void)removeStyleConflict:(StyleType)conflicting
                        from:(StyleType)conflicted
                     forHost:(id<EnrichedViewHost>)host;
+
++ (NSArray *)sortedArray:(NSArray *)array
+               bySortKey:(NSInteger (^)(id item))sortKeyForItem;
```
szydlovsky: Honestly it has nothing to do with `style` utils as such - let's move it to maybe some new `ArrayExtension` in `/extension` catalog.

ios/styles/ItalicStyle.mm
```diff
@@ -17,21 +57,147 @@ - (BOOL)isParagraph {
...
+                         }
+                         [self applyItalicForFont:font
+                                    inTextStorage:textStorage
+                                          inRange:fontRange];
+                       }];
+}
+
+- (void)applyItalicForFont:(UIFont *)font
+             inTextStorage:(NSTextStorage *)textStorage
+                   inRange:(NSRange)range {
```
szydlovsky: I think we could move that method and also all of the related helpers/interfaces to some new `ItalicUtils` in `/utils` catalog. They just litter this style file which should be rather comprehensible. Things I'd move too:
- `kObliquenessFallback`
- `ItalicKind`
- `NonNeutralCharacters`
- `kindForCluster`
- `applyKind`

and
`FontCoversCharacters` could go to `FontExtension` as e.g. `coversCharacters` method defined on a given font.

ios/textHtmlParser/TextHtmlParser.mm
```diff
@@ -153,8 +153,15 @@ - (void)applyProcessedStyles:(NSArray *_Nonnull)processedStyles {
     }
   }
 
+  // Respect the styling priority
+  NSArray *sortedInlineApply =
+      [StyleUtils sortedArray:pendingInlineApply
+                    bySortKey:^NSInteger(NSArray *entry) {
+                      return [((StyleBase *)entry[0]) stylePriority];
+                    }];
+
```
szydlovsky, line 162: Don't we need such styling priority tech for input parser too?

ios/utils/ItalicUtils.mm
szydlovsky: @kacperzolkiewski I beg to differ. Style files should be minimal and easily comprehensible for anyone taking a peek. Properly named utils and helpers tucked away somewhere actually make the picture clearer.

szydlovsky: As for `LinkStyle` and `MentionStyle`, there's just a lot of cases to check, nothing in there really clouds the view because the viewer needs to know the details. Here, however, the detail of applying obliqueness to some font isn't something we want to see from inside out - we just need to know that some method *applied* the style one way or another.

ios/textHtmlParser/TextHtmlParser.mm
hejsztynx: I beg to differ. Here in `TextHtmlParser` the styles are sorted, so we can apply them in that order. Input's parser doesn't apply the styles on its own, it only registers them and adds the dirty ranges. Eventually the `InputAttributesManager` actually applies them, so they are sorted there

ios/extensions/FontExtension.mm
```diff
@@ -40,8 +41,6 @@ - (UIFont *)setItalic {
   if (fontDescriptor != nullptr) {
     return [UIFont fontWithDescriptor:fontDescriptor size:0];
   } else {
-    RCTLogWarn(
-        @"[EnrichedTextInput]: Couldn't apply italic trait to the font.");
```
hejsztynx, line 44: I also noticed that this console warning is now quite possibly unwanted, as we apply a slant if italic isn't supported

szydlovsky: @hejsztynx yeah, good call.

ios/utils/ItalicUtils.h
```diff
@@ -0,0 +1,23 @@
...
+  // text attachments)
+  ItalicKindNone,
+  // font has a real italic glyph for the character
+  ItalicKindFont,
+  // no italic glyph available, the slant has to be used
+  ItalicKindOblique,
+};
+
+// slant used when a font has no italic face
+extern const CGFloat kObliquenessFallback;
```
szydlovsky: Can't it just stay as a local variable declared only in the `.mm` file? I'm pretty sure it will anyway be only used by the methods shared through `ItalicUtils`

szydlovsky (approved):
Looks very nice now!

PR #746, opened 2026-08-05
feat: add linkOnPaste prop to EnrichedTextInput (iOS, Android, Web)

jtatar:
# Summary

Adds an opt-in `linkOnPaste` prop to `<EnrichedTextInput />` (iOS, Android, Web).

When enabled, pasting clipboard content that is **solely a URL** over a
**non-empty selection** turns the selected text into a link pointing to that
URL, instead of replacing the selection with the pasted text. This is a common
rich-text convenience (it mirrors the "paste URL onto selected text" behavior
of editors like Google Docs, Notion, Slack).

**Disabled by default** (`linkOnPaste={false}`), so existing behavior is
unchanged for current users — the feature only activates when explicitly turned
on.

Behavior details:

- The pasted text is recognized as a URL only when it **fully** matches the
  configured `linkRegex` (or the default link-detection patterns when
  `linkRegex` is not set). Clipboard text containing a URL plus other text
  (e.g. `see https://example.com`) is **not** treated as a link and falls back
  to a normal paste.
- Scheme-less URLs (e.g. `www.example.com`) get an `https://` prefix in the
  resulting link href, while the visible selected text is preserved.
- Falls back to the normal paste when the selection is empty/whitespace-only,
  or when the link style cannot be applied at the selection (inside inline code
  or a code block).
- Has no effect when link detection is disabled with `linkRegex={null}`.

Implementation:

- **JS**: prop on the codegen spec, `types.ts`, default props (`false`), and the
  native/web `EnrichedTextInput` wrappers.
- **iOS**: gated branch in `EnrichedInputTextView`'s `paste:`; a new
  `tryAddLinkAt:...` (returns whether the link was applied) and
  `linkURLIfEntireString:` on `EnrichedTextInputView`; a
  `matchesEntireLinkRegexWithConfig:` full-string variant on `LinkStyle`.
- **Android**: gated branch in `handleTextPaste`; a `linkifySelectionOnPaste`
  helper plus a `linkExactRegex` (whole-string) companion to `linkRegex`, and a
  `setLinkOnPaste` ViewManager setter.
- **Web**: `handleLinkOnPaste` wired into the TipTap editor's `handlePaste`,
  reusing the existing autolink regex and the `EnrichedLink.setLink` command
  (which already bails out when the link style is blocked).

## Test Plan

**Web (automated):** added a `test-links linkOnPaste` suite to
`.playwright/tests/links.spec.ts` covering: linkifying the selection with a full
URL, `https://` prefixing for a scheme-less URL, and the two fallback cases
(non-bare-URL text, and no selection). Run with `yarn test:e2e:web`. The full
links suite (35 tests) passes.

**Manual (all platforms).** The example app enables `linkOnPaste` on its editor
screens. To verify:

1. Type `Hello world`, select `world`.
2. Copy a URL that the editor's `linkRegex` accepts and paste over the selection
   → `world` becomes a link to that URL (visible text unchanged).
3. Copy a scheme-less form (`www.…`) → the link href is prefixed with `https://`.
4. Copy text that is not a bare URL (e.g. `see https://…`) → normal paste
   (selection replaced, not linkified).
5. Set `linkOnPaste={false}` → step 2 becomes a normal paste again (selection
   replaced), confirming the default/off behavior is unchanged.

`yarn typecheck`, `yarn lint`, and `yarn lint-clang` all pass; the Android
Gradle `lintVerify` builds the module cleanly.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |     ✅      |
| Android |     ✅      |
| Web     |     ✅      |

## Checklist

- [x] E2E tests are passing

### DIFF
--- a/ios/EnrichedTextInputView.h
+++ b/ios/EnrichedTextInputView.h
@@ -36,11 +36,18 @@ NS_ASSUME_NONNULL_BEGIN
 @public
   BOOL useHtmlNormalizer;
 @public
+  BOOL linkOnPaste;
+@public
   NSValue *dotReplacementRange;
 @public
   NSArray<NSDictionary *> *textShortcuts;
 }
 - (CGSize)measureSize:(CGFloat)maxWidth;
+- (BOOL)tryAddLinkAt:(NSInteger)start
+                 end:(NSInteger)end
+                text:(NSString *)text
+                 url:(NSString *)url;
+- (nullable NSString *)linkURLIfEntireString:(NSString *)text;
 - (void)emitOnLinkDetectedEvent:(LinkData *)linkData range:(NSRange)range;
 - (void)emitOnMentionEvent:(NSString *)indicator text:(nullable NSString *)text;
 - (void)emitOnPasteImagesEvent:(NSArray<NSDictionary *> *)images;
--- a/ios/EnrichedTextInputView.mm
+++ b/ios/EnrichedTextInputView.mm
@@ -693,6 +693,11 @@ Class<RCTComponentViewProtocol> EnrichedTextInputViewCls(void) {
     useHtmlNormalizer = newViewProps.useHtmlNormalizer;
   }
 
+  // linkOnPaste
+  if (newViewProps.linkOnPaste != oldViewProps.linkOnPaste) {
+    linkOnPaste = newViewProps.linkOnPaste;
+  }
+
   // textShortcuts
   bool textShortcutsChanged =
       newViewProps.textShortcuts.size() != oldViewProps.textShortcuts.size();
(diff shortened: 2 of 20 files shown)


hejsztynx:
I've adjusted a couple things here:

- merged two helpers into one - `.playwright/helpers/clipboard.ts`
- removed the http-appending logic, we don't want that, we don't do anything like that in the code and it breaks custom link schemes, which we allow, e.g. `enriched://`
- removed some duplicated logic - e.g. `linkExactRegex` prop on Android, and two methods on iOS
- made the web implementation go through our custom `setLink` flow, just like `ref.setLink()` method
- stopped being strict about "regex matching the whole word" - it's more consistent how automatic link detection is already implemented and this changed allowed for a considerable code simplification

ios/EnrichedTextInputView.h
```diff
@@ -43,6 +45,11 @@ NS_ASSUME_NONNULL_BEGIN
   BOOL preserveTypingAttributesOnNextEmptyCheck;
 }
 - (CGSize)measureSize:(CGFloat)maxWidth;
+- (BOOL)addLinkAt:(NSInteger)start
+              end:(NSInteger)end
+             text:(NSString *)text
+              url:(NSString *)url;
+- (nullable NSString *)linkTextIfMatchesLinkRegex:(NSString *)text;
```
szydlovsky: NIT; but we could move it to the end of the declared methods list, no real reason to insert it there.

ios/EnrichedTextInputView.h
```diff
@@ -35,6 +35,8 @@ NS_ASSUME_NONNULL_BEGIN
   BOOL blockEmitting;
 @public
   BOOL useHtmlNormalizer;
+@public
+  BOOL linkOnPaste;
```
szydlovsky: Nit; this could also be placed as the last declared property

ios/EnrichedTextInputView.mm
```diff
@@ -1520,27 +1525,49 @@ - (void)toggleCheckboxList:(BOOL)checked {
   }
 }
 
-- (void)addLinkAt:(NSInteger)start
+- (BOOL)addLinkAt:(NSInteger)start
```
szydlovsky, line 1529: if we changed the method signature, let's at least add a short comment what the return value is - in this case something along `return value informs us whether the link has been properly added or not`

ios/EnrichedTextInputView.mm
```diff
@@ -1520,27 +1525,49 @@ - (void)toggleCheckboxList:(BOOL)checked {
...
+  LinkData *linkData = [[LinkData alloc] init];
+  linkData.text = text;
+  linkData.url = url;
+  linkData.isManual = YES;
+  [linkStyleClass addLink:linkData range:linkRange withSelection:YES];
+  [self anyTextMayHaveBeenModified];
+  return YES;
+}
+
+- (NSString *)linkTextIfMatchesLinkRegex:(NSString *)text {
```
szydlovsky: This method anyway returns the very text that we provided. I think it's cleaner if it just returned a `BOOL` whether the given text matches the regex or not.

src/spec/EnrichedTextInputNativeComponent.ts
```diff
@@ -369,6 +369,7 @@ export interface NativeProps extends ViewProps {
   htmlStyle?: HtmlStyleInternal;
   scrollEnabled?: boolean;
   linkRegex?: LinkNativeRegex;
+  linkOnPaste?: boolean;
```
szydlovsky: We discussed it together and decided to rename this prop to `applyLinkOnPaste`, let's change it everywhere.

ios/EnrichedTextInputView.mm
szydlovsky: We only need the `LinkStyle` and `Config` to determine whether some `url` matches the link regex. No need to have this method in the componentView whatsoever, we can move it back to the `EnrichedInputTextView` subclass. `LinkStyle` can be queried from the public `stylesDict` on the `input` and `config` is also a public prop there.

Also - when we move it - this method doesn't need to be returning the very same string that it's been given, it can be just a bool check.

hejsztynx: Nice refactor idea, previously it was a bit clunky, I agree. Did what you suggested aside from the `stylesDict` query, as that `LinkStyle` class method is static and already available

ios/enrichedInputTextView/EnrichedInputTextView.mm
```diff
@@ -226,6 +251,30 @@ - (void)paste:(id)sender {
   [typedInput anyTextMayHaveBeenModified];
 }
 
+- (BOOL)linkTextIfMatchesLinkRegex:(NSString *)text {
```
szydlovsky: Now the old name doesn't really fit here
```suggestion
- (BOOL)textMatchesLinkRegex:(NSString *)text {
```

szydlovsky (approved):

PR #754, opened 2026-08-10
feat(iOS): add fontSize and fontFamily to customStyle

kacperzolkiewski:
# Summary

Add fontSize and fontFamily to customStyle

## Test Plan
Run example app and experiment with applying `fontSize` and `fontFamily` through the toolbar across different text styles.

## Screenshots / Videos

https://github.com/user-attachments/assets/20adf200-2661-44bf-a931-1948e1419afa

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |
| Web     |    ❌     |

## Checklist

- [x] E2E tests are passing


hejsztynx (commented):
Overall looks great, but it seems like the custom style fonts does not respect system's accessibility settings, I think they should, wdyt?

https://github.com/user-attachments/assets/352497dc-9dcc-4141-905d-fdaa833b404b

ios/config/EnrichedConfig.mm
```diff
@@ -560,13 +560,15 @@ - (void)setAllowFontScaling:(BOOL)newValue {
   }
 }
 
-- (NSNumber *)scaledPrimaryFontSize {
+- (CGFloat)scaledFontSizeForValue:(CGFloat)fontSize {
```
hejsztynx, line 541: If we're introducing such function, let's use it also with `h[1-6]FontSize` and `lineHeight` functions here in this file

hejsztynx (changes requested):

ios/config/EnrichedConfig.mm
kacperzolkiewski: done

hejsztynx (approved):
🚀

ios/styles/CustomStyle.mm
```diff
@@ -24,6 +26,46 @@ - (NSInteger)stylePriority {
   return 1;
 }
 
+- (UIFont *)resolveFont:(UIFont *)font forData:(CustomStyleData *)data {
+  UIFont *resolvedFont = font ?: [self.host.config primaryFont];
+  UIFont *originalFont = resolvedFont;
+
+  if (data.fontFamily.length > 0) {
+    resolvedFont = [RCTFont updateFont:resolvedFont
+                            withFamily:data.fontFamily
```
szydlovsky, line 35: I think we should handle a case when `fontFamily` can't be found, the method docs say:
```
/**
 * Update a font with a given font-family, size, weight and style.
 * If parameters are not specified, they'll be kept as-is.
 * If font is nil, the default system font of size 14 will be used.
 */
```
So we could maybe guard the `nil` value and if it happens, just don't replace the font attribute to the system font of size 14 but rather keep whatever was before (+ emit some warning too). I think setting this random font will break our fonts flow.

ios/EnrichedTextInputView.mm
szydlovsky, line 1: Same as with foreground/background color, I think we should be returning a definite value (style state) only if the whole selection has the same one.

Image

Image

apps/example/src/components/Toolbar.tsx
szydlovsky, line 1: NIT: the font family text in toolbar isn't aligned the greatest for two-word strings (or generally long words).

Image

szydlovsky (changes requested):

PR #755, opened 2026-08-10
feat: maxLength

### LINKED ISSUE #732: [Feature] maxLength for EnrichedTextInput, including paste and dictation
**Is your feature request related to a problem? Please describe.**

`EnrichedTextInput` has no `maxLength`, so there is no way to cap how much text a user can enter. In a form that mixes plain `TextInput`s with enriched ones, the same character limit usually has to apply to every field, and the enriched ones are currently the only fields where it cannot be enforced.

Doing it from JS does not really work: the value is HTML, so the limit has to be measured on parsed plain text, and the only way to enforce it is to set the value back after the fact, which resets the selection and fights the undo stack. Paste is worse still, because the text is already in the editor by the time JS sees it.

**Describe the solution you'd like**

A `maxLength?: number` prop that counts the editor's plain text, ignoring formatting, and matches `TextInput` semantics: typing past the limit is rejected, longer input is truncated to what still fits.

The parts that are easy to get subtly wrong:

- Paste truncates the pasted fragment instead of being rejected, leaving whatever followed the caret intact.
- Dictation and IME input arrive as multi character replacements and need the same truncation, not the rejection used for typing.
- The cut point snaps outwards, so emoji, surrogate pairs and combining marks never get split.
- Truncated text still goes through the regular editing path, so formatting, link detection and mentions keep working, and `onChangeHtml` / `onChangeText` fire with what actually ended up in the editor.

**Describe alternatives you've considered**

Truncating in JS after every change loses the selection and cannot handle paste. Blocking submission instead of input leaves users pasting a wall of text into a field that will then refuse to submit.

**Additional context**

We currently carry this as a native patch on top of the library, on both Android and iOS.

hejsztynx:
# Summary

Closes #732.

Provided the implementation of the `maxLength` prop to the `EnrichedTextInput` component.

- The prop is optional and if assigned with a value, the `EnrichedTextInput`'s content will be limited to the given number of characters
- The prop is respected on every input's entry-point, such as `setValue`, `defaultValue`, typing, pasting content or programmatically via `ref.setImage()`, `ref.setImage()`, `ref.setLink()`
- Provided relevant e2e tests on all platforms

## Test Plan

There's an available interface on mobile on the `TestScreen` and under `/test-max-length` URL on web.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| Web     |    ✅     |

## Checklist

- [x] E2E tests are passing


android/src/main/java/com/swmansion/enriched/textinput/utils/MaxLengthFilter.kt
```diff
@@ -0,0 +1,104 @@
...
+  fun plainLengthOf(
+    text: CharSequence,
+    start: Int = 0,
+    end: Int = text.length,
+  ): Int {
+    var length = 0
+    for (i in start until end) {
+      if (text[i] != EnrichedConstants.ZWS) length++
+    }
+    return length
```
exploIF: Why not 
```kt
val length = text.count { it != EnrichedConstants.ZWS }
```

docs/docs/api-reference/enriched-text-input.md
```diff
@@ -232,6 +233,22 @@ For details, see [sanitization](/core-functionalities/web-support).
...
+
+| Type     | Default Value | Platform          |
+| -------- | ------------- | ----------------- |
+| `number` | no limit      | iOS, Android, Web |
+
+:::note
+
+Programmatic insertions respect the limit too - e.g. `setLink()`, `setMention()`, `setImage()`.
+
+:::
```
exploIF: Why it's not documented in the original comment together with `defaultValue` and `setValue`, but instead introduced as a new "notes" entry?

android/src/main/java/com/swmansion/enriched/textinput/utils/MaxLengthFilter.kt
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  /**
+   * Index in `[start, end]` at which [text] has to be cut so that at most [capacity] plain
+   * characters are kept. The cut point snaps outwards to a whole grapheme cluster, so emoji,
+   * surrogate pairs and combining marks never end up split in half.
```
exploIF: What does `grapheme cluster` or `surrogate pairs`  mean?

android/src/main/java/com/swmansion/enriched/textinput/styles/ParametrizedStyles.kt
```diff
@@ -343,18 +343,20 @@ class ParametrizedStyles(
     val spannable = view.text as SpannableStringBuilder
     val (start, originalEnd) = view.selection.getInlineSelection()
 
-    if (start == originalEnd) {
-      spannable.insert(start, EnrichedConstants.ORC_STRING)
```
exploIF, line 347: How this is handled after your changes?

android/src/main/java/com/swmansion/enriched/textinput/utils/EnrichedSpannable.kt
```diff
@@ -20,6 +20,26 @@ fun Spannable.getSafeSpanBoundaries(
   return Pair(safeStart, safeEnd)
 }
 
+/**
+ * Replaces `[start, end)` with [text] and returns how many characters actually landed in the
+ * buffer - the `maxLength` filter may have shortened, or entirely rejected, the replacement.
+ */
+fun SpannableStringBuilder.replaceCountingInserted(
```
exploIF: This name is not self descriptive

android/src/main/java/com/swmansion/enriched/textinput/utils/MaxLengthFilter.kt
```diff
@@ -0,0 +1,104 @@
...
+import com.swmansion.enriched.textinput.EnrichedTextInputView
+import java.text.BreakIterator
+
+/**
+ * Helpers enforcing the `maxLength` prop.
+ */
+object MaxLength {
+  const val UNLIMITED = -1
+
+  /** Length of [text] as seen by the user (zero width spaces excluded). */
```
exploIF: If this method is named differently there is no need to add additional comments. That's a general comment as I see this pattern in other places as well

android/src/main/java/com/swmansion/enriched/textinput/utils/MaxLengthFilter.kt
```diff
@@ -0,0 +1,104 @@
...
+    start: Int,
+    end: Int,
+    capacity: Int,
+  ): Int {
+    var index = start
+    var kept = 0
+
+    while (index < end) {
+      if (text[index] != EnrichedConstants.ZWS) {
+        // zero width spaces are free, any other character needs the capacity
```
exploIF: What does `free` mean in this context?

android/src/main/java/com/swmansion/enriched/textinput/styles/ParametrizedStyles.kt
```diff
@@ -393,12 +391,13 @@ class ParametrizedStyles(
...
     view.runAsATransaction {
-      spannable.replace(start, end, text)
+      val insertedLength = spannable.replaceCountingInserted(start, end, text)
+      val (safeStart, safeEnd) = spannable.getSafeSpanBoundaries(start, start + insertedLength)
 
-      val span = EnrichedInputMentionSpan(text, indicator, attributes, view.htmlStyle)
-      val spanEnd = start + text.length
-      val (safeStart, safeEnd) = spannable.getSafeSpanBoundaries(start, spanEnd)
-      spannable.setSpan(span, safeStart, safeEnd, EnrichedSpanFlags.forSpan(span))
+      if (insertedLength == text.length) {
```
exploIF, line 397: Before we haven't had that check right? What's the reason?

apps/example-web/src/testScreens/TestMaxLength.tsx
```diff
@@ -0,0 +1,254 @@
...
+  const [selEndInput, setSelEndInput] = useState('0');
+  const [linkStartInput, setLinkStartInput] = useState('0');
+  const [linkEndInput, setLinkEndInput] = useState('0');
+  const [linkTextInput, setLinkTextInput] = useState('link');
+  const [linkUrlInput, setLinkUrlInput] = useState('https://example.com');
+  const [mentionIndicatorInput, setMentionIndicatorInput] = useState('@');
+  const [mentionTextInput, setMentionTextInput] = useState('Jane');
+  const [imageSrcInput, setImageSrcInput] = useState('/pw-e2e-ok.png');
+  const [imageWidthInput, setImageWidthInput] = useState('40');
+  const [imageHeightInput, setImageHeightInput] = useState('40');
```
exploIF: Either simplify it, or use objects, eventually `useReducer`

exploIF (changes requested):

docs/docs/api-reference/enriched-text-input.md
hejsztynx: Removed the note and added `setLink` to that original comment to make the listed example fuller. I'm not sure if it's worth adding all of those three function names there

android/src/main/java/com/swmansion/enriched/textinput/styles/ParametrizedStyles.kt
hejsztynx: the general `spannable.insert` / `spannable.replace` mechanism is now handled in the `replaceAndCountInserted` function, run below

android/src/main/java/com/swmansion/enriched/textinput/utils/MaxLengthFilter.kt
hejsztynx: Cleaned up other functions as well

android/src/main/java/com/swmansion/enriched/textinput/utils/MaxLengthFilter.kt
hejsztynx: 'free', as they don't 'cost' any capacity we are progressively using. But I agree, worth rewriting

android/src/main/java/com/swmansion/enriched/textinput/styles/ParametrizedStyles.kt
hejsztynx: We don't want to apply the mention span if we didn't insert the full mention. E.g. we wanted `@JohnDoe`, but maxLength didn't let us insert it and we are left with `@Joh`. It's not a valid mention, we leave that text unstyled

android/src/main/java/com/swmansion/enriched/textinput/utils/MaxLengthFilter.kt
hejsztynx: Generally, basic emojis are stored as surrogate pairs, as they fill up double the memory space a single character normally uses. Grapheme clusters is a general term where multiple characters are rendered into one, e.g. emojis, CJK. Ultimately we don't want to split them in half

PR #758, opened 2026-08-11
feat(android): add fontSize and fontFamily to customStyle

kacperzolkiewski:
# Summary

Add fontSize and fontFamily to customStyle

## Test Plan
Run example app and experiment with applying `fontSize` and `fontFamily` through the toolbar across different text styles.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |
| Web     |    ❌     |

## Checklist

- [x] E2E tests are passing


android/src/main/java/com/swmansion/enriched/textinput/styles/CustomStyles.kt
```diff
@@ -18,19 +19,43 @@ class CustomStyles(
...
 
-      view.spanState?.setCustomStyle(finalFg, finalBg)
+    if (start == end) {
+      val merged = styleUpdater(view.spanState?.customStyle ?: CustomStyle())
+      view.spanState?.setCustomStyle(
+        merged.foregroundColor,
+        merged.backgroundColor,
+        merged.fontSize,
+        merged.fontFamily,
+      )
```
hejsztynx, line 56: maybe it's worth a minor refactor? Currently we're just unpacking and reinstantiating `CustomStyle` object
```suggestion
      val merged = styleUpdater(view.spanState?.customStyle ?: CustomStyle())
      view.spanState?.setCustomStyle(merged)
```

android/src/main/java/com/swmansion/enriched/textinput/EnrichedTextInputView.kt
```diff
@@ -148,7 +148,9 @@ class EnrichedTextInputView :
   private var defaultValueDirty: Boolean = false
 
   private var inputMethodManager: InputMethodManager? = null
-  private val spannableFactory = EnrichedTextInputSpannableFactory()
+
+  private fun spannableFactory() = EnrichedTextInputSpannableFactory(context.assets, allowFontScaling)
```
hejsztynx: now it looks like instance a new `EnrichedTextInputSpannableFactory` everytime we use it. `context.assets` should be constant (?) and we could update `allowFontScaling` in its setter. Or at least cache the factory.

The same change would apply to `EnrichedTextSpannableFactory` usage

hejsztynx (changes requested):

android/src/main/java/com/swmansion/enriched/textinput/styles/CustomStyles.kt
kacperzolkiewski: If I wanted to pass the merged `CustomStyle` object directly to `setCustomStyle`, I would need to refactor `setCustomStyle` to accept a single `CustomStyle` argument instead of f`gColor: Int?, bgColor: Int?, fontSize: Float? = null, fontFamily: String? = null`. Since I also use this function elsewhere, I would prefer not to make that change.

android/src/main/java/com/swmansion/enriched/textinput/EnrichedTextInputView.kt
kacperzolkiewski: Fixed, I created `cachedSpannableFactory` that it is invalidated on every `allowFontScaling` prop change

hejsztynx (approved):
LGTM

PR #759, opened 2026-08-12
feat(web): add fontSize and font Family to customStyle

kacperzolkiewski:
# Summary

Add fontSize and font Family to customStyle

## Test Plan

Run example web app and experiment with applying `fontSize` and `fontFamily` through the toolbar across different text styles.

## Screenshots / Videos

https://github.com/user-attachments/assets/29268496-e19d-4fb0-a8ea-2cc4e4c850dd

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ❌     |
| Web     |    ✅     |

## Checklist

- [X] E2E tests are passing


src/web/formats/EnrichedCustomStyle.ts
```diff
@@ -88,27 +151,89 @@ export const EnrichedCustomStyle = Mark.create({
...
+          }
+          if ('fontSize' in attrs) {
+            patch.fontSize = resolveFontSize(attrs.fontSize);
+          }
+          if ('fontFamily' in attrs) {
+            patch.fontFamily = normalizeFontFamily(attrs.fontFamily);
+          }
+
+          const mergeAttrs = (
+            existing: { [key: string]: any } | undefined
```
hejsztynx: NIT: maybe let's get rid of the `any`? And later when call this function let's use `existing?.attrs as CustomStyleAttrs | undefined`. wdyt?


```suggestion
            existing: CustomStyleAttrs | undefined
```

hejsztynx (changes requested):
Currently when you apply `Courier New` font (labeled as "Mono") it is not preserved in round-trip. I debugged a bit and it seems like tiptap renders
 `<span style="font-family: &quot;Courier New&quot;;">` 
and in the normalizer we do
`const fontFamily = htmlNode.style.fontFamily;`
and that makes `fontFamily` be equal to `"Courier New"` (including the quotes).
We then inject that quoted string into `style=""` which breaks it.

`const fontFamily = escapeText(htmlNode.style.fontFamily)` excaping those characters here should solve the issue from what I've tested.

Worth checking if native's normalizer need a similar fix

src/web/formats/EnrichedCustomStyle.ts
kacperzolkiewski: Fixed, right now I use `Attrs`  from `@tiptap/pm/model`

hejsztynx (approved):
Looks good!

PR #760, opened 2026-08-12
chore: upgrade example app and library deps to RN 0.87

kacperzolkiewski:
# Summary

Upgrade example app and library deps to RN 0.87

## Test Plan

- Ensure example app builds and works fine
- Ensure example-web app builds and works fine
- Ensure library builds properly

## Screenshots / Videos

n/a

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| Web     |    ❌     |

## Checklist

- [x] E2E tests are passing


kacperzolkiewski (commented):
Update compatibility table in docs when merged

package.json
hejsztynx, line 149: In our [docs](https://docs.swmansion.com/react-native-enriched-html/misc/compatibility#supported-react-native-versions) we have marked that with the `1.2.0` will drop RN `0.81` support. If that's true, we should probably put there `>=0.82` or `0.82-0.87`?

szydlovsky (approved):
All the apps run properly. I'd just address the comment @hejsztynx made about version compat.

hejsztynx:
We used `baseUrl` in `docs/tsconfig.ts` which is now deprecated starting from typescript [6.0](https://github.com/microsoft/TypeScript/issues/62508#issuecomment-3348649259). I removed it from our `tsconfig.ts`, but it is still present in the docs' upstream `@docusaurus/tsconfig`, which causes the `typecheck:docs` to fail. We'd most likely want to bump the docusaurus to at least `3.10` as this was fixed [there](https://github.com/facebook/docusaurus/releases/tag/v3.10.0).

hejsztynx:
> We used `baseUrl` in `docs/tsconfig.ts` which is now deprecated starting from typescript [6.0](https://github.com/microsoft/TypeScript/issues/62508#issuecomment-3348649259). I removed it from our `tsconfig.ts`, but it is still present in the docs' upstream `@docusaurus/tsconfig`, which causes the `typecheck:docs` to fail. We'd most likely want to bump the docusaurus to at least `3.10` as this was fixed [there](https://github.com/facebook/docusaurus/releases/tag/v3.10.0).

I've tried bumping the docusaurus dependency, but `@swmansion/t-rex-ui` isn't compatible with it. I reverted my `baseUrl` change and silenced that typescript error. We can come back to it after the upstream `t-rex-ui` gets updated

package.json
```diff
@@ -51,7 +51,7 @@
     "typecheck:lib": "tsc -p tsconfig.build.json",
     "typecheck:example": "tsc -p apps/example/tsconfig.json",
     "typecheck:example-web": "tsc -p apps/example-web/tsconfig.app.json",
-    "typecheck:docs": "tsc -p docs/tsconfig.json",
+    "typecheck:docs": "cd docs && yarn typecheck",
```
hejsztynx, line 54: We used wrong typescript in that script. The previously used root's typescript got bumped which caused the CI docs builds to fail

exploIF (commented):
Have you tested this library version with every supported RN version?

package.json
kacperzolkiewski: I updated it for nightly.

kacperzolkiewski:
@exploIF, 
Yes I tested it with various supported RN versions.

package.json
kacperzolkiewski: fixed

hejsztynx (approved):

exploIF (approved):

PR #770, opened 2026-08-13
fix(ios): conflicting styles in typing attributes not correctly detected

hejsztynx:
# Summary

There was a bug which you can see in the attached video below. The issue was that when applying `codeblock` style, we apply it to a whole paragraph, and during this process we look for conflicting styles in the paragraph's range. The problem is, we don't look at current typing attributes for conflicting styles, as there is a case when:

- there is no conflicting style in the given paragraph range
- there is a conflicting style in typing attributes

The implemented fix is minimal, as I didn't want to tamper with current logic - that's why we catch that specific case with an `if`.

I've run the e2e tests, everything passes.

## Test Plan

Reproduce the bug from the attached video, play around with toggling different styles.

## Screenshots / Videos

Before:

https://github.com/user-attachments/assets/109cfc99-2614-422c-9f7d-3b0c058e08fe

After:

https://github.com/user-attachments/assets/c10d6d2c-ca1e-432b-a7ae-268bced9c465

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |
| Web     |    ❌     |

## Checklist

- [x] E2E tests are passing


szydlovsky (approved):
Actually nice find and I love the fix!

PR #773, opened 2026-08-14
fix(ios): typing attributes clearing on keystrokes

hejsztynx:
# Summary

There was a bug when in some cases, between selection change and a keystroke, the typing attributes would get cleared, resulting in unstyled text. More info in the `Screenshots / Videos` section.

All iOS e2e tests pass.

## Test Plan

Reproduce the flow from the attached videos, it should be fine now.

## Screenshots / Videos

Before:

When changing the selection programmatically, via arrow keys, the typing attributes are correctly preserved

https://github.com/user-attachments/assets/12cba319-1dd9-4cf0-8d05-8aa5b96228a7

When using screen taps though, the typing attributes are not preserved resulting in the unstyled text, even though (in this case bold) was marked as active

https://github.com/user-attachments/assets/3cbb75e0-0f0d-4c85-82bc-94424899979f

After:

Typing attributes are correctly preserved whatever the selection change method

https://github.com/user-attachments/assets/7f18e399-10ac-4910-9b50-ab2f70f2e912

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |
| Web     |    ❌     |

## Checklist

- [ ] E2E tests are passing


ios/inputAttributesManager/InputAttributesManager.mm
```diff
@@ -11,13 +11,15 @@ @implementation InputAttributesManager {
   NSMutableArray<NSValue *> *_dirtyRanges;
   NSSet *_customAttributesKeys;
   NSMutableSet *_removedTypingAttributes;
+  BOOL _recentManagedTypingAttributesOnlySelection;
```
szydlovsky, line 14: maybe something simpler:

```suggestion
  BOOL _recentOnlySelectionStatus;
```

ios/EnrichedTextInputView.mm
```diff
@@ -2000,6 +2000,14 @@ - (bool)textView:(UITextView *)textView
     return NO;
   }
 
+  // To be sure, we re-run typingAttributes management right before the
+  // character actually lands. Sometimes, between a selection change and the
+  // next keystroke, typing attributes might get removed - this seems like a
+  // native TextKit issue.
+  if (textView.markedTextRange == nil && range.length == 0 && text.length > 0) {
```
szydlovsky, line 2007: I think we might want any `range` length to be working here; pasting some multicharacter stuff should also use these repeated typing attributes

szydlovsky (commented):
I'm kinda shocked that we find such issue this late and there was no community report about it. Could you please try checkouting some much older version (ideally something from, say, April this year) to see if it is an actual issue we haven't addressed for this long, or a regression?

And regardless of the result, we definitely want some E2E for this case!

ios/EnrichedTextInputView.mm
hejsztynx: I agree, but FYI: it seems like pasted text has a custom handle that inserts it without looking at typingAttributes at all, so pasting some plain text at the very end of a bold word will result in an unstyled text. That is also the case with both android and web. It might worth fixing, but it looks like it deserves a separate PR

hejsztynx:
> I'm kinda shocked that we find such issue this late and there was no community report about it. Could you please try checkouting some much older version (ideally something from, say, April this year) to see if it is an actual issue we haven't addressed for this long, or a regression?
> 
> And regardless of the result, we definitely want some E2E for this case!

I've checked a commit from April and the issue is present there. I've also added e2e tests for that specific case.

szydlovsky (approved):
Everything checks out for me now 👍

PR #775, opened 2026-08-14
fix(ios): synced newline typing attributes with actual textStorage attributes

hejsztynx:
# Summary

Managing typing attributes was implemented in such way that when you moved the cursor to an empty line, the typing attributes got reset. This was to prevent picking up the style from the previous line, but it was done unconditionally. That caused the typingAttributes to be reset even though a style is already present in the newline and valid.

## Test Plan

Experiment around, follow the test flow from the attached videos

## Screenshots / Videos

Before:

https://github.com/user-attachments/assets/e0670f31-abf8-4953-b9e9-c97f7f50c1ef

After:

https://github.com/user-attachments/assets/2f9df226-4f90-451e-a611-4b501df50697

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |
| Web     |    ❌     |

## Checklist

- [ ] E2E tests are passing


ios/inputAttributesManager/InputAttributesManager.mm
```diff
@@ -152,6 +168,16 @@ - (void)manageTypingAttributesWithOnlySelection:(BOOL)onlySelectionChanged {
 
       [ParagraphAttributesUtils resetTypingAttributes:_input
                                   preservingAlignment:savedAlignment];
+
+      // newlines could have been made within a valid inline style, so
+      // we want to preserve typingAttributes so they are correctly
+      // acknowledged when typing
+      if (paragraphRange.length == 1) {
```
szydlovsky, line 175: Just unsure about the only `length = 1` case here. What about empty lines? Is it even a realistic case here or am I just confused?

hejsztynx: We check here for present styles, and when `length = 0` it's impossible to have a style there, so there's nothing to check. I tested including the `length=0` condition and everythink seems to work the same. That's how I see it, but I might be confused as well.

Here in the video, that last line is a truly empty line and we have no way of coming back to it and restoring the information that bold should be here. We could make every lines in the editor never empty, by adding a ZWS and it would probably allow us to handle this better. Anyway that's something to consider first, and outside of this scope IMO.

https://github.com/user-attachments/assets/325f7204-48e7-42b7-8357-ce898d329d75

szydlovsky: Yeah, I agree it's outside of the scope. We anyway have no solid way of remembering typing attributes for each empty line, as they are always calculated on-the-go.

szydlovsky (approved):

PR #779, opened 2026-08-19
feat: SPM compatibility

hejsztynx:
# Summary

Mainly, it was needed to change the iOS imports structure, as SPM and CocoaPods have different preferences.

Every change in the `apps/example` is just for testing purposes and will not be committed in the merge. The `react-native-image-picker` functionality got stripped, as it doesn't support SPM.

SPM works with the internal example app. I also simulated a non-local environment by installing the recently published `react-native-enriched-html` release into a fresh RN 0.87 bare-bone example app and applied the `ios` changes - it works there as well.

## Test Plan

Run the example app with SPM

In the `apps/example` directory:

1. first build the app normally with CocoaPods, with `yarn ios`, verify it works
2. in `react-native-config.js` set `automaticPodsInstallation` to `false`
3. remove the `react-native-image-picker` dependency from `package.json` and reinstall dependencies
4. run `npx react-native spm scaffold`, `Package.swift` should not be changed by that
5. in the root `Package.swift` add `"apps"` and `"docs"` to the `exclude` list. This is needed as we physically have example apps within the library source code and don't want to try to resolve source there to compile; otherwise SPM would get confused (duplicated resources, infinite symlink loops)
6. we run `npx react-native spm --deintegrate` to remove CocoaPods from this app
7. we can finally run the build with `yarn ios`

More info: [RN blog post](https://reactnative.dev/blog/2026/08/11/react-native-0.87#experimental-swift-package-manager-support-for-ios)

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |
| Web     |    ❌     |


apps/example/src/hooks/useEditorState.ts
```diff
@@ -212,10 +211,7 @@ export function useEditorState() {
       return;
     }
 
-    const response = await launchImageLibrary({
-      mediaType: 'photo',
-      selectionLimit: 1,
-    });
+    const response: any = undefined;
```
hejsztynx, line 214: Stripped `react-native-image-picker` functionalities for testing purposes only. This will be removed before merging

Package.swift
```diff
@@ -0,0 +1,175 @@
...
+            name: "ReactNativeEnrichedHtml",
+            dependencies: [.product(name: "ReactHeaders", package: "ReactNative"), .product(name: "ReactNativeHeaders", package: "ReactNative"), .product(name: "ReactNativeDependenciesHeaders", package: "ReactNative"), .product(name: "ReactAppHeaders", package: "React-GeneratedCode")],
+            path: ".",
+            exclude: [
+              "node_modules",
+              "android",
+              "lib",
+              "src",
+            ],
+            sources: [
```
szydlovsky, line 51: Will we need to modify this sources list each time we add some new native iOS files? If so, we could add a CI that checks if all the files within the native iOS project are properly listed there and fail otherwise

szydlovsky (approved):
Overall, looks and works good (tested both cases).

szydlovsky (approved):

hejsztynx:
Marking this PR as draft, as it's worth adding SPM builds on the CI

hejsztynx:
> Marking this PR as draft, as it's worth adding SPM builds on the CI

After some discussion, that requires a lot of tampering in the repo and maybe it's not doing for now, as SPM is still experimental

docs/docs/misc/compatibility.mdx
```diff
@@ -86,3 +86,7 @@ Sanitization **is** enforced on the web, as injecting unsafe HTML directly into
 
 On Web those methods are no-ops. These include: `measure`,
 `measureInWindow`, `measureLayout` and `setNativeProps`.
+
+## Swift Package Manager
```
exploIF: This is really not needed, from library side it's working fine

> It's not recommended for production use yet, and its behavior may change in future releases, including breaking changes.

That's up to RN, not something on our side, so it does not need any explanation

exploIF (commented):
I think we need some documentation in the Contributing part about SPM maintenance

hejsztynx:
Nicely integrating the SPM builds into our CI and overall example-app focused development workflow, is blocked by the `react-native-image-picker` dependency in our example app, which doesn't allow that package to be set up with `spm scaffold`. This requires a small fix with compiler flags, I've submitted a [PR](https://github.com/react-native-image-picker/react-native-image-picker/pull/2419).

react-native-spm-prefix.h
```diff
@@ -0,0 +1,6 @@
+#ifdef __OBJC__
```
kacperzolkiewski: Did you run into any compilation errors without that file?
I'm working on adding SPM support for `react-native-screens`, and it seems to work without that extra file.

Package.swift
```diff
@@ -0,0 +1,38 @@
...
+                "cpp/GumboParser",
+                "cpp/parser",
+                "ios"
+            ],
+            publicHeadersPath: "ios",
+            cSettings: [.headerSearchPath("cpp/GumboParser"), .headerSearchPath("cpp/parser"), .headerSearchPath("ios"), .headerSearchPath("ios/config"), .headerSearchPath("ios/enrichedInputTextView"), .headerSearchPath("ios/enrichedTextTextView"), .headerSearchPath("ios/extensions"), .headerSearchPath("ios/generated/ReactCodegen/ReactNativeEnrichedSpec"), .headerSearchPath("ios/htmlParser"), .headerSearchPath("ios/inputAttributesManager"), .headerSearchPath("ios/inputHtmlParser"), .headerSearchPath("ios/interfaces"), .headerSearchPath("ios/internals"), .headerSearchPath("ios/textHtmlParser"), .headerSearchPath("ios/utils"), .headerSearchPath("."), .unsafeFlags(["-include", "react-native-spm-prefix.h"])],
```
kacperzolkiewski: Nit: The `cSettings` and `cxxSettings` arrays are identical. Maybe you could extract them into a single array and reuse it in both places?

react-native-spm-prefix.h
hejsztynx: Yes, currently without that file the build breaks. We use implicit `UIKit` and `Foundation` imports in the codebase - we don't need to import them explicitly to successfully use those packages. We could explicitly import them in every file that needs it and then we could drop that `react-native-spm-prefix.h`. Worth considering

hejsztynx: It turned out we only needed to add imports in a few places. Now there is no `react-native-spm-prefix.h` in the codebase 🎉

exploIF (approved):
Code-wise looks good. If both cases (SPM and Cocoapods) work correctly, then let's merge it

PR #784, opened 2026-08-27
fix(android): initial measuring tweaks

hejsztynx:
# Summary

This fixes a bug where the component's initial height calculation would break if `defaultValue` was provided without an explicit `fontSize` in `htmlStyle` prop. The root cause was a mix of a bad initial estimate and invalid measurements caching:

During `initialMeasure`, missing a passed `fontSize` in props, meant it defaulted to 0, which caused the entire height estimate to be wrong. I fixed this by adding a fallback to the default font size, matching how `EnrichedText` measurements already handle it.

After `initialMeasure`, `setValue` (caused by the present `defaultValue` prop) is run, which tries to update the shadow node state, but `stateWrapper` is still `null`. This caused the `MeasurementStore` internal cache to be out of sync with the actual shadow node's state. Fixed this by providing an early return in `invalidateLayout()` - we don't call `Measurement.store()` at all, if `stateWrapper` is unavailable.

I've also tweaked two small things: in `EnrichedTextInputShadowNode.cpp` the `forceHeightRecalculationCounter_` was never initialized before accessing its value, which would make it have random, garbage values. I've also adjusted the provided `forceHeightRecalculationCounter` to the `stateWrapper`, as its value was always pre-increment during `invalidateLayout()`. I believe it's better to have the shadow node's and component's counters' values consistent.

Sometimes the `setValue`'s layout invalidation could run before the `initialMeasure`. That made the `MeasurementStore` cached measurements correct from the start, as after `setValue` runs, the actual component's size is measured, not estimated like it's the case with `initialMeasure`. You can see this race condition in the attached video.

## Test Plan

I've provided a modified example app, so you can see the bug yourself - you can run

```sh
git checkout 212b2a3e
```

and then you can check out how it works after the fix

```sh
git checkout 4c478090
```

## Screenshots / Videos

The race condition in action (stumbled across in a different app):

https://github.com/user-attachments/assets/1ded9bcc-ce87-49a1-82d8-3300dd41a4a3

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |
| Web     |    ❌     |

## Checklist

- [x] E2E tests are passing


android/src/main/java/com/swmansion/enriched/textinput/MeasurementStore.kt
```diff
@@ -44,16 +44,23 @@ object MeasurementStore {
...
     val paintParams = PaintParams(paint.typeface, paint.textSize)
 
-    data[id] = MeasurementParams(initialized, cachedWidth, size, spannable, paintParams)
-    return cachedSize != size
+    // stateWrapper might be null during the first measures,
+    // we want to cache the size only when it was actually
+    // handled in the shadow node. The cached size and the actual
+    // rendered one should be in sync.
+    val invalidated = cachedSize != size && invalidateShadowNode()
+    val storedSize = if (invalidated) size else cachedSize
```
exploIF: What's the point of measuring size if it's not invalidated due to missing `stateWrapper`?

android/src/main/java/com/swmansion/enriched/textinput/EnrichedTextInputViewLayoutManager.kt
```diff
@@ -11,14 +11,16 @@ class EnrichedTextInputViewLayoutManager(
     val text = view.text
     val paint = view.paint
 
-    val needUpdate = MeasurementStore.store(view.id, text, paint)
-    if (!needUpdate) return
+    MeasurementStore.store(view.id, text, paint) {
```
exploIF: I think previous API was cleaner and would advocate for keeping it as it was and just fixing the bug. Please don't introduce changes that are out of scope, this makes review harder and it's a common source of introducing bugs by accident

exploIF (changes requested):

android/src/main/java/com/swmansion/enriched/textinput/MeasurementStore.kt
hejsztynx: right, now there's a single early return at the top of `invalidateLayout()`

exploIF (approved):

PR #785, opened 2026-08-28
docs: known limitation with blurring the input

hejsztynx:
# Summary

Tapping outside of `EnrichedTextInput` does not blur it on mobile. This is a RN limitation and was mentioned here #184 . Provided mention about that in the docs

### DIFF
--- a/docs/docs/misc/known-limitations.md
+++ b/docs/docs/misc/known-limitations.md
@@ -32,3 +32,8 @@ We may add support for more tags and related functionality over time - see the
 not supported yet, the best way to let us know is to open a
 [GitHub issue](https://github.com/software-mansion/react-native-enriched-html/issues)
 in our repo.
+
+## Tapping outside the editor doesn't blur it
+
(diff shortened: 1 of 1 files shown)


docs/docs/misc/known-limitations.md
```diff
@@ -32,3 +32,8 @@ We may add support for more tags and related functionality over time - see the
 not supported yet, the best way to let us know is to open a
 [GitHub issue](https://github.com/software-mansion/react-native-enriched-html/issues)
 in our repo.
+
+## Tapping outside the editor doesn't blur it
```
exploIF: This is incorrect statement, such functionality does not work in React Native out of the box. We should be clear that what does not work is:
- `Keyboard.dismiss()`
- `keyboardShouldPersistTaps`

We could also explain why it's not working (React Native is not exposing `TextInputState`), so our input cannot be globally registered

exploIF (changes requested):

exploIF (approved):

PR #786, opened 2026-09-01
feat(ios): unify expected import structure across cocoapods and spm

hejsztynx:
# Summary

In the current setup `Package.swift` requires maintaining `cSettings` and `cxxSettings`, which list paths are used to resolve headers. That means on every potential directory structure rename/refactor, we would need to remember to properly update it. We currently have no SPM builds on CI, so it would effectively break SPM compatibility if left unnoticed.

Instead of listing every `ios/` subdirectory in `Package.swift`, we can make the `Package.swift` stable with the `ios` directory listed in the search path headers, which makes maintenance much easier. This forces all the imports in the iOS source code to be `ios/` relative and we want to check that structure during build time. Our overall development workflow uses the CocoaPods based example app to test if the app builds successfully, so I would opt for implementing that import structure check in the current `.podspec` file. That way there is no case that code builds with Pods but not with SPM, the behavior is unified, and we don't need to maintain those `Package.swift` header search paths.

## Test Plan

Build the example app to see it still builds, revert one of the imports and the build should now fail.

You can repeat those steps after setuping SPM from this PR's base's instructions.


szydlovsky (approved):

szydlovsky (approved):
Clever trick with the podspec requirement 👍

exploIF (approved):

PR #788, opened 2026-09-07
fix(ios, android): sync img placeholder colors with current text color

hejsztynx:
# Summary

The inline image's placeholder colors were inconsistent across all platforms:
- web: the color matched the current text color (correct)
- iOS: the color was fixed
- Android: the color was fixed

## Test Plan

Create an inline image with an invalid url to create a img placeholder. Next adjust the `EnrichedTextInput` `style` prop, with a changed `color`. Notice the updated color on the placeholder.

## Screenshots / Videos

Before(iOS):

Screenshot 2026-09-07 at 10 52 52

After (iOS):

Screenshot 2026-09-07 at 10 51 53

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| Web     |    ❌     |

## Checklist

- [x] E2E tests are passing

### DIFF
--- a/android/src/main/java/com/swmansion/enriched/common/AsyncDrawable.kt
+++ b/android/src/main/java/com/swmansion/enriched/common/AsyncDrawable.kt
@@ -13,6 +13,7 @@ import android.os.Build
 import android.os.Handler
 import android.os.Looper
 import android.util.Log
+import androidx.core.graphics.drawable.DrawableCompat
 import androidx.core.graphics.drawable.toDrawable
 import com.swmansion.enriched.R
 import java.net.URL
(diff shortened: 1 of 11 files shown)


android/src/main/java/com/swmansion/enriched/common/AsyncDrawable.kt
```diff
@@ -57,7 +57,9 @@ class AsyncDrawable(
       } catch (e: Exception) {
         Log.e("AsyncDrawable", "Failed to load: $url", e)
 
-        loadPlaceholderImage()
+        mainHandler.post {
+          loadPlaceholderImage()
+        }
```
hejsztynx, line 62: This was caught by Copilot. Previously `loadPlaceholderImage` which mutates `AsyncDrawable` internal state, could be run from a secondary thread. This may lead to some state desynchronization, so we delegate it to the main thread, just like we already do few lines above

szydlovsky (approved):
iOS looks good

PR #794, opened 2026-09-10
fix(web): make input focusable in the whole component's area

### LINKED ISSUE #793: Can't focus by clicking below the first line on web with minHeight
**Describe the bug**
When minHeight is greater than the height of the first line, visually, EnrichedTextInput looks like one large input, so it would be nice if pressing anywhere within the EnrichedTextInput area would give it focus.

**To Reproduce**

1. Use the example from <https://docs.swmansion.com/react-native-enriched-html/fundamentals/your-first-editor#rendering-the-input> which uses a `minHeight` greater than one line.
2. On web with Firefox or Chrome, clicking on the first line gives focus but clicking below that within the minHeight doesn't give focus.

Image

**Expected behavior**
Pressing below the first line would give it focus.

 - Device: Web (Firefox & Chrome)
 - OS: Fedora Linux
 - Version: 44

**Additional context**
Great project, thanks!

hejsztynx:
# Summary

Fixes #793 

All web e2e tests pass. Effectively, nothing should have changed visually. 

## Test Plan

In the example app, experiment with `height`, `minHeight` and `maxHeight` props. Write some text, so the input becomes scrollable. Verify everything looks in place.

## Screenshots / Videos

When `minHeight` was set:

Before:

https://github.com/user-attachments/assets/211a4fa4-c4b4-47a2-b808-fce50272d865

After:

https://github.com/user-attachments/assets/cc7ba1cb-9289-4cde-ab7f-2aa50a696451

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ❌     |
| Web     |    ✅     |

## Checklist

- [x] E2E tests are passing


src/web/EnrichedText.css
```diff
@@ -1,12 +1,21 @@
-.tiptap.ProseMirror p {
+.eti-editor .tiptap.ProseMirror p {
   margin-top: 0;
   margin-bottom: 0;
 }
 
-.tiptap.ProseMirror-focused {
+.eti-editor .tiptap.ProseMirror-focused {
```
hejsztynx, line 6: this was tweaked, so we create CSS rules that will be applied to our components only, not every component that uses `.tiptap` in the whole application.

szydlovsky (approved):

.playwright/tests/images.spec.ts
```diff
@@ -168,6 +168,7 @@ test.describe('images', () => {
 
     for (const key of toolbarOrder) {
       const editor = await focusEnrichedEditable(page);
+      await page.waitForTimeout(10);
```
hejsztynx, line 171: For some reason after main merge, this test broke, and now it needs a slight timeout

exploIF (approved):

PR #796, opened 2026-09-11
fix: repo's typescript and ESlint configuration and errors

hejsztynx:
# Summary

I've done some changes to TypeScript and ESlint related changes to fix already existing errors in the codebase. These errors were not caught by the project's scripts like `yarn lint` or `yarn typecheck`, but were visible in the IDE. This is probably because of the differences how `npm` and IDE runs TS. I've verified that these errors were present in two different IDEs - VSCode and Zed on different machines, so it's a shared issue and worth fixing even if they don't persist into CI.

1. Pinned root's `tsconfig.json` types resolution paths, as we had `react` installed in both root and `apps/example` `node_modules`, and TS would get confused resulting in errors, e.g. in `src/native/EnrichedTextInput.tsx`.

```
// Argument of type 'null' is not assignable to parameter of type 'never'.
focus: () => {
     Commands.focus(nullthrows(nativeRef.current));
},
```

A similar issue was already mentioned in `apps/example/tsconfig.json`:

```
      // Pin a single `@types/react` copy so the library source typechecked here
      // doesn't mix react types with the example's copy (which makes codegen
      // command ref types collapse to `never`).
      "react": ["../../node_modules/@types/react"],
      "react/*": ["../../node_modules/@types/react/*"]
```

2. With the RN 0.87 bump, we also bumped root TypeScript to `6.x`, and in `example-web` we still use TS `5.x`. I believe it's better to have the `example-web` use the root TS, just like the `apps/example` already does, and have a single TS across the repo, whenever possible

3. `example-web` had its own separate ESLint and the its code wasn't linted to those rules. There were two options: tweak the `example-web` code so it respects the configured standalone ESLint, or remove that ESLint and let the root ESLint handle the `example-web`. I opt for the latter one, as we keep linting rules consistent across the repo, we don't have to maintain another ESLint configuration, it doesn't require modifying the example app's code, and we've been using the root ESLint in `example-web` via `yarn lint` anyway

4. In `.playwright`, the TS wasn't setup correctly and we got errors e.g. in `images.spec.ts`, we got:

```
Cannot find name 'Buffer'. Do you need to install type definitions for node? Try `npm i --save-dev @types/node` and then add 'node' to the types field in your tsconfig.
```

I added the `@types/nodes` package and wired up `.playwright` with TS

5. Linted the code where it wasn't.

6. Removed deprecated `React.ElementRef` in favor of `React.ComponentRef` in `src/spec/EnrichedTextInputNativeComponent.ts`

## Test Plan

What I've verified:

- `yarn test`
- `yarn test:e2e:web`
- `yarn test:e2e:mobile` - the tests build and run, but I haven't run them all through
- `yarn ios`
- `yarn android`
- `yarn lint`
- `yarn typecheck`

You can also run these scripts - everything should pass

## Checklist

- [x] E2E tests are passing


apps/example-web/package.json
```diff
@@ -5,28 +5,19 @@
   "type": "module",
   "scripts": {
     "dev": "vite",
-    "build:web": "tsc -b && vite build",
-    "preview": "vite preview",
-    "lint": "eslint . --ext ts,tsx"
+    "build:web": "vite build",
```
hejsztynx, line 8: I believe we can safely remove that `tsc` call. It was there, because we had a standalone TS, but now we use the root one and we use the root `yarn typecheck`

szydlovsky (approved):
nice cleanup

